PR #15, opened 2018-08-21
Return LocalPath instead of AbsoluteUri for absolute URIs

EasyRhinoMSFT:
File.Exists doesn't handle URI-formatted local paths


lgolding:
Ship it!

Issue #16, opened 2018-08-21
Viewer shows expansion chrome but expanding provides no additional information


michaelcfanning:
![image](https://user-images.githubusercontent.com/13526455/44407525-36f83d00-a513-11e8-930e-15ded5da09ef.png)

Load one of the new FxCop SARIF files to see this (checked in shortly)

EasyRhinoMSFT:
Needs SDK package with Microsoft/sarif-sdk#pull/1048 and Microsoft/sarif-sdk/pull#1050

PR #21, opened 2018-08-27
Fix issue #16 result description

EasyRhinoMSFT:
When we create an error list entry, we check the Message string (the full rule description) for embedded links. Currently, if none are found, an inline is created for the entire string. What *should* happen is no inlines be returned, and a different part of the code provides a shortened string. The fix is to demote the scope of the inline block so it only executes when embedded links are present.


lgolding:
Ship it!

Issue #22, opened 2018-08-28
Re-enable unit test after next SDK update


EasyRhinoMSFT:
SarifErrorListItemTests.SarifErrorListItem_WhenResultRefersToExistingMessageString_ContainsExpectedMessage
Need SDK PR Microsoft/sarif-sdk/pull/1048 before this test will pass.

PR #23, opened 2018-08-30
Prepare for v2 Viewer release

EasyRhinoMSFT:
- Add support for transforming v1 logs
- Update SDK package references
- Re-enable unit test (issue #22)
- Update extension package icon


src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -22,13 +25,17 @@ namespace Microsoft.Sarif.Viewer.ErrorList
 {
     public class ErrorListService
     {
+        private const int IDYES = 6;
```
lgolding: Are these defined somewhere central? I see you use `OLEMSGICON` constants in your call to `ShowMessageBox`.

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -38,6 +45,78 @@ public static void ProcessLogFile(string filePath, Solution solution, string too
             if (toolFormat.MatchesToolFormat(ToolFormat.None))
             {
                 logText = File.ReadAllText(filePath);
+
+                if (logText.Contains(@"""version"": ""1.0.0"""))
```
lgolding: You need a regex because you also have to match `"version"   :   \t\n  "1.0.0"`.

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -38,6 +45,78 @@ public static void ProcessLogFile(string filePath, Solution solution, string too
...
+                    // They're opening a v1 log, so we need to transform it
+                    // Ask if they'd like to save the v2 log
+                    int response = VsShellUtilities.ShowMessageBox(SarifViewerPackage.ServiceProvider,
+                                                                   Resources.TransformV1_DialogMessage,
+                                                                   null, // title
+                                                                   OLEMSGICON.OLEMSGICON_QUERY,
+                                                                   OLEMSGBUTTON.OLEMSGBUTTON_YESNOCANCEL,
+                                                                   OLEMSGDEFBUTTON.OLEMSGDEFBUTTON_FIRST);
+
+                    if (response == IDCANCEL)
```
lgolding: 1. Why "yes/no/cancel" as opposed to just "yes/no"?
2. If you do use "yes/no/cancel" then you need to return on `IDNO`, too.

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -38,6 +45,78 @@ public static void ProcessLogFile(string filePath, Solution solution, string too
...
+                    {
+                        ContractResolver = SarifContractResolverVersionOne.Instance,
+                    };
+
+                    SarifLogVersionOne v1Log = JsonConvert.DeserializeObject<SarifLogVersionOne>(logText, settingsV1);
+                    var transformer = new SarifVersionOneToCurrentVisitor();
+                    transformer.VisitSarifLogVersionOne(v1Log);
+                    log = transformer.SarifLog;
+
+                    if (response == IDYES)
```
lgolding: Related to above comment: Why go through the work of transforming if it turns out they said "No"?

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -38,6 +45,78 @@ public static void ProcessLogFile(string filePath, Solution solution, string too
             if (toolFormat.MatchesToolFormat(ToolFormat.None))
             {
                 logText = File.ReadAllText(filePath);
+
+                if (logText.Contains(@"""version"": ""1.0.0"""))
+                {
+                    // They're opening a v1 log, so we need to transform it
```
lgolding: Periods at ends of sentences.

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -38,6 +45,78 @@ public static void ProcessLogFile(string filePath, Solution solution, string too
...
+
+                    SarifLogVersionOne v1Log = JsonConvert.DeserializeObject<SarifLogVersionOne>(logText, settingsV1);
+                    var transformer = new SarifVersionOneToCurrentVisitor();
+                    transformer.VisitSarifLogVersionOne(v1Log);
+                    log = transformer.SarifLog;
+
+                    if (response == IDYES)
+                    {
+                        // Prompt for a location to save the transformed log
+                        var saveFileDialog = new SaveFileDialog();
```
lgolding: None of this code is unit testable because it pops UI. Please file an issue to revisit and add tests.

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -38,6 +45,78 @@ public static void ProcessLogFile(string filePath, Solution solution, string too
...
+
+                        try
+                        {
+                            File.WriteAllText(filePath, JsonConvert.SerializeObject(log, settingsV2));
+                        }
+                        catch (UnauthorizedAccessException)
+                        {
+                            error = string.Format(Resources.SaveTransformedV1LogFail_Access_DialogMessage, filePath);
+                        }
+                        catch (IOException ex)
```
lgolding: `Exception` instead of `IOException`?

It used to be that the guidelines said not to catch `Exception`, because there were some exceptions (like `StackOverflowException` and `ThreadAbortException` that were dangerous to catch.

I _think_ that nowadays user code can't catch those dangerous exceptions, and it's ok to catch `Exception`. Please look this up and change if appropriate.

src/Sarif.Viewer.VisualStudio/Resources.resx
```diff
@@ -160,6 +160,20 @@
...
+  </data>
+  <data name="SaveTransformedV1LogFail_General_Dialog" xml:space="preserve">
+    <value>The log file couldn't be saved: {0}</value>
+    <comment>{0} will be the exception message</comment>
+  </data>
+  <data name="SaveTransformedV1Log_DialogTitle" xml:space="preserve">
+    <value>Save Transformed Log File</value>
+  </data>
+  <data name="TransformV1_DialogMessage" xml:space="preserve">
+    <value>The log file you have opened is SARIF version 1. We will automatically transform this file to SARIF version 2. Would you like to save the transformed log file?</value>
```
lgolding: Check with a UI guru about saying "we". Perhaps "The log file will automatically be transformed to SARIF version 2...."

lgolding:
@EasyRhinoMSFT Note that I reviewed this in GitHub because it wouldn't open for me in CodeFlow.

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
EasyRhin0: This is true (the CLR will automatically re-throw those). Still, I prefer to be explicit, especially with a method that can throw so many different exceptions. I'd like to leave this one.

---
In reply to: [214177624](https://github.com/Microsoft/sarif-visualstudio-extension/pull/23#discussion_r214177624) [](ancestors = 214177624)

lgolding: Could you tell me a little more? As you say, there might be many types of exceptions. As the code stands, most of them will not be caught, and will (presumably cause the extension to crash).  That can't be the desired behavior, right?

EasyRhin0: Definitely not. I'll catch everything rather than handling all of them separately.

---
In reply to: [214191779](https://github.com/Microsoft/sarif-visualstudio-extension/pull/23#discussion_r214191779) [](ancestors = 214191779)

lgolding: Sorry, I wasn't clear. I certainly didn't mean to suggest that you have several catch clauses, one for each type of exception. I meant to suggest that you add a `catch (Exception ex)` to your existing `try/catch` construct.

Do you have a top-level unhandled-exception-catcher already?

lgolding: (I mean, in some other file?)

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
EasyRhin0: This method returns int, but there is another that uses an enum with the same values. I forgot I had stumbled across that ages ago.

---
In reply to: [214174625](https://github.com/Microsoft/sarif-visualstudio-extension/pull/23#discussion_r214174625) [](ancestors = 214174625)

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
lgolding: Ah! No wonder I didn't see it. :-) Ok, then, standing by.

lgolding: Sorry, I was confused, because when you wrote "I'd like to leave this one," it sounded like "Won't fix". My mistake.

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -38,6 +44,85 @@ public static void ProcessLogFile(string filePath, Solution solution, string too
             if (toolFormat.MatchesToolFormat(ToolFormat.None))
             {
                 logText = File.ReadAllText(filePath);
+                string pattern = @"""version"":\s*""1.0.0""";
```
lgolding: White space is also allowed before the colon.

Issue #46, opened 2018-10-17
Unprocessed tokens appear in secondary error list description


EasyRhinoMSFT:

EasyRhinoMSFT:
![capture](https://user-images.githubusercontent.com/32752981/47117936-a55b3200-d21a-11e8-81d3-077759f20a7f.png)

EasyRhinoMSFT:
This is Comprehensive.sarif

EasyRhinoMSFT:
There is something wrong with our error list data provider implementation. I don't fully understand yet, but I think we are not handling snapshots correctly.

PR #51, opened 2018-11-15
Fix #28 and #46 error list item text

### LINKED ISSUE #28: Error list shows all sentences in multisentence output
We correctly create an expander for multisentence error list entries. But we show the full contents for the unexpanded error list entry. we should only show the first sentence for this case (and the remainder of the error should display when expanded).

EasyRhinoMSFT:
The only caveat here is with result messages that contain hyperlinks: we can't render hyperlinks in the expando area, which means we can't limit the item's text to the first sentence in case there are links beyond that. We COULD add code to handle this case (truncate as early as possible so all links are included) but I elected not to do that in this change.


michaelcfanning:
Do you have any suggestions for test files that help illustrate this change? If you had any generated test content that you used even to manually verify your work, that would be helpful to check in.

michaelcfanning:
LGTM

PR #61, opened 2018-11-27
Revamp Stacks view

EasyRhinoMSFT:
Screenshots here: \\\\scratch2\scratch\v-chrmey\SARIF Viewer\Stacks


michaelcfanning:
Please merge line details into core column. Please name that column 'Name' as in the VS UI. We can preserve the file column. Nice work, greatly improved!

PR #62, opened 2018-11-29
Use VS 'category' column for result baseline state, if present

michaelcfanning:
Chris, a proposal. The VS 'Category' columns picks up the contents of a property named 'category' if one exists. It turns out that displaying baseline state information would be useful in the UI. Can we repurpose the Category column for this? This PR shows that change. Note that I left the Category in the Rule model (but can remove that if you indicate I should).

Alternately, we could try to create a new error list column entirely if that is supported (i don't know if it is).


michaelcfanning:
baseline_state_as_category

PR #73, opened 2019-02-15
Update to SARIF SDK v2.0.0-csd.2.beta.2019-01-24.1

EasyRhinoMSFT:


michaelcfanning:
Update package name for interop dll. Otherwise let's :shipit:

PR #88, opened 2019-05-23
Update SDK package refs + prerelease transformer logic

EasyRhin0:
@michaelcfanning I think this is what you're asking for...


src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -79,8 +79,15 @@ public static void ProcessLogFile(string filePath, Solution solution, string too
                 }
                 else
                 {
-                    // They're opening a v2 log, so send it through the pre-release compat transformer
-                    PrereleaseCompatibilityTransformer.UpdateToCurrentVersion(logText, Formatting.Indented, out logText);
+                    // They're opening a v2 log; check to see if it's the version we are referencing.
+                    pattern = $@"""version""\s*:\s*""{VersionConstants.StableSarifVersion}""";
```
michaelcfanning: >pattern [](start = 20, length = 7)

1) I'm curious, maybe we should test whether this is in fact saving us any time. i.e., get an extremely large log file and test the version sniff vs. running it through the transformer.
2) one of the symptoms of that other empty log file was that I observed we seemed to be producing a temporary version of the file (I assumed because it had been transformed. I don't think this change will fix that because you aren't setting the output path. (i think you should do this below in the event that match.Success is true?)
3) we just updated the SDK on nuget.org, any chance we can sneak in an upgrade to it?

michaelcfanning: btw - scott louvau can provide a SARIF log file with a large # of results for testing if you'd like to explore #1. I'm fine with deferring this work for now if you are busy. I think items #2 and #3 shouldn't cost much, so hopeful you can take them on. thanks a ton for this contribution, easy. 

---
In reply to: [287412015](https://github.com/microsoft/sarif-visualstudio-extension/pull/88#discussion_r287412015) [](ancestors = 287412015)

EasyRhin0: OK thx will investigate.

EasyRhin0: So what I did is:
1. Only run the regex against the first 100 characters. The spec says the version property should be first, so I think this is a fair assumption. A messagebox pops if the version property isn't found.
2. Fixed the temp file stuff. If the log is current, it's used in-place.
3. I was stupidly doing a second regex if the log wasn't v1. Refactored as part of item 1.
4. Updated the SDK package reference.

EasyRhin0: 5. Added a prompt to save the prerelease-transformed log, just like we do for v1->v2 transforms.

src/Sarif.Viewer.VisualStudio/Resources.resx
```diff
@@ -181,11 +181,18 @@
...
     <comment>{0} will be the exception message</comment>
   </data>
+  <data name="SaveTransformedPrereleaseV2Log_DialogTitle" xml:space="preserve">
+    <value>Save Transformed Log File</value>
+  </data>
   <data name="SaveTransformedV1Log_DialogTitle" xml:space="preserve">
     <value>Save Transformed Log File</value>
   </data>
+  <data name="TransformPrereleaseV2_DialogMessage" xml:space="preserve">
+    <value>The log file you have opened is formatted with a pre-release SARIF version 2 schema. This file will be automatically transformed to SARIF version {0}. Would you like to save the transformed file?</value>
```
lgolding: >is formatted with [](start = 40, length = 17)

"is formatted with" => "conforms to"

src/Sarif.Viewer.VisualStudio/Resources.resx
```diff
@@ -194,4 +201,7 @@
   <data name="UnknownToolName" xml:space="preserve">
     <value>Unknown</value>
   </data>
+  <data name="VersionPropertyNotFound_DialogTitle" xml:space="preserve">
+    <value>The sarifLog.version property was not found at the beginning of the sarifLog object. Per the SARIF specification, the version property should appear first.</value>
```
lgolding: >The [](start = 11, length = 3)

"The SARIF specification recommends that the version property appear at the beginning of the log file. This log file does not conform to that recommendation."

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -29,58 +29,110 @@ namespace Microsoft.Sarif.Viewer.ErrorList
...
 
             if (toolFormat.MatchesToolFormat(ToolFormat.None))
             {
                 logText = File.ReadAllText(filePath);
-                string pattern = @"""version""\s*:\s*""1.0.0""";
-                Match match = Regex.Match(logText, pattern, RegexOptions.Compiled | RegexOptions.Multiline);
+                Match match;
+                
+                // Get the schema version of the unmodified input log
+                using (StringReader reader = new StringReader(logText))
```
lgolding: >StringReader [](start = 23, length = 12)

`var` is ok when type is explicit on the rhs.

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -29,58 +29,110 @@ namespace Microsoft.Sarif.Viewer.ErrorList
...
+                {
+                    // Read the first 200 characters.
+                    char[] buffer = new char[200];
+                    reader.ReadBlock(buffer, 0, buffer.Length);
+                    match = Regex.Match(new string(buffer), VersionRegexPattern, RegexOptions.Compiled);
+                }
 
-                if (match.Success)
+                if (match?.Success == true)
```
lgolding: >? [](start = 25, length = 1)

No need for the null conditional operator (and thus, no need for the explicit comparison against `true`. `Regex.Match` is guaranteed to return a non-null `Match` object.

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -29,58 +29,110 @@ namespace Microsoft.Sarif.Viewer.ErrorList
...
                 {
-                    // They're opening a v1 log, so we need to transform it.
-                    // Ask if they'd like to save the v2 log.
-                    MessageDialogCommand response = PromptToSaveProcessedLog(Resources.TransformV1_DialogMessage);
+                    string inputVersion = match.Groups["version"].Value;
 
-                    if (response == MessageDialogCommand.Cancel)
+                    if (inputVersion == "1.0.0")
```
lgolding: >"1.0.0" [](start = 40, length = 7)

We have `SarifUtilities.V1_0_0`.

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
lgolding: But wait. You already slurped the whole file. You don't need a reader, just:
```
const int HeadLength = 200;
string head = logText.Substring(0, HeadLength);
Match match = Regex.Match(head, ...);
if (match.Success)
{
    ...
}
```

---
In reply to: [294052626](https://github.com/microsoft/sarif-visualstudio-extension/pull/88#discussion_r294052626) [](ancestors = 294052626)

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -29,58 +29,110 @@ namespace Microsoft.Sarif.Viewer.ErrorList
...
+                        outputPath = filePath;
+                        saveOutputFile = false;
+                    }
                 }
                 else
                 {
-                    // They're opening a v2 log, so send it through the pre-release compat transformer
-                    PrereleaseCompatibilityTransformer.UpdateToCurrentVersion(logText, Formatting.Indented, out logText);
+                    // The version property wasn't found within the first 100 characters.
+                    // Per the spec, it should appear first in the sarifLog object.
```
lgolding: >should [](start = 40, length = 6)

"Should" is the operative word. The spec suggests that so that a consumer can quickly determine the version by sniffing the file. If we're going to slurp the whole file anyway, you might consider JSON-tokenizing it and just finding the `version` property.

Having said that, we need to get this fix in, so I'd be ok with considering this for a follow-on change.

lgolding:
I suggested some improvements -- I do think you should make the simple change that avoids the user of a reader -- but we can defer further consideration of how to find the `version` property. So,LGTM.

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
EasyRhin0: OK

Issue #100, opened 2019-08-06
ErrorListService is not unit testable


lgolding:
The `ErrorListService` class is over 400 lines long and is not unit testable. It uses `VsShellUtilities` to prompt the user, and it write directly to the file system.

PR #101, opened 2019-08-06
Fix #98: Some files won't open.

### LINKED ISSUE #98: Won't open files even though version property appears at the beginning
When opening a valid SARIF file, the viewer complains:

> The SARIF specification recommends that the version property appear at the beginning of the log file. This log file does not conform to that recommendation.

... and refuses to open the file. But the `version` property _does_ appear at the beginning and the file opened fine last week.

lgolding:
In the last commit, the regex that describes the `version` property was broken. Spurious backslash characters were introduced, and the optional whitespace before the colon character (which _is_ permitted by the JSON spec) was removed.

Also:
- Even though `ErrorListService` as a whole is not unit testable (I filed #100), I extracted a small static method that I could test. This is the Sprout Method (59) pattern from Feathers, "Working Effectively With Legacy Code".

Adding the unit tests revealed another bug: the extension would not have opened a SARIF file that was fewer than 200 characters long.

### DIFF
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -21,3 +21,6 @@
 * FEATURE: add support for embedded hyperlinks in the error list that point to web URLs
 * Updated SDK compatibility with [v2.1.0-rtm.2](https://www.nuget.org/packages/Sarif.Sdk/2.1.2)
 * Optimized file loading, transform, and SARIF version check logic
+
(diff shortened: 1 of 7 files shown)


lgolding:

lgolding:

lgolding:

lgolding:
@jeffersonking is added to the review. #Closed

PR #103, opened 2019-08-09
Fix #85: Can't open files containing empty string URIs

### LINKED ISSUE #85: Cannot open SARIF files containing URIs that are empty strings
If I attempt to open the /analyze:log sarif output from Visual Studio 2019 (16.0.3 release or 16.1.0, preview 2), both Visual Studio 2019 and Visual Studio 2017 produce the same error dialog.  It says, "Value cannot be null.  Parameter name: path."  I have no information within the standard that says a path parameter exists.  Please refer to [section I.5 of the official standard ](http://docs.oasis-open.org/sarif/sarif/v2.0/sarif-v2.0.html).  

I suppose there is a small chance that the _environmentVariables_ area under invocations is the cause.  However, when I search the sarif file output from Visual Studio 2019, there is no entry anywhere in the file that is tagged _environmentVariables_.  If this is an issue with Visual Studio 2019, then it would make sense.  This file output type is not listed on [the official analyze option page for 2019](https://docs.microsoft.com/en-us/cpp/build/reference/analyze-code-analysis?view=vs-2019).  For this issue, use the Microsoft [Sample C++ project for code analysis](https://docs.microsoft.com/en-us/visualstudio/code-quality/demo-sample?view=vs-2019).

lgolding:
This PR updates the VSIX to use version 2.1.13 of the SARIF SDK, which fixes Microsoft/sarif-sdk#1632, the root cause of #85.

### DIFF
--- a/src/build.props
+++ b/src/build.props
@@ -19,7 +19,7 @@
     <Company Condition=" '$(Company)' == '' ">Microsoft</Company>
     <Product Condition=" '$(Product)' == '' ">Microsoft SARIF Viewer for Visual Studio</Product>
     <Copyright Condition=" '$(Copyright)' == '' ">© Microsoft Corporation. All rights reserved.</Copyright>
-    <VersionPrefix Condition=" '$(VersionPrefix)' == ''">2.1.0</VersionPrefix>
+    <VersionPrefix Condition=" '$(VersionPrefix)' == ''">2.1.4</VersionPrefix>
     <VersionSuffix Condition=" '$(VersionSuffix)' == ''">rtm.4</VersionSuffix>
   </PropertyGroup>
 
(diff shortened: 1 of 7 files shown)


lgolding:

lgolding:

lgolding:

src/build.props
```diff
@@ -19,7 +19,7 @@
     <Company Condition=" '$(Company)' == '' ">Microsoft</Company>
     <Product Condition=" '$(Product)' == '' ">Microsoft SARIF Viewer for Visual Studio</Product>
     <Copyright Condition=" '$(Copyright)' == '' ">© Microsoft Corporation. All rights reserved.</Copyright>
-    <VersionPrefix Condition=" '$(VersionPrefix)' == ''">2.1.0</VersionPrefix>
```
lgolding: >0 [](start = 61, length = 1)

We neglected to update this the last few times.

PR #105, opened 2019-09-16
Fix #104: File open dialog restricts filename extension

### LINKED ISSUE #104: File open dialog restricts filename extension
1. Select menu item **Tools**, **Import Static Analysis Log File to Error List**, and then select any sub-menu item (for example, **Import SARIF file** or **Import Pylint log file**. The Open File dialog opens.

image

Note that the filter dropdown only offers one option: `Pylint log files (*.json)`. It should also have an option `All files (*.*)` in case the file has a different extension.

You _can_ open files with other extensions, but you have to type the filename by hand because it won't appear in the files list.

lgolding:


lgolding:
@michaelcfanning is added to the review. #Closed

lgolding:

lgolding:

lgolding:

PR #118, opened 2020-04-06
Fix #117: Embedded link with slash within anchor text does not work

### LINKED ISSUE #117: Embedded link with slash within anchor text does not work.
For example, try opening a test 

'[..\directory\myfile.cpp](https://example.com)]' is not correct.

This is due to the regex driving link discovery in `SdkUIUtilities.EmbeddedLinkPattern`. The <text> group will halt on encountering a backslash.  See the ^\\\\\\] segment below:

~~~
        // Embedded link format: [link text](n|uri) where n is a non-negative integer, or uri is an absolute URL
        private const string EmbeddedLinkPattern = @"\[(?<text>[^\\\]]+)\]\((?<target>[^)]+)\)";
~~~

@EasyRhinoMSFT may have some insight into why this exists or maybe there is a clue in the unit tess

lgolding:
The regex explicitly forbade backslashes in the link text for no obvious reason.

Also:
- Improve test names.
- Clarify regex with pattern white space.

With this fix, this test file:

[Bug117.sarif.txt](https://github.com/microsoft/sarif-visualstudio-extension/files/4440889/Bug117.sarif.txt)

... renders (and functions) properly:

LinkWithBackslashes

### DIFF
--- a/src/Sarif.Viewer.VisualStudio.UnitTests/SdkUiUtilitiesTests.cs
+++ b/src/Sarif.Viewer.VisualStudio.UnitTests/SdkUiUtilitiesTests.cs
@@ -15,7 +15,7 @@ namespace Microsoft.Sarif.Viewer.VisualStudio.UnitTests
         private void Hyperlink_Click(object sender, RoutedEventArgs e) { }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


lgolding:
@michaelcfanning is added to the review. #Closed

PR #120, opened 2020-04-07
Omit suppressed and absent

lgolding:
**DRAFT**. Do not merge.


src/Sarif.Viewer.VisualStudio/Models/SarifErrorListItem.cs
```diff
@@ -60,9 +60,12 @@ public SarifErrorListItem(Run run, Result result, string logFilePath, ProjectNam
             Category = result.GetCategory();
             Region = result.GetPrimaryTargetRegion();
             Level = result.Level != FailureLevel.Warning ? result.Level : Rule.FailureLevel;
-            SuppressionState = (result.Suppressions != null && result.Suppressions.Count > 0) ?
-                result.Suppressions[0].State :
-                SuppressionState.None;
+
+            if (result.Suppressions?.Count > 0)
```
lgolding, line 64: >Count [](start = 37, length = 5)

What if all of the `suppression` objects have `status == "rejected"`?

PR #125, opened 2020-04-10
Fix #119, Fix #121: Integer & embedded content links don't work without regions

### LINKED ISSUE #119: Integer links don't work
Given the attached SARIF file, if you click on the link in the first issue (rule `TST0001`), it doesn't work. This is an integer link, and there is a `relatedLocation` with that `id`. If you click on the link in the second issue (rule `TST0002`), it works (assuming you have a real file at the location specified in the `file:` URI link).

[Bug119.sarif.txt](https://github.com/microsoft/sarif-visualstudio-extension/files/4441309/Bug119.sarif.txt)

### LINKED ISSUE #121: VS doesn't extract and display files in the absence of region data
* Create a SARIF file w/results that reference artifacts with textual or binary contents data (but no regions associated with results.
* Open SARIF file in VS viewer and double-click result.

**Result**
File isn't expanded from log file and loaded into viewer.

If a modify the log file to add any sort of region to the result location physicalLocation, everything works as expected.

@lgolding

lgolding:
Whoever knows most about the VS Extension please take this PR. I'm afraid it might be @michaelcfanning but don't want to put this on him unless necessary.

In both scenarios (clicking an integer link, and navigating to embedded context without a region specified), the code went down a path that refused to navigate unless there was a region. I fixed it by adding to `CodeLocationObject.NavigateTo` just that portion of `ResultTextMarker.NavigateTo` that handled the file navigation, leaving out the portion that dealt with highlighting the region.

Normally I would refactor the code that is common between the two methods, but I didn't see an easy way to do that here.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs
+++ b/src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs
@@ -4,6 +4,7 @@
 using System;
 using System.ComponentModel;
 using System.Diagnostics;
+using System.IO;
 using Microsoft.CodeAnalysis.Sarif;
 using Microsoft.VisualStudio.Shell.Interop;
 using Microsoft.VisualStudio.Text;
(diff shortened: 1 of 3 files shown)


michaelcfanning:
Hello. Any chance for a test here? Or have we not plumbed things through sufficiently?

michaelcfanning:
Absent a automated test, I can try to compile and test it. will get your branch.

---
In reply to: [612186112](https://github.com/microsoft/sarif-visualstudio-extension/pull/125#issuecomment-612186112) [](ancestors = 612186112)

lgolding:
The code base isn't set up for it. There are no unit tests of navigation at all. It would require mocking interfaces such as IVsUIShellOpenDocument, IVsRunningDocumentTable, IVsWindowFrame, and IServiceProvider. It definitely needs to be done but I don't think doing it now is in line with your priorities.

---
In reply to: [612186112](https://github.com/microsoft/sarif-visualstudio-extension/pull/125#issuecomment-612186112) [](ancestors = 612186112)

PR #126, opened 2020-04-11
Display related location messages in SARIF Explorer

### LINKED ISSUE #92: Info tab does not display level
image

lgolding:
Also:
- Fix #92: Info tab does not display level.
- Emit the locations in the correct order (they were reversed).

@rtaket FYI

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/Models/SarifErrorListItem.cs
+++ b/src/Sarif.Viewer.VisualStudio/Models/SarifErrorListItem.cs
@@ -76,20 +76,22 @@ namespace Microsoft.Sarif.Viewer
             Invocation = run.Invocations?[0]?.ToInvocationModel();
             WorkingDirectory = Path.Combine(Path.GetTempPath(), _runId.ToString());
 
-            if (result.Locations != null)
+            if (result.Locations?.Any() == true)
             {
-                foreach (Location location in result.Locations)
+                // Adding in reverse order will make them display in the correct order in the UI.
+                for (int i = result.Locations.Count - 1; i >= 0; --i)
                 {
-                    Locations.Add(location.ToLocationModel());
+                    Locations.Add(result.Locations[i].ToLocationModel());
                 }
             }
 
-            if (result.RelatedLocations != null)
+            if (result.RelatedLocations?.Any() == true)
             {
-                foreach (Location location in result.RelatedLocations)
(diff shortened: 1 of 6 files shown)


src/Sarif.Viewer.VisualStudio/Models/SarifErrorListItem.cs
```diff
@@ -76,20 +76,22 @@ public SarifErrorListItem(Run run, Result result, string logFilePath, ProjectNam
             Invocation = run.Invocations?[0]?.ToInvocationModel();
             WorkingDirectory = Path.Combine(Path.GetTempPath(), _runId.ToString());
 
-            if (result.Locations != null)
+            if (result.Locations?.Any() == true)
```
michaelcfanning, line 79: >== true [](start = 40, length = 7)

you can remove direct comparison to boolean here and elsewhere in this change, doesn't materially improve readability

src/Sarif.Viewer.VisualStudio/Models/SarifErrorListItem.cs
```diff
@@ -76,20 +76,22 @@ public SarifErrorListItem(Run run, Result result, string logFilePath, ProjectNam
...
-                    Locations.Add(location.ToLocationModel());
+                    Locations.Add(result.Locations[i].ToLocationModel());
                 }
             }
 
-            if (result.RelatedLocations != null)
+            if (result.RelatedLocations?.Any() == true)
             {
-                foreach (Location location in result.RelatedLocations)
+                for (int i = result.RelatedLocations.Count - 1; i >= 0; --i)
```
michaelcfanning, line 90: >result [](start = 29, length = 6)

hm I'll bet @easyrhinomsft could have fixed this in the XAML. :)

src/Sarif.Viewer.VisualStudio/Models/SarifErrorListItem.cs
lgolding: No, because the left-hand side is a nullable bool (values true, false, or null) and so can't be implicitly converted to bool.

src/Sarif.Viewer.VisualStudio/Models/SarifErrorListItem.cs
lgolding: Very likely :)

PR #134, opened 2020-04-28
Fix #132, file open does not work in absence of artifactLocation URI.

### LINKED ISSUE #132: Cannot open a file when artifactLocation.uri is null
[ArtifactLocationUriNull.sarif.txt](https://github.com/microsoft/sarif-visualstudio-extension/files/4495307/ArtifactLocationUriNull.sarif.txt)

1. Save the attached file as .sarif.

2. Open it in the VS viewer.

Expected: The file opens and displays one result.

Actual: The file does not open, and the viewer displays an exception:

ValueCannotBeNull

The call stack is:
```
System.ArgumentNullException
  HResult=0x80004003
  Message=Value cannot be null.
Parameter name: key
  Source=mscorlib
  StackTrace:
   at System.Collections.Generic.Dictionary`2.FindEntry(TKey key)
   at System.Collections.Generic.Dictionary`2.ContainsKey(TKey key)
   at Microsoft.Sarif.Viewer.ProjectNameCache.SetName(String fileName) in C:\dev\sarif-visualstudio-extension\src\Sarif.Viewer.VisualStudio\ProjectNameCache.cs:line 32

  This exception was originally thrown at this call stack:
    System.Collections.Generic.Dictionary<TKey, TValue>.FindEntry(TKey)
    System.Collections.Generic.Dictionary<TKey, TValue>.ContainsKey(TKey)
    Microsoft.Sarif.Viewer.ProjectNameCache.SetName(string) in ProjectNameCache.cs
```

The viewer does not handle the case where an `artifactLocation` object contains _only_ an `index` property (pointing into `run.artifacts`) and does not contain a `uri` property.

The VSCode viewer and the web viewer both handle this case.

@michaelcfanning

michaelcfanning:
@lgolding, this is a much needed fix. Could use a test but I don't have time to work on one. 

What we really need here is better helper support in the SDK for retrieving a comprehensively populated location object, analogous to the Result.GetRule(Run) helper.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/CodeFlowToTreeConverter.cs
+++ b/src/Sarif.Viewer.VisualStudio/CodeFlowToTreeConverter.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Collections.Generic;
 using Microsoft.CodeAnalysis.Sarif;
 using Microsoft.Sarif.Viewer.Models;
@@ -9,7 +10,7 @@ namespace Microsoft.Sarif.Viewer.VisualStudio
 {
     internal static class CodeFlowToTreeConverter
     {
-        internal static List<CallTreeNode> Convert(CodeFlow codeFlow)
+        internal static List<CallTreeNode> Convert(CodeFlow codeFlow, Run run)
         {
             var root = new CallTreeNode { Children = new List<CallTreeNode>() };
             ThreadFlow threadFlow = codeFlow.ThreadFlows?[0];
@@ -22,6 +23,18 @@ namespace Microsoft.Sarif.Viewer.VisualStudio
 
                 foreach (ThreadFlowLocation location in threadFlow.Locations)
                 {
+                    ArtifactLocation artifactLocation = location.Location.PhysicalLocation?.ArtifactLocation;
+
+                    if (artifactLocation != null)
+                    {
+                        Uri uri = location.Location.PhysicalLocation?.ArtifactLocation?.Uri;
+
+                        if (uri == null && artifactLocation.Index > -1)
+                        {
+                            artifactLocation.Uri = run.Artifacts[artifactLocation.Index].Location.Uri;
+                        }
+                    }
+
                     var newNode = new CallTreeNode
                     {
                         Location = location,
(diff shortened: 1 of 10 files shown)


src/Sarif.Viewer.VisualStudio/CodeFlowToTreeConverter.cs
```diff
@@ -22,6 +23,18 @@ internal static List<CallTreeNode> Convert(CodeFlow codeFlow)
 
                 foreach (ThreadFlowLocation location in threadFlow.Locations)
                 {
+                    ArtifactLocation artifactLocation = location.Location.PhysicalLocation?.ArtifactLocation;
```
lgolding, line 26: >. [](start = 73, length = 1)

Do we need to conditionally access `,PhysicalLocation` as well? Is it possible that `Location` contains only a `LogicalLocation`?

michaelcfanning: For sure. Let's please capture this critical test asset scenario, which is a comprehensive result (i.e., one with a threadflow) that is entirely driven by logical locations, i.e., no physical location data available).

---
In reply to: [416794903](https://github.com/microsoft/sarif-visualstudio-extension/pull/134#discussion_r416794903) [](ancestors = 416794903)

src/Sarif.Viewer.VisualStudio/CodeFlowToTreeConverter.cs
```diff
@@ -22,6 +23,18 @@ internal static List<CallTreeNode> Convert(CodeFlow codeFlow)
...
+
+                    if (artifactLocation != null)
+                    {
+                        Uri uri = location.Location.PhysicalLocation?.ArtifactLocation?.Uri;
```
lgolding, line 30: >location.Location.PhysicalLocation?.ArtifactLocation [](start = 34, length = 52)

We already have `artifactLocation` in hand here so no need to re-compute it.

src/Sarif.Viewer.VisualStudio/CodeFlowToTreeConverter.cs
lgolding: Done. Added to [microsoft/tutorials#8](https://github.com/microsoft/sarif-tutorials/issues/8): "[ ] CodeFlow driven entirely by logical locations."

---
In reply to: [416796465](https://github.com/microsoft/sarif-visualstudio-extension/pull/134#discussion_r416796465) [](ancestors = 416796465,416794903)

PR #143, opened 2020-06-15
Expose the ability to close a set or all SARIF log files in the viewer

GabeDeBacker:
Exposes an API that allow consumers to close logs.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio.Interop/SarifViewerInterop.cs
+++ b/src/Sarif.Viewer.VisualStudio.Interop/SarifViewerInterop.cs
@@ -99,97 +99,96 @@ namespace Microsoft.Sarif.Viewer.Interop
         /// Opens the specified SARIF log file in the SARIF Viewer extension.
         /// </summary>
         /// <param name="path">The path of the log file.</param>
-        public async Task<bool> OpenSarifLogAsync(string path)
+        public Task<bool> OpenSarifLogAsync(string path)
         {
             if (string.IsNullOrWhiteSpace(path))
             {
                 throw new ArgumentNullException(nameof(path));
             }
 
-            bool result = false;
-
-            if (IsViewerExtensionInstalled && (IsViewerExtensionLoaded || LoadViewerExtension() != null))
+            return this.CallServiceApiAsync(ViewerLoadServiceInterfaceName, (service) =>
             {
-                // Get the service interface type
-                Type[] types = ViewerExtensionAssembly.GetTypes();
-                Type sarifLoadServiceInterface = types.Where(t => t.Name == ViewerLoadServiceInterfaceName).FirstOrDefault();
+                service.LoadSarifLogs(path);
+                return true;
+            });
+        }
 
-                if (sarifLoadServiceInterface != null)
-                {
-                    // Get a service reference
-                    dynamic sarifLoadService = await ServiceProvider.GetGlobalServiceAsync(sarifLoadServiceInterface);
+        /// <summary>
+        /// Loads the specified SARIF logs in the viewer.
+        /// </summary>
+        /// <param name="paths">The complete path to the SARIF log files.</param>
+        public Task<bool> OpenSarifLogAsync(IEnumerable<string> paths)
+        {
+            return this.OpenSarifLogAsync(paths, promptOnLogConversions: true);
+        }
 
-                    if (sarifLoadService != null)
-                    {
-                        // Call the service API
-                        sarifLoadService.LoadSarifLog(path);
-                        result = true;
-                    }
-                }
-            }
-
-            return result;
+        /// <summary>
+        /// Loads the specified SARIF logs in the viewer.
+        /// </summary>
+        /// <param name="paths">The complete path to the SARIF log files.</param>
+        /// <param name="promptOnLogConversions">Specifies whether the viewer should prompt if a SARIF log needs to be converted.</param>
+        /// <remarks>
+        /// Reasons for SARIF log file conversion include a conversion from a tool's log to SARIF, or a the SARIF schema version is not the latest version.
+        /// </remarks>
+        public Task<bool> OpenSarifLogAsync(IEnumerable<string> paths, bool promptOnLogConversions)
+        {
+            return this.CallServiceApiAsync(ViewerLoadServiceInterfaceName, (service) =>
+            {
+                service.LoadSarifLogs(paths, promptOnLogConversions);
+                return true;
+            });
         }
 
         /// <summary>
         /// Closes the specified SARIF log files in the SARIF Viewer extension.
         /// </summary>
         /// <param name="paths">The paths to the log files.</param>
-        public async Task<bool> CloseSarifLogAsync(IEnumerable<string> paths)
+        public Task<bool> CloseSarifLogAsync(IEnumerable<string> paths)
         {
-            bool result = false;
-
-            if (IsViewerExtensionInstalled && (IsViewerExtensionLoaded || LoadViewerExtension() != null))
+            return this.CallServiceApiAsync(ViewerCloseServiceInterfaceName, (service) =>
             {
-                // Get the service interface type
-                Type[] types = ViewerExtensionAssembly.GetTypes();
-                Type sarifCloseServiceInterface = types.Where(t => t.Name == ViewerCloseServiceInterfaceName).FirstOrDefault();
-
-                if (sarifCloseServiceInterface != null)
-                {
-                    // Get a service reference
-                    dynamic sarifLoadService = await ServiceProvider.GetGlobalServiceAsync(sarifCloseServiceInterface);
-
-                    if (sarifLoadService != null)
-                    {
-                        // Call the service API
-                        sarifLoadService.CloseSarifLog(paths);
-                        result = true;
-                    }
-                }
-            }
-
-            return result;
+                service.CloseSarifLog(paths);
+                return true;
+            });
         }
 
         /// <summary>
         /// Closes all SARIF logs opened in the viewer.
         /// </summary>
-        public async Task<bool> CloseAllSarifLogsAsync()
+        public Task<bool> CloseAllSarifLogsAsync()
         {
-            bool result = false;
-
-            if (IsViewerExtensionInstalled && (IsViewerExtensionLoaded || LoadViewerExtension() != null))
+            return this.CallServiceApiAsync(ViewerCloseServiceInterfaceName, (service) =>
             {
-                // Get the service interface type
-                Type[] types = ViewerExtensionAssembly.GetTypes();
-                Type sarifCloseServiceInterface = types.Where(t => t.Name == ViewerCloseServiceInterfaceName).FirstOrDefault();
+                service.CloseAllSarifLogs();
+                return true;
+            });
+        }
 
-                if (sarifCloseServiceInterface != null)
-                {
-                    // Get a service reference
-                    dynamic sarifLoadService = await ServiceProvider.GetGlobalServiceAsync(sarifCloseServiceInterface);
-
-                    if (sarifLoadService != null)
-                    {
-                        // Call the service API
-                        sarifLoadService.CloseAllSarifLogs();
-                        result = true;
-                    }
-                }
+        private async Task<bool> CallServiceApiAsync(string serviceInterfaceName, Func<dynamic, bool> action)
+        {
+            if (!IsViewerExtensionInstalled || (IsViewerExtensionLoaded && LoadViewerExtension() == null))
+            {
+                return false;
             }
 
-            return result;
+            // Get the service interface type
+            Type[] types = ViewerExtensionAssembly.GetTypes();
+            Type serviceType = types.Where(t => t.Name == serviceInterfaceName).FirstOrDefault();
+
+            if (serviceType == null)
+            {
+                return false;
+            }
+
+            // Get a service reference
+            dynamic serviceInterface = await ServiceProvider.GetGlobalServiceAsync(serviceType);
+
+            if (serviceInterface == null)
+            {
+                return false;
+            }
+
+            return action(serviceInterface);
         }
 
         /// <summary>
--- a/src/Sarif.Viewer.VisualStudio.Interop/SarifViewerInterop.cs
+++ b/src/Sarif.Viewer.VisualStudio.Interop/SarifViewerInterop.cs
@@ -175,7 +175,7 @@ namespace Microsoft.Sarif.Viewer.Interop
             Type[] types = ViewerExtensionAssembly.GetTypes();
             Type serviceType = types.Where(t => t.Name == serviceInterfaceName).FirstOrDefault();
 
-            if (serviceType == default)
+            if (serviceType == default(Type))
             {
                 return false;
             }
--- a/src/Sarif.Viewer.VisualStudio.Interop/SarifViewerInterop.cs
+++ b/src/Sarif.Viewer.VisualStudio.Interop/SarifViewerInterop.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using System.Threading.Tasks;
@@ -14,7 +15,8 @@ namespace Microsoft.Sarif.Viewer.Interop
     public class SarifViewerInterop
     {
         private static readonly string ViewerAssemblyFileName = "Microsoft.Sarif.Viewer";
-        private static readonly string ViewerServiceInterfaceName = "SLoadSarifLogService";
+        private static readonly string ViewerLoadServiceInterfaceName = "SLoadSarifLogService";
+        private static readonly string ViewerCloseServiceInterfaceName = "SCloseSarifLogService";
         private bool? _isViewerExtensionInstalled;
         private bool? _isViewerExtensionLoaded;
         private Assembly _viewerExtensionAssembly;
@@ -110,7 +112,7 @@ namespace Microsoft.Sarif.Viewer.Interop
             {
                 // Get the service interface type
                 Type[] types = ViewerExtensionAssembly.GetTypes();
-                Type sarifLoadServiceInterface = types.Where(t => t.Name == ViewerServiceInterfaceName).FirstOrDefault();
+                Type sarifLoadServiceInterface = types.Where(t => t.Name == ViewerLoadServiceInterfaceName).FirstOrDefault();
 
                 if (sarifLoadServiceInterface != null)
                 {
@@ -121,6 +123,67 @@ namespace Microsoft.Sarif.Viewer.Interop
                     {
                         // Call the service API
                         sarifLoadService.LoadSarifLog(path);
+                        result = true;
+                    }
+                }
+            }
+
+            return result;
+        }
+
+        /// <summary>
+        /// Closes the specified SARIF log files in the SARIF Viewer extension.
+        /// </summary>
+        /// <param name="paths">The paths to the log files.</param>
+        public async Task<bool> CloseSarifLogAsync(IEnumerable<string> paths)
+        {
+            bool result = false;
+
+            if (IsViewerExtensionInstalled && (IsViewerExtensionLoaded || LoadViewerExtension() != null))
+            {
+                // Get the service interface type
+                Type[] types = ViewerExtensionAssembly.GetTypes();
+                Type sarifCloseServiceInterface = types.Where(t => t.Name == ViewerCloseServiceInterfaceName).FirstOrDefault();
+
+                if (sarifCloseServiceInterface != null)
+                {
+                    // Get a service reference
+                    dynamic sarifLoadService = await ServiceProvider.GetGlobalServiceAsync(sarifCloseServiceInterface);
+
+                    if (sarifLoadService != null)
+                    {
+                        // Call the service API
+                        sarifLoadService.CloseSarifLog(paths);
+                        result = true;
+                    }
+                }
+            }
+
+            return result;
+        }
+
+        /// <summary>
+        /// Closes all SARIF logs opened in the viewer.
+        /// </summary>
+        public async Task<bool> CloseAllSarifLogsAsync()
+        {
+            bool result = false;
+
+            if (IsViewerExtensionInstalled && (IsViewerExtensionLoaded || LoadViewerExtension() != null))
+            {
+                // Get the service interface type
+                Type[] types = ViewerExtensionAssembly.GetTypes();
+                Type sarifCloseServiceInterface = types.Where(t => t.Name == ViewerCloseServiceInterfaceName).FirstOrDefault();
+
+                if (sarifCloseServiceInterface != null)
+                {
+                    // Get a service reference
+                    dynamic sarifLoadService = await ServiceProvider.GetGlobalServiceAsync(sarifCloseServiceInterface);
+
+                    if (sarifLoadService != null)
+                    {
+                        // Call the service API
+                        sarifLoadService.CloseAllSarifLogs();
                         result = true;
                     }
                 }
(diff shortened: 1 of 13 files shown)


src/Sarif.Viewer.VisualStudio/CloseSarifLogService.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        public void LoadSarifLog(string path)
+        {
+            if (!string.IsNullOrWhiteSpace(path))
+            {
+                try
+                {
+                    ErrorListService.ProcessLogFile(path, SarifViewerPackage.Dte.Solution, ToolFormat.None);
+                }
+                catch (InvalidCastException) { }
```
michaelcfanning, line 39: >InvalidCastException [](start = 23, length = 20)

Do you know under what circumstances this exception is raised? How did you know to add this handler? Code comment could be useful.

src/Sarif.Viewer.VisualStudio.Interop/SarifViewerInterop.cs
```diff
@@ -97,36 +99,96 @@ public bool IsViewerExtensionLoaded
...
+        /// Reasons for SARIF log file conversion include a conversion from a tool's log to SARIF, or a the SARIF schema version is not the latest version.
+        /// </remarks>
+        public Task<bool> OpenSarifLogAsync(IEnumerable<string> paths, bool promptOnLogConversions)
+        {
+            return this.CallServiceApiAsync(ViewerLoadServiceInterfaceName, (service) =>
             {
-                // Get the service interface type
-                Type[] types = ViewerExtensionAssembly.GetTypes();
-                Type sarifLoadServiceInterface = types.Where(t => t.Name == ViewerServiceInterfaceName).FirstOrDefault();
+                service.LoadSarifLogs(paths, promptOnLogConversions);
```
michaelcfanning, line 137: >service [](start = 16, length = 7)

These helpers don't provide for an error path, return 'false'. is that right? what happens when this handler raises an unhandled exception? does VS hang? no apparent IDE impact?

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -30,7 +30,7 @@ public class ErrorListService
     {
         public static readonly ErrorListService Instance = new ErrorListService();
 
-        public static void ProcessLogFile(string filePath, Solution solution, string toolFormat = ToolFormat.None)
+        public static void ProcessLogFile(string filePath, Solution solution, string toolFormat = ToolFormat.None, bool promptOnLogConversions = true)
```
michaelcfanning, line 33: >promptOnLogConversions [](start = 120, length = 22)

Nice addition.

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -51,7 +51,7 @@ public static void ProcessLogFile(string filePath, Solution solution, string too
                     {
                         // They're opening a v1 log, so we need to transform it.
                         // Ask if they'd like to save the v2 log.
-                        MessageDialogCommand response = PromptToSaveProcessedLog(Resources.TransformV1_DialogMessage);
+                        MessageDialogCommand response = promptOnLogConversions ? PromptToSaveProcessedLog(Resources.TransformV1_DialogMessage) : MessageDialogCommand.No;
```
michaelcfanning, line 54: >MessageDialogCommand [](start = 145, length = 20)

I agree this change in default is useful.

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -176,11 +176,28 @@ public static void ProcessLogFile(string filePath, Solution solution, string too
...
             SarifTableDataSource.Instance.BringToFront();
         }
 
+        /// <summary>
+        /// Closes the specified SARIF log in the viewer.
+        /// </summary>
+        /// <param name="paths">The complete path to the SARIF log file.</param>
+        public static void CloseSarifLogs(IEnumerable<string> paths)
+        {
+            SarifTableDataSource.Instance.CleanErrors(paths);
```
michaelcfanning, line 190: >CleanErrors [](start = 42, length = 11)

Another long overdue change. Nice.

src/Sarif.Viewer.VisualStudio/LoadSarifLogService.cs
```diff
@@ -2,22 +2,46 @@
...
+        public void LoadSarifLogs(IEnumerable<string> paths) => this.LoadSarifLogs(paths, promptOnSchemaUpgrade: false);
+
+        public void LoadSarifLogs(IEnumerable<string> paths, bool promptOnSchemaUpgrade)
+        {
+            if (!paths.Any())
+            {
+                return;
+            }
+
+            foreach (var path in paths)
```
michaelcfanning, line 42: >var [](start = 21, length = 3)

style nit, use 'string' here explicitly as opposed to 'var' keyword. Just for readability.

GabeDeBacker: Yep, we have a different style on our side, sorry for the bleed through. Fixed.

---
In reply to: [440862239](https://github.com/microsoft/sarif-visualstudio-extension/pull/143#discussion_r440862239) [](ancestors = 440862239)

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
GabeDeBacker: No problem.

---
In reply to: [440859935](https://github.com/microsoft/sarif-visualstudio-extension/pull/143#discussion_r440859935) [](ancestors = 440859935)

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
GabeDeBacker: Yep, works great.

---
In reply to: [440859508](https://github.com/microsoft/sarif-visualstudio-extension/pull/143#discussion_r440859508) [](ancestors = 440859508)

src/Sarif.Viewer.VisualStudio/CloseSarifLogService.cs
GabeDeBacker: This was a copy\paste on my part and the catch has been removed.

I removed it from here and the original location as well because it shouldn't be there at all (especially without a comment explaining why).

---
In reply to: [440857524](https://github.com/microsoft/sarif-visualstudio-extension/pull/143#discussion_r440857524) [](ancestors = 440857524)

src/Sarif.Viewer.VisualStudio.Interop/SarifViewerInterop.cs
GabeDeBacker: You are correct that the "Services" do not actually return any type of value. The original API was shipped this way so I did not want to introduce an public API change.

Since the service itself is internal and can be changed to return a value without changing the API surface, I'm am going to leave this code as is.

---
In reply to: [440858559](https://github.com/microsoft/sarif-visualstudio-extension/pull/143#discussion_r440858559) [](ancestors = 440858559)

GabeDeBacker:
@roywill is added to the review. #Closed

src/Sarif.Viewer.VisualStudio/CloseSarifLogService.cs
```diff
@@ -0,0 +1,41 @@
...
+using Microsoft.CodeAnalysis.Sarif.Converters;
+using Microsoft.Sarif.Viewer.ErrorList; 
+
+namespace Microsoft.Sarif.Viewer
+{
+    public class CloseSarifLogService : SCloseSarifLogService, ICloseSarifLogService
+    {
+        /// <summary>
+        /// Closes all SARIF logs opened in the viewer.
+        /// </summary>
```
roywill, line 16: Suggest replacing this with:
        /// <inheritdoc/>

src/Sarif.Viewer.VisualStudio/CloseSarifLogService.cs
```diff
@@ -0,0 +1,41 @@
...
+        public void CloseAllSarifLogs()
+        {
+            ErrorListService.CloseAllSarifLogs();
+        }
+
+        /// <summary>
+        /// Closes the specified SARIF log in the viewer.
+        /// </summary>
+        /// <param name="paths">The complete path to the SARIF log file.</param>
```
roywill, line 25: Same

src/Sarif.Viewer.VisualStudio/CloseSarifLogService.cs
```diff
@@ -0,0 +1,41 @@
...
+        public void CloseSarifLogs(IEnumerable<string> paths)
+        {
+            ErrorListService.CloseSarifLogs(paths);
+        }
+
+        public void LoadSarifLog(string path)
```
roywill, line 31: I don't understand why this is here on a Close object.  Also is this not part of ILoadSarifLogService?

src/Sarif.Viewer.VisualStudio/LoadSarifLogService.cs
```diff
@@ -2,22 +2,42 @@
...
+using System.Linq;
 using Microsoft.CodeAnalysis.Sarif.Converters;
 using Microsoft.Sarif.Viewer.ErrorList; 
 
 namespace Microsoft.Sarif.Viewer
 {
-    public class LoadSarifLogService : SLoadSarifLogService, ILoadSarifLogService
+    public class LoadSarifLogService : SLoadSarifLogService, ILoadSarifLogService, ILoadSarifLogService2
     {
+        public void LoadSarifLog(string path, bool promptOnSchemaUpgrade = true)
```
roywill, line 14: /// <inheritdoc/>

src/Sarif.Viewer.VisualStudio/LoadSarifLogService.cs
```diff
@@ -2,22 +2,42 @@
...
+        {
+            if (string.IsNullOrWhiteSpace(path))
+            {
+                return;
+            }
+
+            ErrorListService.ProcessLogFile(path, SarifViewerPackage.Dte.Solution, ToolFormat.None, promptOnSchemaUpgrade);
+        }
+
         public void LoadSarifLog(string path)
```
roywill, line 24: /// <inheritdoc/>

roywill:
>         public static IServiceProvider ServiceProvider;

I don't understand why this is class static and sent on each constructor instance of SarifViewerPackage.

---
Refers to: src/Sarif.Viewer.VisualStudio/SarifViewerPackage.cs:38 in 3046b16. [](commit_id = 3046b16bed5b0de73f3281d408a77b55fbeb9e2a, deletion_comment = False)

src/Sarif.Viewer.VisualStudio/SarifViewerPackage.cs
```diff
@@ -31,6 +31,7 @@ namespace Microsoft.Sarif.Viewer
     [ProvideEditorExtension(typeof(SarifEditorFactory), ".sarif", 128)]
     [ProvideToolWindow(typeof(SarifToolWindow), Style = VsDockStyle.Tabbed, Window = "3ae79031-e1bc-11d0-8f78-00a0c9110057")]
     [ProvideService(typeof(SLoadSarifLogService))]
+    [ProvideService(typeof(SCloseSarifLogService))]
     public sealed class SarifViewerPackage : AsyncPackage
     {
         public static DTE2 Dte;
```
roywill, line 37: Same comment as below

roywill:
>             catch (Exception)

This looks like a bug farm.  How about etw here?

---
Refers to: src/Sarif.Viewer.VisualStudio/SarifViewerPackage.cs:88 in 3046b16. [](commit_id = 3046b16bed5b0de73f3281d408a77b55fbeb9e2a, deletion_comment = False)

roywill:
>         /// Initialization of the package; this method is called right after the package is sited, so this is the place

This summary is Initialization of the package.  The rest should be in a remarks section

---
Refers to: src/Sarif.Viewer.VisualStudio/SarifViewerPackage.cs:98 in 3046b16. [](commit_id = 3046b16bed5b0de73f3281d408a77b55fbeb9e2a, deletion_comment = False)

src/Sarif.Viewer.VisualStudio/SCloseSarifLogService.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Sarif.Viewer
+{
+    internal interface SCloseSarifLogService
```
roywill, line 6: Why does this interface not start with I and how about a comment what it will be used for?

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -176,11 +176,28 @@ public static void ProcessLogFile(string filePath, Solution solution, string too
...
+            ProcessSarifLog(log, outputPath, solution, showMessageOnNoResults: promptOnLogConversions);
 
             SarifTableDataSource.Instance.BringToFront();
         }
 
+        /// <summary>
+        /// Closes the specified SARIF log in the viewer.
+        /// </summary>
+        /// <param name="paths">The complete path to the SARIF log file.</param>
+        public static void CloseSarifLogs(IEnumerable<string> paths)
```
roywill, line 188: Just a question.  Should this move to uri's?

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -176,11 +176,28 @@ public static void ProcessLogFile(string filePath, Solution solution, string too
...
+
+        /// <summary>
+        /// Closes all SARIF logs opened in the viewer.
+        /// </summary>
+        public static void CloseAllSarifLogs()
+        {
+            SarifTableDataSource.Instance.CleanAllErrors();
+        }
+
         private const string VersionRegexPattern = @"""version""\s*:\s*""(?<version>[\d.]+)""";
```
roywill, line 201: >VersionRegexPattern [](start = 29, length = 19)

This is not cool.  First this should be a Regex type and second it is compiled.  If compiled then it should be cached, but then again this causes a DLL to be created and never unloaded for something that is used what 1 time?  I suggest passing instead

CultureInvariant | ExplicitCapture | Singleline

roywill:
>             return Regex.Match(headSegment, VersionRegexPattern, RegexOptions.Compiled);

here

---
Refers to: src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs:208 in 3046b16. [](commit_id = 3046b16bed5b0de73f3281d408a77b55fbeb9e2a, deletion_comment = False)

roywill:
>             saveFileDialog.InitialDirectory = Path.GetDirectoryName(inputFilePath);

Suggest 
()
{
}

syntax.

---
Refers to: src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs:229 in 3046b16. [](commit_id = 3046b16bed5b0de73f3281d408a77b55fbeb9e2a, deletion_comment = False)

src/Sarif.Viewer.VisualStudio.Interop/SarifViewerInterop.cs
```diff
@@ -97,36 +99,96 @@ public bool IsViewerExtensionLoaded
...
+            if (!IsViewerExtensionInstalled || (IsViewerExtensionLoaded && LoadViewerExtension() == null))
+            {
+                return false;
+            }
+
+            // Get the service interface type
+            Type[] types = ViewerExtensionAssembly.GetTypes();
+            Type serviceType = types.Where(t => t.Name == serviceInterfaceName).FirstOrDefault();
+
+            if (serviceType == null)
```
roywill, line 178: If you use default instead it may be more intuitive that firstordefault failed.

src/Sarif.Viewer.VisualStudio/Sarif/Rule.extensions.cs
```diff
@@ -30,7 +30,7 @@ public static RuleModel ToRuleModel(this ReportingDescriptor rule, string defaul
                     DefaultFailureLevel = rule.DefaultConfiguration != null ?
                                     rule.DefaultConfiguration.Level :
                                     FailureLevel.Warning, // Default level
-                    HelpUri = rule.HelpUri?.AbsoluteUri
+                    HelpUri = rule.HelpUri?.IsAbsoluteUri == true ? rule.HelpUri?.AbsoluteUri : null
```
GabeDeBacker, line 33: >? [](start = 80, length = 1)

This ? operator does not need to exist.

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
GabeDeBacker: The existing APIs use strings, so I wanted to keep that pattern.

---
In reply to: [440942501](https://github.com/microsoft/sarif-visualstudio-extension/pull/143#discussion_r440942501) [](ancestors = 440942501)

src/Sarif.Viewer.VisualStudio/CloseSarifLogService.cs
GabeDeBacker: Copy paste error. Removed

---
In reply to: [440935382](https://github.com/microsoft/sarif-visualstudio-extension/pull/143#discussion_r440935382) [](ancestors = 440935382)

src/Sarif.Viewer.VisualStudio/SCloseSarifLogService.cs
GabeDeBacker: This defines the "service" that is used for various IServiceProvider (ish) techniques used in visual studio. It's a well established pattern.

---
In reply to: [440941253](https://github.com/microsoft/sarif-visualstudio-extension/pull/143#discussion_r440941253) [](ancestors = 440941253)

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
lgolding: This regex isn't used just once, it's evaluated every time you load a SARIF log file, so IMO it should be compiled once and cached:

```
private const string VersionRegexPattern = @"""version""\s*:\s*""(?<version>[\d.]+)""";
private static readonly s_versionRegex = new Regex(VersionRegexPattern, RegexOptions.Compiled | RegexOptions.CultureInvariant);
```

and then at the point of call:

```
return s_versionRegex.Match(headSegment);
```

`RegexOptions.SingleLine` (treat the input string as a single line by allowing `'.'` to match `'\n'`) isn't necessary because there are no `'.'`s in the pattern (the `'.'` in the character class is a literal `'.'`, not a metacharacter).

`Regex.ExplicitCapture` (don't capture unnamed groups) isn't necessary because there are no unnamed groups in the pattern.

---
In reply to: [440944742](https://github.com/microsoft/sarif-visualstudio-extension/pull/143#discussion_r440944742) [](ancestors = 440944742)

roywill:
Explicit capture means that the regex engine does not have to allocate for groups.

src/Sarif.Viewer.VisualStudio/Sarif/Rule.extensions.cs
```diff
@@ -30,7 +30,7 @@ public static RuleModel ToRuleModel(this ReportingDescriptor rule, string defaul
                     DefaultFailureLevel = rule.DefaultConfiguration != null ?
                                     rule.DefaultConfiguration.Level :
                                     FailureLevel.Warning, // Default level
-                    HelpUri = rule.HelpUri?.AbsoluteUri
+                    HelpUri = rule.HelpUri?.IsAbsoluteUri == true ? rule.HelpUri.AbsoluteUri : null
```
lgolding, line 33: Good change.

lgolding:
We're both half-right :-)

According to [https://docs.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-options#explicit-captures-only](https://docs.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-options#explicit-captures-only), `ExplicitCapture` only affects the behavior of un-named, un-numbered groups, that is, groups that are "...used only to apply quantifiers to multiple language elements, and the captured substrings are of no interest. ... Capturing groups that are not subsequently used can be expensive, because the regular expression engine must populate both the GroupCollection and CaptureCollection collection objects." This regex has exactly one group, and it is named, so `ExplicitCapture` would have no effect.

**BUT**! The `<version>` group is not actually _used_ by the caller, which I think is your point. So we should remove the group name `<version>`. If that's _all_ we did, we'd want to add `ExplicitCapture` to avoid allocating for the unnamed group. But in fact, even the unnamed group is unnecessary, because it's not quantified. We can dispense with the group altogether, and then once more `ExplicitCapture` is unnecessary:

```
private const string VersionRegexPattern = @"""version""\s*:\s*""[\d.]+""";
private static readonly s_versionRegex = new Regex(VersionRegexPattern, RegexOptions.Compiled | RegexOptions.CultureInvariant);
```

---
In reply to: [644879780](https://github.com/microsoft/sarif-visualstudio-extension/pull/143#issuecomment-644879780) [](ancestors = 644879780)

lgolding:
Nope, I'm wrong about the `version` group. It _is_ used, both in the product code and in the tests. So I'm back to my original position: there is only one group in this regex, and it's named, so `ExplicitCapture` has no effect.

---
In reply to: [644889454](https://github.com/microsoft/sarif-visualstudio-extension/pull/143#issuecomment-644889454) [](ancestors = 644889454,644879780)

lgolding:
I filed [https://github.com/microsoft/sarif-visualstudio-extension/issues/144](https://github.com/microsoft/sarif-visualstudio-extension/issues/144), "Improve treatment of version regex".

---
In reply to: [644891893](https://github.com/microsoft/sarif-visualstudio-extension/pull/143#issuecomment-644891893) [](ancestors = 644891893,644889454,644879780)

lgolding:
>         /// Initialization of the package; this method is called right after the package is sited, so this is the place

I filed [https://github.com/microsoft/sarif-visualstudio-extension/issues/145](https://github.com/microsoft/sarif-visualstudio-extension/issues/145), "Improve organization of doc comment."

---
In reply to: [644836186](https://github.com/microsoft/sarif-visualstudio-extension/pull/143#issuecomment-644836186) [](ancestors = 644836186)

---
Refers to: src/Sarif.Viewer.VisualStudio/SarifViewerPackage.cs:98 in 3046b16. [](commit_id = 3046b16bed5b0de73f3281d408a77b55fbeb9e2a, deletion_comment = False)

lgolding:
>             saveFileDialog.InitialDirectory = Path.GetDirectoryName(inputFilePath);

I filed [https://github.com/microsoft/sarif-visualstudio-extension/issues/146](https://github.com/microsoft/sarif-visualstudio-extension/issues/146), "Use object initializer syntax".

---
In reply to: [644840701](https://github.com/microsoft/sarif-visualstudio-extension/pull/143#issuecomment-644840701) [](ancestors = 644840701)

---
Refers to: src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs:229 in 3046b16. [](commit_id = 3046b16bed5b0de73f3281d408a77b55fbeb9e2a, deletion_comment = False)

lgolding:
>             catch (Exception)

I filed [https://github.com/microsoft/sarif-visualstudio-extension/issues/147](https://github.com/microsoft/sarif-visualstudio-extension/issues/147), "Don't just ignore exceptions from GetService."

---
In reply to: [644835816](https://github.com/microsoft/sarif-visualstudio-extension/pull/143#issuecomment-644835816) [](ancestors = 644835816)

---
Refers to: src/Sarif.Viewer.VisualStudio/SarifViewerPackage.cs:88 in 3046b16. [](commit_id = 3046b16bed5b0de73f3281d408a77b55fbeb9e2a, deletion_comment = False)

lgolding:
>         public static IServiceProvider ServiceProvider;

They are static because they are used in places that don't have access to the `SarifViewerPackage` instance. So for example, in `LoadSarifLogService.LoadSarifLog`, we have:

```
ErrorListService.ProcessLogFilePath(path, SarifViewerPackage.Dte.Solution, ToolFormat.None);
```

---
In reply to: [644834157](https://github.com/microsoft/sarif-visualstudio-extension/pull/143#issuecomment-644834157) [](ancestors = 644834157)

---
Refers to: src/Sarif.Viewer.VisualStudio/SarifViewerPackage.cs:38 in 3046b16. [](commit_id = 3046b16bed5b0de73f3281d408a77b55fbeb9e2a, deletion_comment = False)

Issue #148, opened 2020-06-16
FixModel_ApplyFix unit-test fails unless test deployment is clean


GabeDeBacker:
In the last line of the unit-tests, this Should().Be() fails but only on secondary runs of the test
```csharp
        public void FixModel_ApplyFix()
        {
           ...
            offsets[233].Should().Be(0);
        }
```

PR #155, opened 2020-07-22
Allow for text fixes. Bump version.

### LINKED ISSUE #154: VS fix preview crashes if replacement content uses Text rather than Binary content.

michaelcfanning:
@lgolding

This fixes #154. I did not provide a test because this test area needs more work per #148. We need a spot-fix on this asap to unblock a key partner and will return to that technical debt later.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/Models/FixModel.cs
+++ b/src/Sarif.Viewer.VisualStudio/Models/FixModel.cs
@@ -6,6 +6,8 @@ using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.IO;
 using System.Linq;
+using System.Text;
+
 using Microsoft.CodeAnalysis.Sarif;
 
 namespace Microsoft.Sarif.Viewer.Models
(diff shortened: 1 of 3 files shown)


src/Sarif.Viewer.VisualStudio/Models/FixModel.cs
```diff
@@ -237,11 +239,17 @@ internal bool TryFixFile(string filePath, IEnumerable<ReplacementModel> replacem
...
 
-                    if (replacement.InsertedBytes.Length > 0)
+                    if (replacement.InsertedBytes?.Length > 0)
                     {
                         bytes.InsertRange(offset, replacement.InsertedBytes);
                         thisDelta += replacement.InsertedBytes.Length;
                     }
+                    else if (replacement.InsertedString?.Length > 0)
+                    {
+                        byte[] toInsert = Encoding.UTF8.GetBytes(replacement.InsertedString);
```
lgolding, line 249: >UTF8 [](start = 51, length = 4)

This assumes that the file being fixed is encoded in UTF-8. Is that ok?

Issue #163, opened 2020-08-14
LoadSarifLogs API implementation only shows result from last log file


GabeDeBacker:
The LoadSarifLogs API enumerates through a list of SARIF log paths passed to it.

It currently calls the LoadSarifLog API to do the work of loading each individual SARIF log.

The implementation of LoadSarifLog causes any existing SARIF results to be cleared, so only the last path passed to LoadSarifLogs will display any results.

Miaku:
PR is ready to close out on this. 

ETA to PR completion: EoD Today.

GabeDeBacker:
Verified fixed this morning.

PR #164, opened 2020-08-14
Add option for cleaning results to prevent clearing

GabeDeBacker:
This addresses issue #163.

The issue was simply that the LoadSarifLogs implementation was relying on LoadSarifLog which always cleared the error list.

LoadSarifLogs does not leverage LoadSarifLog to process the logs in a loop, and an option as added to the error service indicating whether to clear previous errors when it processes a log.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/LoadSarifLogService.cs
+++ b/src/Sarif.Viewer.VisualStudio/LoadSarifLogService.cs
@@ -34,19 +34,9 @@ namespace Microsoft.Sarif.Viewer
         /// <inheritdoc/>
         public void LoadSarifLogs(IEnumerable<string> paths, bool promptOnSchemaUpgrade)
         {
-            if (!paths.Any())
-            {
-                return;
-            }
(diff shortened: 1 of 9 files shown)


src/Sarif.Viewer.VisualStudio/LoadSarifLogService.cs
```diff
@@ -39,9 +39,16 @@ public void LoadSarifLogs(IEnumerable<string> paths, bool promptOnSchemaUpgrade)
                 return;
             }
 
+            var cleanErrors = true;
             foreach (string path in paths)
             {
-                this.LoadSarifLog(path, promptOnSchemaUpgrade);
+                if (string.IsNullOrWhiteSpace(path))
```
roywill, line 45: You could make this part of the filter by using a where

GabeDeBacker: Yep, I'll make this change

Issue #174, opened 2020-08-20
On a fresh open of Visual Studio after reboot, SARIF Viewer is the default Explorer that is opened. I believe we want this to be the Solution Explorer


HenokAddis:
![image](https://user-images.githubusercontent.com/46505299/90806347-ba4ef380-e2d1-11ea-980c-a8a9cff3f8b9.png)

GabeDeBacker:
Fixed in 2.1.12

Issue #170, opened 2020-08-18
Unable to open SARIF files through Open -> File 


HenokAddis:
If you open multiple SARIF files and a normal text file, you will be reproducing the issue:
![image](https://user-images.githubusercontent.com/46505299/90568410-9f10a680-e160-11ea-925f-84c0f4892e63.png)

Only the _objects.mac file opened...where are sarifs? They do not open the actual SARIF files which is the obvious expectation if I open the file itself. We cannot see the content in VS at all. 
![image](https://user-images.githubusercontent.com/46505299/90568513-e434d880-e160-11ea-8aea-2072c20381ab.png)

GabeDeBacker:
This is also inconsistent with the VSCode implementation.

Issue #169, opened 2020-08-18
Unable to open SARIF file from Explorer or manage/clear/filter from the Explorer


HenokAddis:
When you open a SARIF file, you cannot look at the SARIF generated content in the log file, they just populate the ErrorList.
![image](https://user-images.githubusercontent.com/46505299/90568627-24945680-e161-11ea-94b9-90886f91b2be.png)

Issue #168, opened 2020-08-18
When you open multiple SARIF files in Visual Studio (not BUILD a directory that creates multiple sarif files) it only takes the most recent log file that was parsed and clears all the rest


HenokAddis:
Every time they tell VS theres a new SARIF to process, they clear out the SARIF explorer... so multi-select does not work at all just the last sarif file. If the last sarif has no results then you see nothing.

![image](https://user-images.githubusercontent.com/46505299/90569128-214d9a80-e162-11ea-9347-e742e483d7d5.png)

![image](https://user-images.githubusercontent.com/46505299/90569100-15fa6f00-e162-11ea-9cb2-f4cae39707ef.png)

PR #179, opened 2020-08-24
Handle multi-file selection from VS's open file dialog

GabeDeBacker:
This addresses issues #168, #169, #170 .

Currently there is an inconsistency between the SARIF viewer's implementation in Visual Studio Code and Visual Studio with respect to what happens if a user directly opens a SARIF log file through the file open dialog (or drag and drop from the Windows Explorer), etc.

In Visual Studio Code, what happens is that the file is presented to the user in an editor, and the SARIF explorer populates with any errors present in the SARIF log.

In Visual Studio, no file is presented and only the error list populates. In order to open the error log, the user must navigate to an error in the error list, double click on it, and then scroll down to the very bottom of the details in click on the log file the error originated from. (In the case of Prefast running on the Windows OS source code, the command line in the details pane is huge, so there is a lot of scrolling involved).

Further more, before this change, each log file is processed individual from VS's open file dialog, and any previous errors where cleared from the list. The result is that you only see the results from the last log file that VS's file open dialog sends the SARIF viewer.

This change allows Visual Studio to create an editor viewer for the SARIF log file and when a text-view is created, the error list is populated. When the text view is closed, the errors are cleared from the error list.

Other very notable changes are the removal of static instance variables in the SARIF viewer package class. These are very dangerous when MEF composition is used because exported MEF components (such as a text view creation listeners) can be created before the package is loaded. This also lead to the discover that there were a lot of calls being made that were not properly being asserted as being on VS's UI thread. This is all goodness as it leaves "markers" as fixes that need to occur to address issue #160  (hanging Visual Studio's UI thread).

Finally, it contains a quick fix for #174


src/Sarif.Viewer.VisualStudio/SarifViewerPackage.cs
```diff
@@ -28,21 +31,19 @@ namespace Microsoft.Sarif.Viewer
...
+    [ProvideToolWindow(typeof(SarifToolWindow), Style = VsDockStyle.Tabbed, Window = "3ae79031-e1bc-11d0-8f78-00a0c9110057", Transient = true)]
     [ProvideService(typeof(SLoadSarifLogService))]
     [ProvideService(typeof(SCloseSarifLogService))]
     public sealed class SarifViewerPackage : AsyncPackage
     {
-        public static DTE2 Dte;
-        public static IServiceProvider ServiceProvider;
-
-        private SarifEditorFactory _sarifEditorFactory;
+        private DTE2 Dte;
```
GabeDeBacker, line 39: Is this used anymore?

GabeDeBacker: Nope. Neither one actually. Which is goodness. Removing both.

---
In reply to: [475869907](https://github.com/microsoft/sarif-visualstudio-extension/pull/179#discussion_r475869907) [](ancestors = 475869907)

src/Sarif.Viewer.VisualStudio/SarifViewerPackage.cs
```diff
@@ -67,11 +68,18 @@ public static SarifToolWindow SarifToolWindow
...
-                SarifToolWindow toolWindow = package?.FindToolWindow(typeof(SarifToolWindow), 0, true) as SarifToolWindow;
-
-                return toolWindow;
+                ThreadHelper.ThrowIfNotOnUIThread();
+                IVsPackage package;
+                IVsShell vsShell = Microsoft.VisualStudio.Shell.ServiceProvider.GlobalProvider.GetService(typeof(SVsShell)) as IVsShell;
+                if (vsShell == null ||
+                    (vsShell.IsPackageLoaded(PackageGuid, out package) != VSConstants.S_OK &&
+                    vsShell.LoadPackage(PackageGuid, out package) != VSConstants.S_OK) ||
+                    !(package is Package vsPackage))
```
roywill, line 77: Weird

src/Sarif.Viewer.VisualStudio/SarifViewerPackage.cs
```diff
@@ -67,11 +58,26 @@ public static SarifToolWindow SarifToolWindow
...
-                return toolWindow;
+                ThreadHelper.ThrowIfNotOnUIThread();
+                IVsPackage package;
+                IVsShell vsShell = ServiceProvider.GlobalProvider.GetService(typeof(SVsShell)) as IVsShell;
+                if (vsShell == null)
+                {
+                    return null;
+                }
+                   
+                if (vsShell.IsPackageLoaded(PackageGuid, out package) != VSConstants.S_OK &&
```
sherryyshi, line 69: >&& [](start = 90, length = 2)

should this be a `||`?

src/Sarif.Viewer.VisualStudio/ContentTypes/SarifContentType.cs
```diff
@@ -0,0 +1,29 @@
+﻿namespace Microsoft.Sarif.Viewer.ContentTypes
+{
+    using Microsoft.VisualStudio.Utilities;
+    using System.ComponentModel.Composition;
+
+    public static class SarifContentType
+    {
+        /// <summary>
+        /// The content type name for Windows.NMake.
```
sherryyshi, line 9: >Windows.NMake. [](start = 38, length = 14)

update?

src/Sarif.Viewer.VisualStudio/ContentTypes/SarifContentType.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// </summary>
+        public const string ContentTypeName = "SARIF";
+
+        /// <summary>
+        /// Gets the base content type definition for Windows.NMake.
```
sherryyshi, line 14: > Windows.NMake. [](start = 53, length = 15)

update?

src/Sarif.Viewer.VisualStudio/SarifViewerPackage.cs
```diff
@@ -67,11 +58,26 @@ public static SarifToolWindow SarifToolWindow
         {
             get
             {
-                SarifViewerPackage package = SarifViewerPackage.ServiceProvider as SarifViewerPackage;
-
-                SarifToolWindow toolWindow = package?.FindToolWindow(typeof(SarifToolWindow), 0, true) as SarifToolWindow;
-
-                return toolWindow;
+                ThreadHelper.ThrowIfNotOnUIThread();
+                IVsPackage package;
```
lgolding, line 62: >IVsPackage [](start = 16, length = 10)

Nit/suggestion: Move declaration down to right above point of use (or even collapse it into the `IsPackageLoaded` call (`out IVsPackage package`).

src/Sarif.Viewer.VisualStudio/SarifTextViewCreationListener.cs
```diff
@@ -0,0 +1,107 @@
...
+            {
+                return false;
+            }
+
+            if (!(vsTextLines is IPersistFileFormat persistFile))
+            {
+                return false;
+            }
+
+            return persistFile.GetCurFile(out fileName, out var _) == VSConstants.S_OK;
```
lgolding, line 102: >var [](start = 60, length = 3)

You don't even need to say `var` for a "discard" parameter. `out _` suffices.

Issue #185, opened 2020-08-28
Performance improvements needed for tag/highlighting


GabeDeBacker:
After speaking with an VS Developer who works on the editor, he made several comments about the viewers implementation.

Should be using IPersistentSpanFactory to create tracking spans as those will survive the user editing, saving, closing and re-opening a file.

Should only be listening to caret notifications in as few places as possible (today ever tag\highlight present in every result listens to to the caret changes which is very inefficient)

Should implement our own tagger rather than relying on the "Simple Tagger" so that the highlight tags do not disappear if a user does and edit that causes a tag\highlight and then undoes that edit. (Today the highlights disappear due to how the simple tagger works). The tagger can be associated with the textview's buffer.

The tagger will be aware of the "persistent tags" and can handle caret changes quickly and more efficiently.

Issue #183, opened 2020-08-26
Cannot edit code that has highlights around it


GabeDeBacker:
When the SARIF viewer adds code markers and highlights, they cannot be edited anymore.

Issue #182, opened 2020-08-26
Code highlights (markers) are not removed when errors are cleared\fixed


GabeDeBacker:
	Closing and reopening the file doesn’t work.
	Clearing the error list doesn’t work.
	Rebuild doesn’t clear the highlights on fixed errors (!!!)

PR #191, opened 2020-09-01
More efficient tagging and caret watching for VS's editor

GabeDeBacker:
This address issues #185 and #183.

We've had reports that the performance of Visual Studio's editor was sluggish after selecting an SARIF result from the error list and having the SARIF pane (tool window) visible. 

I would go beyond that and say it wasn't usable at all :) You couldn't actually get the caret in the editor inside any highlighted text that the SARIF extension highlighted.

All of this was due to an event loop that didn't end up in a stack overflow, but a never ending "go here, go there, select this, select that" events being queued to the VS UI thread to handle. Result, super super slow VS performance.

I worked with a Visual Studio developer who specifically works on the editor to go through this issue and make sure that my changes where appropriate. After this change, the editor is usable and it solve this immediate issue.

Notable changes:

Using "persistent span factories" to create the tags/highlights so that if the user closes and re-opens a file with issues after making edits the tags/highlights remain intact.
Instead of each tag listening to a text view's caret change notification, the tagger responsible for the tags/highlights has one listener per view. Thereby increasing the performance of moving around the editor. The event firing also does not continue repeatedly if the user moves the caret within the same tag/highlight.
Highlights are properly removed when an issue is "cleared" either via the interop APIs or the error list UI.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs
+++ b/src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs
@@ -208,28 +208,6 @@ namespace Microsoft.Sarif.Viewer
             LineMarker?.AddHighlightMarker(SelectedSourceHighlightColor);
         }
 
-        private IVsTextView GetTextViewFromFrame(IVsWindowFrame frame)
-        {
-            ThreadHelper.ThrowIfNotOnUIThread();
-            // Get the document view from the window frame, then get the text view
-            object docView;
-            int hr = frame.GetProperty((int)__VSFPROPID.VSFPROPID_DocView, out docView);
-            if ((hr != 0 && hr != 1) || docView == null)
-            {
-                return null;
-            }
-
-            IVsCodeWindow codeWindow = docView as IVsCodeWindow;
-            IVsTextView textView;
-            codeWindow.GetLastActiveView(out textView);
-            if (textView == null)
-            {
-                codeWindow.GetPrimaryView(out textView);
-            }
-
-            return textView;
-        }
-
         /// <summary>
         /// An overridden method for reacting to the event of a document window
         /// being opened
@@ -239,7 +217,7 @@ namespace Microsoft.Sarif.Viewer
             ThreadHelper.ThrowIfNotOnUIThread();
 
             // For these cases, this event has nothing to do with this item
-            if (frame == null || LineMarker.IsTracking(docCookie) || string.Compare(documentName, FilePath, StringComparison.OrdinalIgnoreCase) != 0)
+            if (frame == null || LineMarker.IsTracking || string.Compare(documentName, FilePath, StringComparison.OrdinalIgnoreCase) != 0)
             {
                 return;
             }
@@ -267,12 +245,12 @@ namespace Microsoft.Sarif.Viewer
                 line = 1;
             }
 
-            IVsTextView textView = GetTextViewFromFrame(frame);
-            if (textView != null)
+            IVsTextView vsTextView = SdkUIUtilities.GetTextViewFromFrame(frame);
+            if (vsTextView != null)
             {
                 // Locate the specific line/column position in the text view and go there
                 IVsTextLines textLines;
-                textView.GetBuffer(out textLines);
+                vsTextView.GetBuffer(out textLines);
                 if (textLines != null)
                 {
                     int lastLine;
@@ -296,47 +274,23 @@ namespace Microsoft.Sarif.Viewer
 
                 // Call a bunch of functions to get the WPF text view so we can perform the highlighting only
                 // if we haven't yet
-                IWpfTextView wpfTextView = GetWpfTextView(textView);
+                IWpfTextView wpfTextView = SdkUIUtilities.GetWpfTextView(vsTextView);
                 if (wpfTextView != null)
                 {
-                    AttachMarkerToTextView(wpfTextView, docCookie, marker,
+                    AttachMarkerToTextView(frame, wpfTextView, docCookie, marker,
                         line, sourceLocation.StartColumn, line + (sourceLocation.EndLine - sourceLocation.StartLine), sourceLocation.EndColumn);
                 }
             }
         }
 
         /// <summary>
-        /// Helper method for getting a IWpfTextView from a IVsTextView object
-        /// </summary>
-        /// <param name="textView"></param>
-        /// <returns></returns>
-        private IWpfTextView GetWpfTextView(IVsTextView textView)
-        {
-            IWpfTextViewHost textViewHost = null;
-            IVsUserData userData = textView as IVsUserData;
-            if (userData != null)
-            {
-                Guid guid = Microsoft.VisualStudio.Editor.DefGuidList.guidIWpfTextViewHost;
-                object wpfTextViewHost = null;
-                userData.GetData(ref guid, out wpfTextViewHost);
-                textViewHost = wpfTextViewHost as IWpfTextViewHost;
-            }
-
-            if (textViewHost == null)
-            {
-                return null;
-            }
-            return textViewHost.TextView;
-        }
-
-        /// <summary>
         /// Highlight the source code on a particular line
         /// </summary>
-        private static void AttachMarkerToTextView(IWpfTextView textView, long docCookie, ResultTextMarker marker,
+        private static void AttachMarkerToTextView(IVsWindowFrame vsWindowFrame, IWpfTextView wpfTextView, long docCookie, ResultTextMarker marker,
             int line, int column, int endLine, int endColumn)
         {
             // If for some reason the start line is not correct, just skip the highlighting
-            ITextSnapshot textSnapshot = textView.TextSnapshot;
+            ITextSnapshot textSnapshot = wpfTextView.TextSnapshot;
             if (line > textSnapshot.LineCount)
             {
                 return;
@@ -365,7 +319,7 @@ namespace Microsoft.Sarif.Viewer
                     coerced = true;
                 }
 
-                // Calculate the end marker bound. Perform coersion on the values if they aren't consistent
+                // Calculate the end marker bound. Perform coercion on the values if they aren't consistent
                 if (endColumn <= 0 && endColumn >= endTextLine.Length)
                 {
                     endColumn = endTextLine.Length;
@@ -386,7 +340,7 @@ namespace Microsoft.Sarif.Viewer
                 markerEnd = endTextLine.Start.Position + endColumn;
                 spanToColor = Span.FromBounds(markerStart, markerEnd);
 
-                marker.AddTracking(textView, textSnapshot, docCookie, spanToColor);
+                marker.AddTracking(vsWindowFrame, wpfTextView, spanToColor);
             }
             catch (Exception e)
             {
--- a/src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs
+++ b/src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs
@@ -35,17 +35,17 @@ namespace Microsoft.Sarif.Viewer
         {
             get
             {
-                return _region;
+                return this._region;
             }
             set
             {
-                if (value != _region)
+                if (value != this._region)
                 {
                     _region = value;
 
-                    if (LineMarker != null)
+                    if (this._lineMarker != null)
                     {
-                        LineMarker.Region = _region;
+                        this._lineMarker.Region = _region;
                     }
 
                     NotifyPropertyChanged("Region");
@@ -68,17 +68,17 @@ namespace Microsoft.Sarif.Viewer
         {
             get
             {
-                return _filePath;
+                return this._filePath;
             }
             set
             {
-                if (value != _filePath)
+                if (value != this._filePath)
                 {
-                    _filePath = value;
+                    this._filePath = value;
 
-                    if (this.LineMarker != null)
+                    if (this._lineMarker != null)
                     {
-                        this.LineMarker.FullFilePath = _filePath;
+                        this._lineMarker.FullFilePath = this._filePath;
                     }
 
                     NotifyPropertyChanged("FilePath");
@@ -90,17 +90,17 @@ namespace Microsoft.Sarif.Viewer
         {
             get
             {
-                return _uriBaseId;
+                return this._uriBaseId;
             }
             set
             {
-                if (value != _uriBaseId)
+                if (value != this._uriBaseId)
                 {
-                    _uriBaseId = value;
+                    this._uriBaseId = value;
 
-                    if (this.LineMarker != null)
+                    if (this._lineMarker != null)
                     {
-                        this.LineMarker.UriBaseId = _uriBaseId;
+                        this._lineMarker.UriBaseId = this._uriBaseId;
                     }
 
                     NotifyPropertyChanged("UriBaseId");
--- a/src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs
+++ b/src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs
@@ -3,14 +3,10 @@
 
 using System;
 using System.ComponentModel;
-using System.Diagnostics;
 using System.IO;
 using Microsoft.CodeAnalysis.Sarif;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
-using Microsoft.VisualStudio.Text;
-using Microsoft.VisualStudio.Text.Editor;
-using Microsoft.VisualStudio.TextManager.Interop;
 
 namespace Microsoft.Sarif.Viewer
 {
@@ -32,10 +28,6 @@ namespace Microsoft.Sarif.Viewer
                 }
 
                 return _lineMarker;
-            }
-            set
-            {
-                _lineMarker = value;
             }
         }
 
@@ -216,138 +208,7 @@ namespace Microsoft.Sarif.Viewer
         {
             ThreadHelper.ThrowIfNotOnUIThread();
 
-            // For these cases, this event has nothing to do with this item
-            if (frame == null || LineMarker.IsTracking || string.Compare(documentName, FilePath, StringComparison.OrdinalIgnoreCase) != 0)
-            {
-                return;
-            }
-
-            AttachToDocumentWorker(frame, docCookie, LineMarker);
-        }
-
-        /// <summary>
-        /// Check that current <paramref name="marker"/> point to correct line position 
-        /// and attach it to <paramref name="docCookie"/> for track changes.
-        /// </summary>
-        private void AttachToDocumentWorker(IVsWindowFrame frame, long docCookie, ResultTextMarker marker)
-        {
-            ThreadHelper.ThrowIfNotOnUIThread();
-
-            var sourceLocation = marker.GetSourceLocation();
-            int line = sourceLocation.StartLine;
-
-            // Coerce the line numbers so we don't go out of bound. However, if we have to
-            // coerce the line numbers, then we won't perform highlighting because most likely
-            // we will highlight the wrong line. The idea here is to just go to the top or bottom
-            // of the file as our "best effort" to be closest where it thinks it should be
-            if (line <= 0)
-            {
-                line = 1;
-            }
-
-            IVsTextView vsTextView = SdkUIUtilities.GetTextViewFromFrame(frame);
-            if (vsTextView != null)
-            {
-                // Locate the specific line/column position in the text view and go there
-                IVsTextLines textLines;
-                vsTextView.GetBuffer(out textLines);
-                if (textLines != null)
-                {
-                    int lastLine;
-                    int length;
-                    int hr = textLines.GetLastLineIndex(out lastLine, out length);
-                    if (hr != 0)
-                    {
-                        return;
-                    }
-
-                    // our source code lines are 1-based, and the VS API source code lines are 0-based
-
-                    lastLine = lastLine + 1;
-
-                    // Same thing here, coerce the line number if it's going out of bound
-                    if (line > lastLine)
-                    {
-                        line = lastLine;
-                    }
-                }
-
-                // Call a bunch of functions to get the WPF text view so we can perform the highlighting only
-                // if we haven't yet
-                IWpfTextView wpfTextView = SdkUIUtilities.GetWpfTextView(vsTextView);
-                if (wpfTextView != null)
-                {
-                    AttachMarkerToTextView(frame, wpfTextView, docCookie, marker,
-                        line, sourceLocation.StartColumn, line + (sourceLocation.EndLine - sourceLocation.StartLine), sourceLocation.EndColumn);
-                }
-            }
-        }
-
-        /// <summary>
-        /// Highlight the source code on a particular line
-        /// </summary>
-        private static void AttachMarkerToTextView(IVsWindowFrame vsWindowFrame, IWpfTextView wpfTextView, long docCookie, ResultTextMarker marker,
-            int line, int column, int endLine, int endColumn)
-        {
-            // If for some reason the start line is not correct, just skip the highlighting
-            ITextSnapshot textSnapshot = wpfTextView.TextSnapshot;
-            if (line > textSnapshot.LineCount)
-            {
-                return;
-            }
-
-            Span spanToColor;
-            int markerStart, markerEnd = 0;
-
-            try
-            {
-                // Fix up the end line number if it's inconsistent
-                if (endLine <= 0 || endLine < line)
-                {
-                    endLine = line;
-                }
-
-                bool coerced = false;
-
-                // Calculate the start and end marker bound. Adjust for the column values if
-                // the values don't make sense. Make sure we handle the case of empty file correctly
-                ITextSnapshotLine startTextLine = textSnapshot.GetLineFromLineNumber(Math.Max(line - 1, 0));
-                ITextSnapshotLine endTextLine = textSnapshot.GetLineFromLineNumber(Math.Max(endLine - 1, 0));
-                if (column <= 0 || column >= startTextLine.Length)
-                {
-                    column = 1;
-                    coerced = true;
-                }
-
-                // Calculate the end marker bound. Perform coercion on the values if they aren't consistent
-                if (endColumn <= 0 && endColumn >= endTextLine.Length)
-                {
-                    endColumn = endTextLine.Length;
-                    coerced = true;
-                }
-
-                // If we are highlighting just one line and the column values don't make
-                // sense or we corrected one or more of them, then simply mark the
-                // entire line
-                if (endLine == line && (coerced || column >= endColumn))
-                {
-                    column = 1;
-                    endColumn = endTextLine.Length;
-                }
-
-                // Create a span with the calculated markers
-                markerStart = startTextLine.Start.Position + column - 1;
-                markerEnd = endTextLine.Start.Position + endColumn;
-                spanToColor = Span.FromBounds(markerStart, markerEnd);
-
-                marker.AddTracking(vsWindowFrame, wpfTextView, spanToColor);
-            }
-            catch (Exception e)
-            {
-                // Log the exception and move ahead. We don't want to bubble this or fail.
-                // We just don't color the problem line.
-                Debug.Print(e.Message);
-            }
+            LineMarker.TryAttachToDocument(documentName, frame);
         }
     }
 }
--- a/src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs
+++ b/src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs
@@ -152,7 +152,7 @@ namespace Microsoft.Sarif.Viewer
 
             if (LineMarker != null)
             {
-                LineMarker?.NavigateTo(usePreviewPane);
+                LineMarker?.TryNavigateTo(usePreviewPane);
             }
             else
             {
--- a/src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs
+++ b/src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs
@@ -208,7 +208,7 @@ namespace Microsoft.Sarif.Viewer
         {
             ThreadHelper.ThrowIfNotOnUIThread();
 
-            LineMarker.TryAttachToDocument(documentName, frame);
+            LineMarker.TryTagDocument(documentName, frame);
         }
     }
 }
--- a/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
+++ b/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
@@ -123,7 +123,7 @@ namespace Microsoft.Sarif.Viewer.ErrorList
                 {
                     // The version property wasn't found within the first 100 characters.
                     // Per the spec, it should appear first in the sarifLog object.
-                    VsShellUtilities.ShowMessageBox(Microsoft.VisualStudio.Shell.ServiceProvider.GlobalProvider,
+                    VsShellUtilities.ShowMessageBox(ServiceProvider.GlobalProvider,
                                                     Resources.VersionPropertyNotFound_DialogTitle,
                                                     null, // title
                                                     OLEMSGICON.OLEMSGICON_QUERY,
@@ -205,7 +205,7 @@ namespace Microsoft.Sarif.Viewer.ErrorList
         /// </summary>
         public static void CloseAllSarifLogs()
         {
-            SarifTableDataSource.Instance.CleanAllErrors();
+            CleanAllErrors();
         }
 
         private const string VersionRegexPattern = @"""version""\s*:\s*""(?<version>[\d.]+)""";
@@ -226,7 +226,7 @@ namespace Microsoft.Sarif.Viewer.ErrorList
             // is fixed.
             ThreadHelper.ThrowIfNotOnUIThread();
 
-            int result = VsShellUtilities.ShowMessageBox(Microsoft.VisualStudio.Shell.ServiceProvider.GlobalProvider,
+            int result = VsShellUtilities.ShowMessageBox(ServiceProvider.GlobalProvider,
                                                          dialogMessage,
                                                          null, // title
                                                          OLEMSGICON.OLEMSGICON_QUERY,
@@ -292,7 +292,7 @@ namespace Microsoft.Sarif.Viewer.ErrorList
 
             if (error != null)
             {
-                VsShellUtilities.ShowMessageBox(Microsoft.VisualStudio.Shell.ServiceProvider.GlobalProvider,
+                VsShellUtilities.ShowMessageBox(ServiceProvider.GlobalProvider,
                                                 error,
                                                 null, // title
                                                 OLEMSGICON.OLEMSGICON_CRITICAL,
@@ -306,10 +306,7 @@ namespace Microsoft.Sarif.Viewer.ErrorList
             // Clear previous data
             if (cleanErrors)
             {
-                CodeAnalysisResultManager.Instance.ClearCurrentMarkers();
-                SarifTableDataSource.Instance.CleanAllErrors();
-                CodeAnalysisResultManager.Instance.RunDataCaches.Clear();
-                CodeAnalysisResultManager.Instance.CurrentRunId = -1;
+                CleanAllErrors();
             }
 
             bool hasResults = false;
@@ -338,10 +335,10 @@ namespace Microsoft.Sarif.Viewer.ErrorList
 
             if (!hasResults && showMessageOnNoResults)
             {
-                ThreadHelper.JoinableTaskFactory.Run(async ()  =>
+               ThreadHelper.JoinableTaskFactory.RunAsync(async ()  =>
                {
                    await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
-                   VsShellUtilities.ShowMessageBox(Microsoft.VisualStudio.Shell.ServiceProvider.GlobalProvider,
+                   VsShellUtilities.ShowMessageBox(ServiceProvider.GlobalProvider,
                                                    string.Format(Resources.NoResults_DialogMessage, logFilePath),
                                                    null, // title
                                                    OLEMSGICON.OLEMSGICON_INFO,
@@ -349,6 +346,14 @@ namespace Microsoft.Sarif.Viewer.ErrorList
                                                    OLEMSGDEFBUTTON.OLEMSGDEFBUTTON_FIRST);
                });
             }
+        }
+
+        public static void CleanAllErrors()
+        {
+            CodeAnalysisResultManager.Instance.ClearCurrentMarkers();
+            SarifTableDataSource.Instance.CleanAllErrors();
+            CodeAnalysisResultManager.Instance.RunDataCaches.Clear();
+            CodeAnalysisResultManager.Instance.CurrentRunId = -1;
         }
 
         private ErrorListService()
@@ -465,5 +470,5 @@ namespace Microsoft.Sarif.Viewer.ErrorList
                 }
             }
         }
-    }
+     }
 }
--- a/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
+++ b/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.IO;
 using System.Linq;
 using System.Security;
@@ -350,8 +351,9 @@ namespace Microsoft.Sarif.Viewer.ErrorList
 
         public static void CleanAllErrors()
         {
-            CodeAnalysisResultManager.Instance.ClearCurrentMarkers();
             SarifTableDataSource.Instance.CleanAllErrors();
+
+            CodeAnalysisResultManager.Instance.ClearCurrentMarkers();
             CodeAnalysisResultManager.Instance.RunDataCaches.Clear();
             CodeAnalysisResultManager.Instance.CurrentRunId = -1;
         }
--- a/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
+++ b/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
@@ -21,6 +21,7 @@ using Microsoft.CodeAnalysis.Sarif.Visitors;
 using Microsoft.CodeAnalysis.Sarif.Writers;
 using Microsoft.Sarif.Viewer.Models;
 using Microsoft.Sarif.Viewer.Sarif;
+using Microsoft.Sarif.Viewer.Tags;
 using Microsoft.VisualStudio.PlatformUI;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
@@ -199,6 +200,22 @@ namespace Microsoft.Sarif.Viewer.ErrorList
         public static void CloseSarifLogs(IEnumerable<string> logFiles)
         {
             SarifTableDataSource.Instance.ClearErrorsForLogFiles(logFiles);
+
+            List<int> runIdsToClear = new List<int>();
+
+            foreach (string logFile in logFiles)
+            {
+                runIdsToClear.AddRange(CodeAnalysisResultManager.Instance.RunDataCaches.
+                    Where(runDataCacheKvp => runDataCacheKvp.Value.LogFilePath.Equals(logFile, StringComparison.OrdinalIgnoreCase)).
+                    Select(runDataCacheKvp => runDataCacheKvp.Key).
+                    ToList());
+            }
+
+            foreach (int runIdToClear in runIdsToClear)
+            {
+                CodeAnalysisResultManager.Instance.RunDataCaches.Remove(runIdToClear);
+                SarifTagger.RemoveAllTagsForRun(runIdToClear);
+            }
         }
 
         /// <summary>
@@ -352,7 +369,7 @@ namespace Microsoft.Sarif.Viewer.ErrorList
         public static void CleanAllErrors()
         {
             SarifTableDataSource.Instance.CleanAllErrors();
-
+            SarifTagger.RemoveAllTags();
             CodeAnalysisResultManager.Instance.ClearCurrentMarkers();
             CodeAnalysisResultManager.Instance.RunDataCaches.Clear();
             CodeAnalysisResultManager.Instance.CurrentRunId = -1;
@@ -364,7 +381,7 @@ namespace Microsoft.Sarif.Viewer.ErrorList
 
         private int WriteRunToErrorList(Run run, string logFilePath)
         {
-            RunDataCache dataCache = new RunDataCache(run);
+            RunDataCache dataCache = new RunDataCache(run, logFilePath);
             CodeAnalysisResultManager.Instance.RunDataCaches.Add(++CodeAnalysisResultManager.Instance.CurrentRunId, dataCache);
             CodeAnalysisResultManager.Instance.CacheUriBasePaths(run);
             List<SarifErrorListItem> sarifErrors = new List<SarifErrorListItem>();
--- a/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
+++ b/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
@@ -214,7 +214,7 @@ namespace Microsoft.Sarif.Viewer.ErrorList
             foreach (int runIdToClear in runIdsToClear)
             {
                 CodeAnalysisResultManager.Instance.RunDataCaches.Remove(runIdToClear);
-                SarifTagger.RemoveAllTagsForRun(runIdToClear);
+                SarifLocationTagger.RemoveAllTagsForRun(runIdToClear);
             }
         }
 
@@ -369,7 +369,7 @@ namespace Microsoft.Sarif.Viewer.ErrorList
         public static void CleanAllErrors()
         {
             SarifTableDataSource.Instance.CleanAllErrors();
-            SarifTagger.RemoveAllTags();
+            SarifLocationTagger.RemoveAllTags();
             CodeAnalysisResultManager.Instance.RunDataCaches.Clear();
             CodeAnalysisResultManager.Instance.CurrentRunId = -1;
         }
--- a/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
+++ b/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
@@ -370,7 +370,6 @@ namespace Microsoft.Sarif.Viewer.ErrorList
         {
             SarifTableDataSource.Instance.CleanAllErrors();
             SarifTagger.RemoveAllTags();
-            CodeAnalysisResultManager.Instance.ClearCurrentMarkers();
             CodeAnalysisResultManager.Instance.RunDataCaches.Clear();
             CodeAnalysisResultManager.Instance.CurrentRunId = -1;
         }
--- a/src/Sarif.Viewer.VisualStudio/NotifyPropertyChangedObject.cs
+++ b/src/Sarif.Viewer.VisualStudio/NotifyPropertyChangedObject.cs
@@ -12,10 +12,7 @@ namespace Microsoft.Sarif.Viewer
 
         protected void NotifyPropertyChanged(String info)
         {
-            if (PropertyChanged != null)
-            {
-                PropertyChanged(this, new PropertyChangedEventArgs(info));
-            }
+            PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(info));
         }
     }
 }
--- a/src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
+++ b/src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
@@ -205,7 +205,7 @@ namespace Microsoft.Sarif.Viewer
             ThreadHelper.ThrowIfNotOnUIThread();
 
             // For these cases, this event has nothing to do with this item
-            if (this._tag == null)
+            if (this._tag != null)
             {
                 return true;
             }
--- a/src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
+++ b/src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
@@ -262,15 +262,15 @@ namespace Microsoft.Sarif.Viewer
             }
 
 
+            _tag.CaretEnteredTag += CaretEnteredTag;
             _wpfTextView = wpfTextView;
-            _wpfTextView.Closed += TextViewClosed;
             _vsWindowFrame = vsWindowFrame;
-
-            _wpfTextView.Caret.PositionChanged += CaretPositionChanged;
-            _wpfTextView.LayoutChanged += ViewLayoutChanged;
 
             return true;
         }
+
+        // When the VS Editor tag has the caret moved inside of it, let's just pass along the region selection.
+        private void CaretEnteredTag(object sender, EventArgs e) => this.RaiseRegionSelected?.Invoke(this, new EventArgs());
 
         private static bool TryCreateTextSpanWithinDocumentFromSourceRegion(Region region, IVsWindowFrame vsWindowFrame, out TextSpan textSpan)
         {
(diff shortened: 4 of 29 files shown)


GabeDeBacker:
>                     NotifyPropertyChanged("RegionDisplayString");

Change this to nameof

---
Refers to: src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs:52 in cd06626. [](commit_id = cd066266bcaec09b6e172908fac01150f00322aa, deletion_comment = False)

GabeDeBacker:
>                     NotifyPropertyChanged("RegionDisplayString");

Same below.

---
In reply to: [685018245](https://github.com/microsoft/sarif-visualstudio-extension/pull/191#issuecomment-685018245) [](ancestors = 685018245)

---
Refers to: src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs:52 in cd06626. [](commit_id = cd066266bcaec09b6e172908fac01150f00322aa, deletion_comment = False)

src/Sarif.Viewer.VisualStudio/NotifyPropertyChangedObject.cs
```diff
@@ -12,10 +12,7 @@ public abstract class NotifyPropertyChangedObject : INotifyPropertyChanged
 
         protected void NotifyPropertyChanged(String info)
```
GabeDeBacker, line 13: >String inf [](start = 45, length = 10)

This can use compiler services caller name and save a lot nameof() usage

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,553 @@
...
+                SarifLocationTag[] possibleTags = null;
+                using (TagListLock.EnterReadLock())
+                {
+                    if (this.sarifTags.Count == 0)
+                    {
+                        return;
+                    }
+
+                    possibleTags = new SarifLocationTag[sarifTags.Count];
+                    sarifTags.CopyTo(possibleTags, 0);
```
GabeDeBacker, line 511: this can be possbileTags = new List(sarifTags)

src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs
```diff
@@ -238,162 +208,7 @@ internal void AttachToDocument(string documentName, long docCookie, IVsWindowFra
...
-
-                marker.AddTracking(textView, textSnapshot, docCookie, spanToColor);
-            }
-            catch (Exception e)
-            {
-                // Log the exception and move ahead. We don't want to bubble this or fail.
-                // We just don't color the problem line.
-                Debug.Print(e.Message);
-            }
+            LineMarker.TryTagDocument(documentName, frame);
```
GabeDeBacker, line 211: >LineMarker [](start = 12, length = 10)

This can be null.

src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs
```diff
@@ -160,7 +152,7 @@ public void NavigateTo(bool usePreviewPane = true)
 
             if (LineMarker != null)
             {
-                LineMarker?.NavigateTo(usePreviewPane);
+                LineMarker?.TryNavigateTo(usePreviewPane);
```
GabeDeBacker, line 155: >? [](start = 26, length = 1)

Not necessary

src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
```diff
@@ -766,6 +769,79 @@ internal static uint FindDocument(IVsRunningDocumentTable runningDocTable, strin
             return cookieDocLock;
         }
 
+        /// <summary>
+        /// Helper method for getting a IWpfTextView from a IVsTextView object
+        /// </summary>
+        /// <param name="textView"></param>
+        /// <returns></returns>
+        public static IWpfTextView GetWpfTextView(IVsTextView textView)
```
GabeDeBacker, line 777: >GetWpfTextView [](start = 35, length = 14)

Rename to "TryGet" and make IWpfTextView an out param

src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
```diff
@@ -766,6 +769,79 @@ internal static uint FindDocument(IVsRunningDocumentTable runningDocTable, strin
...
+            }
+
+            if (textViewHost == null)
+            {
+                return null;
+            }
+            return textViewHost.TextView;
+        }
+
+        public static IVsTextView GetTextViewFromFrame(IVsWindowFrame frame)
```
GabeDeBacker, line 798: >GetTextViewFromFrame [](start = 34, length = 20)

Rename to "TryGet" and make IVsWindowFramean out param

src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs
```diff
@@ -76,20 +68,20 @@ public virtual string FilePath
         {
             get
             {
-                return _filePath;
+                return this._filePath;
             }
             set
             {
-                if (value != _filePath)
+                if (value != this._filePath)
```
roywill, line 75: Is it not better to use !String::Equals as the instances can be different but the content the same?

src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs
```diff
@@ -98,20 +90,20 @@ internal virtual string UriBaseId
         {
             get
             {
-                return _uriBaseId;
+                return this._uriBaseId;
             }
             set
             {
-                if (value != _uriBaseId)
+                if (value != this._uriBaseId)
```
roywill, line 97: Same

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -57,478 +59,293 @@ public ResultTextMarker(int runId, Region region, string fullFilePath)
...
+        /// <returns>Returns true if a VS editor was opened.</returns>
+        public bool TryNavigateTo(bool usePreviewPane)
+        {
+            ThreadHelper.ThrowIfNotOnUIThread();
+
+            return this.TryNavigateTo(usePreviewPane, retryNaviation: true);
+
+        }
+
+        // This method is called when you click an in-line link, with an integer target, which
```
roywill, line 75: >integer target, [](start = 73, length = 15)

Huh what integer target?

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -57,478 +59,293 @@ public ResultTextMarker(int runId, Region region, string fullFilePath)
...
         /// simply return here (before the fix this code threw an exception which terminated VS).
         /// </summary>
         /// <param name="highlightColor">Color</param>
         public void AddHighlightMarker(string highlightColor)
         {
-            if (!IsTracking())
+            if (this._tag != null)
             {
-                return;
+                _tag.Tag = new TextMarkerTag(highlightColor ?? Color);
```
roywill, line 184: ?
missing this.

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -57,478 +59,293 @@ public ResultTextMarker(int runId, Region region, string fullFilePath)
...
 
-                // Create a span with the calculated markers
-                markerStart = startTextLine.Start.Position + startColumn - 1;
-                markerEnd = endTextLine.Start.Position + endColumn - 1;
-                spanToColor = Span.FromBounds(markerStart, markerEnd);
+            // SARIF regions are 1 based, VS is one based.)
+            textSpan.iStartLine = Math.Max(region.StartLine - 1, 0);
+            textSpan.iEndLine = Math.Max(region.EndLine - 1, 0);
+            textSpan.iStartIndex = Math.Max(region.StartColumn - 1, 0);
+            textSpan.iEndIndex = Math.Max(region.EndColumn - 1, 0);
```
roywill, line 276: Man this looks like a perfect opportunity for an extension method to set a textSpan from a region.

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -57,478 +59,293 @@ public ResultTextMarker(int runId, Region region, string fullFilePath)
...
-
-                var textLineStart = _textView.GetTextViewLineContainingBufferPosition(startPoint);
-                var textLineEnd = _textView.GetTextViewLineContainingBufferPosition(endPoint);
-
-                sourceLocation.StartColumn = startLine.Start.Position - textLineStart.Start.Position;
-                sourceLocation.EndColumn = endLine.End.Position - textLineEnd.Start.Position;
-                sourceLocation.StartLine = startLine.LineNumber + 1;
-                sourceLocation.EndLine = endLine.LineNumber + 1;
+                textSpan.iEndLine = 0;
             }
```
roywill, line 305: How did this become less than zero after above?

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -57,478 +59,293 @@ public ResultTextMarker(int runId, Region region, string fullFilePath)
...
-            if (_textView != null)
+            if (textSpan.iEndLine > lastLine)
             {
-                // If a new snapshot wasn't generated, then skip this layout
-                if (e.NewViewState.EditSnapshot != e.OldViewState.EditSnapshot)
-                {
-                    UpdateAtCaretPosition(_textView.Caret.Position);
-                }
+                textSpan.iEndLine = lastLine;
             }
```
roywill, line 325: Sniff textSpan.iEndLine = Math.Max()

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -57,478 +59,293 @@ public ResultTextMarker(int runId, Region region, string fullFilePath)
...
+            ITextSnapshotLine endTextLine = textSnapshot.GetLineFromLineNumber(textSpan.iEndLine);
 
-            if (handler != null)
+            // If we are highlighting just one line and the column values don't make
+            // sense or we corrected one or more of them, then simply mark the
+            // entire line
+            if (textSpan.iEndLine == textSpan.iStartLine && (coerced || textSpan.iStartIndex >= textSpan.iEndIndex))
             {
-                handler(this, e);
+                textSpan.iStartIndex = 0;
```
roywill, line 344: Weird if coerced, shouldn't iStartIndex already be zero?

src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
```diff
@@ -766,6 +769,79 @@ internal static uint FindDocument(IVsRunningDocumentTable runningDocTable, strin
...
 
+        /// <summary>
+        /// Helper method for getting a IWpfTextView from a IVsTextView object
+        /// </summary>
+        /// <param name="textView"></param>
+        /// <returns></returns>
+        public static IWpfTextView GetWpfTextView(IVsTextView textView)
+        {
+            IWpfTextViewHost textViewHost = null;
+            IVsUserData userData = textView as IVsUserData;
```
roywill, line 780: >        IVsUserData userData = textView as IVsUserData; [](start = 4, length = 55)

sniff looks like is

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -198,14 +200,30 @@ public static void ProcessLogFile(string filePath, string toolFormat, bool promp
...
             SarifTableDataSource.Instance.ClearErrorsForLogFiles(logFiles);
+
+            List<int> runIdsToClear = new List<int>();
+
+            foreach (string logFile in logFiles)
+            {
+                runIdsToClear.AddRange(CodeAnalysisResultManager.Instance.RunDataCaches.
+                    Where(runDataCacheKvp => runDataCacheKvp.Value.LogFilePath.Equals(logFile, StringComparison.OrdinalIgnoreCase)).
+                    Select(runDataCacheKvp => runDataCacheKvp.Key).
+                    ToList());
```
roywill, line 211: AddRange takes an IEnumerator.  Waste of time making a list.

src/Sarif.Viewer.VisualStudio/Models/StackFrameModel.cs
```diff
@@ -44,8 +44,8 @@ public int Line
                 if (value != this._line)
                 {
                     this._line = value;
-                    NotifyPropertyChanged("Line");
-                    NotifyPropertyChanged("FullyQualifiedLocation");
+                    NotifyPropertyChanged();
```
roywill, line 47: Why twice?

src/Sarif.Viewer.VisualStudio/Tags/ISarifLocationTagger.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// <summary>
+        /// Adds a tag to report to visual studio.
+        /// </summary>
+        /// <param name="sourceRegion">The original span from the region in the SARIF log.</param>
+        /// <param name="documentSpan">The span to use to create the tag relative to an open document.</param>
+        /// <param name="runId">The SARIF log run identifier associated with this tag.</param>
+        /// <param name="textMarkerTag">The text marker tag to display for this tag.</param>
+        /// <returns>Returns a new instance of <see cref="ISarifLocationTag"/></returns>
+        /// <remarks>
+        /// This <paramref name="documentSpan"/>is not necessarily the same as <paramref name="documentSpan"/>.
```
roywill, line 22: sp

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,553 @@
...
+        private static ReaderWriterLockSlimWrapper TagListLock = new ReaderWriterLockSlimWrapper(new ReaderWriterLockSlim());
+
+        /// <summary>
+        /// Provides a dictionary from source code files (C/C++, etc.) to a set of tags to display in the VS editor.
+        /// </summary>
+        /// <remarks>
+        /// This is a static instance as a "tagger" is created based on an opened text buffer but the tags" persist beyond that instance
+        /// of a tagger and will be re-used if the text buffer is re-opened. (For example in a file close and re-open scenario).
+        /// </remarks>
+        private static readonly Dictionary<string, List<SarifLocationTag>> SourceCodeFileToSarifTags = new Dictionary<string, List<SarifLocationTag>>();
```
roywill, line 33: case insensitive?

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,553 @@
...
+                }
+            }
+        }
+
+        /// <inheritdoc/>
+        public bool TryGetTag(Region sourceRegion, out ISarifLocationTag existingTag)
+        {
+            using (TagListLock.EnterReadLock())
+            {
+                existingTag = this.sarifTags.FirstOrDefault(sarifTag => sarifTag.SourceRegion.ValueEquals(sourceRegion));
```
roywill, line 239: Why not Any?

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,553 @@
...
+        public void RemoveTag(ISarifLocationTag tag)
+        {
+            if (!(tag is SarifLocationTag sarifTag))
+            {
+                return;
+            }
+
+            using (TagListLock.EnterWriteLock())
+            {
+                using (this.Update())
```
roywill, line 255: Hmm, this is inverted from above?

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,553 @@
...
+
+            using (TagListLock.EnterReadLock())
+            {
+                if (!RunIdToSarifTags.TryGetValue(runId, out List<SarifLocationTag> sarifTagsForRun) ||
+                    sarifTagsForRun.Count == 0)
+                {
+                    return;
+                }
+
+                tagsToRemove = new List<SarifLocationTag>(sarifTagsForRun);
```
roywill, line 286: You are making a copy here.  Comment please.

src/Sarif.Viewer.VisualStudio/Models/StackFrameModel.cs
GabeDeBacker: Because a change to this property also affects the value returned from the "full qualified location" property which is computed in the "getter" below.

---
In reply to: [481408972](https://github.com/microsoft/sarif-visualstudio-extension/pull/191#discussion_r481408972) [](ancestors = 481408972)

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
GabeDeBacker: Any returns a bool and I need the reference to the tag.

---
In reply to: [481416588](https://github.com/microsoft/sarif-visualstudio-extension/pull/191#discussion_r481416588) [](ancestors = 481416588)

src/Sarif.Viewer.VisualStudio/Tags/ISarifLocationTagger.cs
GabeDeBacker: What exactly? Spell checker seems happy.

---
In reply to: [481412872](https://github.com/microsoft/sarif-visualstudio-extension/pull/191#discussion_r481412872) [](ancestors = 481412872)

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
GabeDeBacker: Comment is absolutely no longer valid, and it never was.

---
In reply to: [481394177](https://github.com/microsoft/sarif-visualstudio-extension/pull/191#discussion_r481394177) [](ancestors = 481394177)

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
GabeDeBacker: If this were done in any other place, I might consider it.

---
In reply to: [481396731](https://github.com/microsoft/sarif-visualstudio-extension/pull/191#discussion_r481396731) [](ancestors = 481396731)

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
GabeDeBacker: It really can't, but it does make it clear that the entire line is being selected. (This was existing code, so leaving it)

---
In reply to: [481398586](https://github.com/microsoft/sarif-visualstudio-extension/pull/191#discussion_r481398586) [](ancestors = 481398586)

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -57,478 +59,286 @@ public ResultTextMarker(int runId, Region region, string fullFilePath)
...
 
-            if (IsTracking())
-            {
-                RemoveTracking();
+                vsTextView.EnsureSpanVisible(documentSpan);
+                vsTextView.SetSelection(documentSpan.iStartLine, documentSpan.iStartIndex, documentSpan.iEndLine, documentSpan.iEndIndex);
             }
 
-            _tagger = null;
+            return true;
```
sherryyshi, line 164: >true [](start = 19, length = 4)

seems like this should be false, if nothing was done, but I might not be understanding the full context here

GabeDeBacker: Looking...

---
In reply to: [481461486](https://github.com/microsoft/sarif-visualstudio-extension/pull/191#discussion_r481461486) [](ancestors = 481461486)

GabeDeBacker: You are correct. Fixing this.

---
In reply to: [481462918](https://github.com/microsoft/sarif-visualstudio-extension/pull/191#discussion_r481462918) [](ancestors = 481462918,481461486)

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -57,478 +59,274 @@ public ResultTextMarker(int runId, Region region, string fullFilePath)
...
 
-        private IVsTextView GetTextViewFromFrame(IVsWindowFrame frame)
-        {
-            ThreadHelper.ThrowIfNotOnUIThread();
-            // Get the document view from the window frame, then get the text view
-            object docView;
-            int hr = frame.GetProperty((int)__VSFPROPID.VSFPROPID_DocView, out docView);
-            if ((hr != 0 && hr != 1) || docView == null)
+            if (vsWindowFrame == null ||
+                string.Compare(documentName, this.FullFilePath, StringComparison.OrdinalIgnoreCase) != 0)
```
sherryyshi, line 213: >documentName, this.FullFilePath [](start = 31, length = 31)

any chance of both of these being null and therefore passing this test?

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,548 @@
...
+            if (disposed)
+            {
+                return;
+            }
+
+            disposed = true;
+
+            if (disposing)
+            {
+                using (TagListLock.EnterReadLock())
```
sherryyshi, line 354: >TagListLock [](start = 23, length = 11)

do these locks need to be disposed?

GabeDeBacker: In theory, but they are static so we'd need to do it only package unload. I'll take a look.

---
In reply to: [481470693](https://github.com/microsoft/sarif-visualstudio-extension/pull/191#discussion_r481470693) [](ancestors = 481470693)

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -17,8 +17,8 @@
 namespace Microsoft.Sarif.Viewer
 {
     /// <summary>
-    /// This class represents an instance of "highlighed" line in the editor, holds necessary Shell objects and logic 
-    /// to managed lifecycle and appearance.
+    /// This class represents an instance of "highlighted" line in the editor, holds necessary Shell objects and logic 
+    /// to managed life cycle and appearance.
     /// </summary>
     public class ResultTextMarker
```
lgolding, line 23: >ResultTextMarker [](start = 17, length = 16)

If this class represents an instance of a highlighted line, it should be called `HighlightedLine`. I'm a big fan of naming things what they are.

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -57,478 +59,270 @@ public ResultTextMarker(int runId, Region region, string fullFilePath)
...
-            Uri uri;
-            if (Uri.TryCreate(this.FullFilePath, UriKind.Absolute, out uri))
+
+            return this.TryNavigateTo(usePreviewPane, retryNaviation: true);
+
+        }
+
+        private bool TryNavigateTo(bool usePreviewPane, bool retryNaviation)
+        {
+            ThreadHelper.ThrowIfNotOnUIThread();
```
lgolding, line 77: >ThreadHelper.ThrowIfNotOnUIThread(); [](start = 12, length = 36)

No need for the additional check. The only ways in here are (a) through the public API, which checks, and (b) recursively from this method, so the check has already been done.

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+namespace Microsoft.Sarif.Viewer.Tags
+{
+    using Microsoft.CodeAnalysis.Sarif;
+    using Microsoft.VisualStudio.Shell;
+    using Microsoft.VisualStudio.Text;
+    using Microsoft.VisualStudio.Text.Editor;
+    using Microsoft.VisualStudio.Text.Tagging;
+    using Microsoft.VisualStudio.TextManager.Interop;
+    using Microsoft.VisualStudio.Utilities;
+    using System;
```
lgolding, line 13: >System [](start = 10, length = 6)

Code base convention: `System` includes first.

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+        /// Provides a dictionary from source code files (C/C++, etc.) to a set of tags to display in the VS editor.
+        /// </summary>
+        /// <remarks>
+        /// This is a static instance as a "tagger" is created based on an opened text buffer but the tags" persist beyond that instance
+        /// of a tagger and will be re-used if the text buffer is re-opened. (For example in a file close and re-open scenario).
+        /// </remarks>
+        private static readonly Dictionary<string, List<SarifLocationTag>> SourceCodeFileToSarifTags = new Dictionary<string, List<SarifLocationTag>>(StringComparer.OrdinalIgnoreCase);
+
+        /// <summary>
+        /// Provides a dictionary from SARIF log run Id to a set of tags to display in the VS editor.
```
lgolding, line 36: >og run Id  [](start = 46, length = 10)

Rename "run ID" to "log parse id" because SARIF logs can contain more than one run.

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+    using Microsoft.VisualStudio.Text.Tagging;
+    using Microsoft.VisualStudio.TextManager.Interop;
+    using Microsoft.VisualStudio.Utilities;
+    using System;
+    using System.Collections.Generic;
+    using System.ComponentModel;
+    using System.Linq;
+    using System.Threading;
+
+    internal class SarifLocationTagger : ITagger<TextMarkerTag>, ISarifLocationTagger, ITextViewCreationListener, IDisposable
```
lgolding, line 19: >SarifLocationTagger [](start = 19, length = 19)

Add a comment that indicates that this class is "per text buffer" hence the need for the static dictionaries.

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+
+        /// <summary>
+        /// Used to track nested calls to <see cref="Update"/>.
+        /// </summary>
+        private int updateCount;
+
+        private readonly ITextBuffer textBuffer;
+        private readonly IPersistentSpanFactory persistentSpanFactory;
+        private readonly string fileName;
+        private List<SarifLocationTag> sarifTags;
```
lgolding, line 72: >sarifTags [](start = 39, length = 9)

Deserves a comment and a rename.

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+        private ITrackingSpan batchUpdateSpan;
+
+        /// <summary>
+        /// Used to track nested calls to <see cref="Update"/>.
+        /// </summary>
+        private int updateCount;
+
+        private readonly ITextBuffer textBuffer;
+        private readonly IPersistentSpanFactory persistentSpanFactory;
+        private readonly string fileName;
```
lgolding, line 71: >fileName [](start = 32, length = 8)

`filePath`?

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+        private List<SarifLocationTag> sarifTags;
+
+        private bool disposed;
+
+        /// <summary>
+        /// Create an instance of the SARIF issue tagger.
```
lgolding, line 77: >SARIF issue tagger [](start = 38, length = 18)

<see cref=`SarifLocationTagger` />

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+        /// </summary>
+        /// <param name="textBuffer">The Visual Studio text buffer that the tagger will be associated with.</param>
+        /// <param name="persistentSpanFactory">The persistent span factory that will be used to create individual tags.</param>
+        public SarifLocationTagger(ITextBuffer textBuffer, IPersistentSpanFactory persistentSpanFactory)
+        {
+            ThreadHelper.ThrowIfNotOnUIThread();
+            if (!SdkUIUtilities.TryGetFileNameFromTextBuffer(textBuffer, out string textBufferFilename))
```
lgolding, line 84: >string textBufferFilename [](start = 77, length = 25)

`out this.filePath`

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+            {
+                SarifTaggers.Add(this);
+            }
+        }
+
+        /// <inheritdoc/>
+        public event EventHandler<SnapshotSpanEventArgs> TagsChanged;
+
+        /// <summary>
+        /// Removes tags for a given SARIF log file run ID.
```
lgolding, line 124: >log file run ID [](start = 43, length = 15)

search and replace.

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+            // Next, remove any remaining tags for the run ID from the static lists.
+            using (TagListLock.EnterWriteLock())
+            {
+                if (!RunIdToSarifTags.TryGetValue(runId, out List<SarifLocationTag> sarifTagsForRun) ||
+                    sarifTagsForRun.Count == 0)
+                {
+                    return;
+                }
+
+                foreach (SarifLocationTag sarifTagForRun in sarifTagsForRun)
```
lgolding, line 152: >sarifTagForRun [](start = 42, length = 14)

`sarifTag`

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+        /// <summary>
+        /// Cleans up the static locks and lists.
+        /// </summary>
+        public static void DisposeStaticObjects()
+        {
+            List<SarifLocationTag> tagsToDispose = new List<SarifLocationTag>();
+
+            using (TagListLock.EnterWriteLock())
+            {
+                // There is no need to re-add the tags from the run Id to SARIF tag dictionary as the
```
lgolding, line 192: >re-add [](start = 39, length = 6)

collect? keep track of?

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+                // objects exist in both dictionaries and only need to be disposed once.
+                tagsToDispose.AddRange(SourceCodeFileToSarifTags.Values.SelectMany(sarifTags => sarifTags));
+
+                SourceCodeFileToSarifTags.Clear();
+                RunIdToSarifTags.Clear();
+            }
+
+            tagsToDispose.ForEach(tagToDispose => tagToDispose.Dispose());
```
lgolding, line 200: >ForEach [](start = 26, length = 7)

Don't use. Not in LINQ.

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+
+            // Note that the lock wrapper implementation does not support IDispose
+            // and therefore we still must dispose the "inner lock" held by the wrapper.
+            TagListLock.InnerLock.Dispose();
+            SarifTaggersLock.InnerLock.Dispose();
+        }
+
+            /// <inheritdoc/>
```
lgolding, line 208: >     [](start = 8, length = 4)

indentation

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+        public ISarifLocationTag AddTag(Region sourceRegion, TextSpan documentSpan, int runId, TextMarkerTag tag)
+        {
+            // Start an update so that even on a call (where the caller is not batching using Update themselves)
```
lgolding, line 211: >) [](start = 111, length = 1)

remove parens

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+            // that a "tags changed" event is fired to update VS's editor.
+            using (this.Update())
+            {
+                using (TagListLock.EnterWriteLock())
+                {
+                    SarifLocationTag existingSarifTag = this.sarifTags.FirstOrDefault(
+                        (sarifTag) =>
+                            sarifTag.SourceRegion.ValueEquals(sourceRegion));
```
lgolding, line 219: >) [](start = 75, length = 1)

Check that the "run id" also matches.

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+                    newSarifTag.PropertyChanged += SarifTagPropertyChanged;
+
+                    if (RunIdToSarifTags.TryGetValue(runId, out List<SarifLocationTag> sarifTagsForRun))
+                    {
+                        sarifTagsForRun.Add(newSarifTag);
+                    }
+                    else
+                    {
+                        RunIdToSarifTags.Add(runId, new List<SarifLocationTag> { newSarifTag });
+                    }
```
lgolding, line 250: I like the pattern
```
if (!x.TryGetValue(key, out List<T> list)
{
    list = new List<T>();
    x[key] = list;
}

list.Add(thing);
```

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+
+            using (TagListLock.EnterReadLock())
+            {
+                if (!RunIdToSarifTags.TryGetValue(runId, out List<SarifLocationTag> sarifTagsForRun) ||
+                    sarifTagsForRun.Count == 0)
+                {
+                    return;
+                }
+
+                // Copy so we can update (which can make outgoing calls) outside of lock
```
lgolding, line 311: Please make a pass for non-period-terminated comment sentences.

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+            SarifLocationTag[] possibleTags = null;
+            using (TagListLock.EnterReadLock())
+            {
+                if (this.sarifTags == null)
+                {
+                    yield break;
+                }
+
+                possibleTags = new SarifLocationTag[this.sarifTags.Count];
+                this.sarifTags.CopyTo(possibleTags, 0);
```
lgolding, line 341: >); [](start = 53, length = 2)

make a new list. no need to `CopyTo`.

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+
+        /// <inheritdoc/>
+        public IEnumerable<ITagSpan<TextMarkerTag>> GetTags(NormalizedSnapshotSpanCollection spans)
+        {
+            if (spans.Count == 0)
+            {
+                yield break;
+            }
+
+            SarifLocationTag[] possibleTags = null;
```
lgolding, line 332: >possibleTags [](start = 31, length = 12)

Rename to `currentTags`

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+                possibleTags = new SarifLocationTag[this.sarifTags.Count];
+                this.sarifTags.CopyTo(possibleTags, 0);
+            }
+
+            if (possibleTags == null)
+            {
+                yield break;
+            }
+
+            foreach (var span in spans)
```
lgolding, line 349: >var [](start = 21, length = 3)

Explicit type per code base convention.

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+            {
+                foreach (var possibleTag in possibleTags.Where((possibleTag) => possibleTag.DocumentPersistentSpan.Span != null))
+                {
+                    SnapshotSpan possibleTagSnapshotSpan = possibleTag.DocumentPersistentSpan.Span.GetSpan(span.Snapshot);
```
lgolding, line 353: >GetSpan [](start = 99, length = 7)

Add comment: `GetSpan` maps original span location to new location based on any edits.

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+                this.batchUpdateSpan = snapshotSpan;
+                return;
+            }
+
+            ITextSnapshot snapshot = this.textBuffer.CurrentSnapshot;
+
+            SnapshotSpan currentBatchSpan = this.batchUpdateSpan.GetSpan(snapshot);
+            SnapshotSpan currentUpdate = snapshotSpan.GetSpan(snapshot);
+
+            SnapshotPoint newStart = currentBatchSpan.Start.Position < currentUpdate.Start.Position ? currentBatchSpan.Start : currentUpdate.Start;
```
lgolding, line 433: >Position [](start = 60, length = 8)

Math.Min, Max

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+            if (this.batchUpdateSpan == null)
+            {
+                this.batchUpdateSpan = snapshotSpan;
+                return;
+            }
+
+            ITextSnapshot snapshot = this.textBuffer.CurrentSnapshot;
+
+            SnapshotSpan currentBatchSpan = this.batchUpdateSpan.GetSpan(snapshot);
+            SnapshotSpan currentUpdate = snapshotSpan.GetSpan(snapshot);
```
lgolding, line 431: >currentUpdate [](start = 25, length = 13)

`updateSpan`

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+                    this.UpdateBatchSpan(sarifTag.DocumentPersistentSpan.Span);
+                }
+            }
+        }
+
+        private void UpdateBatchSpan(ITrackingSpan snapshotSpan)
+        {
+            // If there currently is a batch span, update it to include the biggest
+            // range of buffer affected so far.
+            if (this.batchUpdateSpan == null)
```
lgolding, line 422: >batchUpdateSpan [](start = 21, length = 15)

Take the lock that protects the updateBatchSpan.

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+            SnapshotSpan currentUpdate = snapshotSpan.GetSpan(snapshot);
+
+            SnapshotPoint newStart = currentBatchSpan.Start.Position < currentUpdate.Start.Position ? currentBatchSpan.Start : currentUpdate.Start;
+            SnapshotPoint newEnd = currentBatchSpan.End.Position > currentUpdate.End.Position ? currentBatchSpan.End : currentUpdate.End;
+
+            this.batchUpdateSpan = snapshot.CreateTrackingSpan(new SnapshotSpan(newStart, newEnd), this.batchUpdateSpan.TrackingMode);
+        }
+
+        /// <inheritdoc/>
+        public void TextViewCreated(ITextView textView)
```
lgolding, line 440: >) [](start = 54, length = 1)

Comment: This call is forwarded from the `LocationTaggerProvider`.

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+            {
+                return;
+            }
+
+            if (!textViewFileName.Equals(this.fileName, StringComparison.OrdinalIgnoreCase))
+            {
+                return;
+            }
+
+            new TextViewCaretListener(textView, this.sarifTags);
```
lgolding, line 454: >new TextViewCaretListener(textView, this.sarifTags) [](start = 12, length = 51)

We're creating an object to which nobody receives a reference. So we're calling this purely for its side effects. Why not just call a method that produces those side effects?

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+        /// When this object is disposed by the caller and all batch updates have completed, a 
+        /// tags changed event is sent to Visual Studio.
+        /// </remarks>
+        private class BatchUpdate : IDisposable
+        {
+            private readonly SarifLocationTagger tagger;
+            public BatchUpdate(SarifLocationTagger tagger)
+            {
+                this.tagger = tagger;
+                Interlocked.Increment(ref tagger.updateCount);
```
lgolding, line 470: >updateCount [](start = 49, length = 11)

rename to `batchUpdateNestingLevel`

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+            }
+
+            new TextViewCaretListener(textView, this.sarifTags);
+        }
+
+        /// <summary>
+        /// Used to control batch changes to tags.
+        /// </summary>
+        /// <remarks>
+        /// When this object is disposed by the caller and all batch updates have completed, a
```
lgolding, line 461: >all batch updates have completed [](start = 59, length = 32)

"... and we are exiting from the outermost update nesting level"

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+                }
+
+                this.previousTagsCaretWasIn = tagsCaretIsCurrentlyIn;
+            }
+
+            private void TextView_Closed(object sender, EventArgs e)
+            {
+                this.textView.Closed -= this.TextView_Closed;
+                this.textView.LayoutChanged -= this.TextView_LayoutChanged;
+                this.textView.Caret.PositionChanged -= this.TextView_Closed;
```
lgolding, line 569: >TextView_Closed [](start = 60, length = 15)

Bug: `PositionChanged`

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
lgolding: Worth a comment to explain that the `textView` will hold a reference to the newly created object.

---
In reply to: [482264377](https://github.com/microsoft/sarif-visualstudio-extension/pull/191#discussion_r482264377) [](ancestors = 482264377)

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+            {
+                List<SarifLocationTag> possibleTags = null;
+                using (TagListLock.EnterReadLock())
+                {
+                    if (this.sarifTags.Count == 0)
+                    {
+                        return;
+                    }
+
+                    possibleTags = new List<SarifLocationTag>(sarifTags);
```
lgolding, line 536: >possibleTags [](start = 20, length = 12)

Rename to `currentTags`

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+                    UpdateAtCaretPosition(textView.Caret.Position);
+                }
+            }
+
+            private void Caret_PositionChanged(object sender, CaretPositionChangedEventArgs e)
+            {
+                UpdateAtCaretPosition(e.NewPosition);
+            }
+
+            private void UpdateAtCaretPosition(CaretPosition caretPoisition)
```
lgolding, line 526: >caretPoisition [](start = 61, length = 14)

Spelling: `caretPosition`

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+
+                // Keep track of the tags the caret is in now, versus the tags
+                // that the caret was previously in. (Yes, there can be multiple tags per text range).
+                // This is done so we don't keep re-issuing caret entered notifications while
+                // the user is moving the caret around the editor.
+                var tagsCaretIsCurrentlyIn = new List<SarifLocationTag>();
+
+                foreach (var possibleTag in possibleTags)
+                {
+                    SnapshotPoint caretBufferPosition = caretPoisition.BufferPosition;
```
lgolding, line 547: >SnapshotPoint [](start = 20, length = 13)

Do this outside the loop.

lgolding, line 547: >caretBufferPosition [](start = 34, length = 19)

`caretSnapshotPoint`

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -0,0 +1,573 @@
...
+                    if (possibleTag.DocumentPersistentSpan.Span.GetSpan(caretBufferPosition.Snapshot).Contains(caretBufferPosition))
```
lgolding, line 548: >. [](start = 58, length = 1)

null check.

src/Sarif.Viewer.VisualStudio/Tags/ISarifLocationTagger.cs
lgolding, line 22: >documentSpan [](start = 95, length = 12)

`sourceRegion`

src/Sarif.Viewer.VisualStudio/Tags/ISarifLocationTagger.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// Removes tags based on a run ID.
+        /// </summary>
+        /// <param name="runId">The SARIF log run id.</param>
+        void RemoveTagsForRun(int runId);
+
+        /// <summary>
+        /// Called to perform a batch update of tags.
+        /// </summary>
+        /// <returns>
+        /// Returns an IDisposable that when disposed sends an tags changed event to visual studio.
```
lgolding, line 52: >v [](start = 85, length = 1)

V S

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -17,8 +17,8 @@
 namespace Microsoft.Sarif.Viewer
 {
     /// <summary>
-    /// This class represents an instance of "highlighed" line in the editor, holds necessary Shell objects and logic 
-    /// to managed lifecycle and appearance.
+    /// This class represents an instance of "highlighted" line in the editor, holds necessary Shell objects and logic
```
lgolding, line 20: >  [](start = 44, length = 1)

of a

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
GabeDeBacker: So if I remove this line the analyzers still complain. Do we need to update the analyzers? (They are in error in this case).

---
In reply to: [482127215](https://github.com/microsoft/sarif-visualstudio-extension/pull/191#discussion_r482127215) [](ancestors = 482127215)

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
GabeDeBacker: Cannot use Math here because < operator is overloaded and Math doesn't deal with that.

---
In reply to: [482259488](https://github.com/microsoft/sarif-visualstudio-extension/pull/191#discussion_r482259488) [](ancestors = 482259488)

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
GabeDeBacker: Yep, that works, but... it does require more list manipulation than the existing code since the a new list is initialized with the new object and hence no need to call "add" for a new list. So technically the existing code is more efficient (minorly).

---
In reply to: [482236299](https://github.com/microsoft/sarif-visualstudio-extension/pull/191#discussion_r482236299) [](ancestors = 482236299)

src/Sarif.Viewer.VisualStudio/RunDataCache.cs
```diff
@@ -25,6 +25,10 @@ public class RunDataCache
 
         public FileRegionsCache FileRegionsCache { get; }
 
+        public string LogFilePath { get; }
+
+        public int RunId { get; }
```
lgolding, line 30: >RunId [](start = 19, length = 5)

RunIndex?

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
lgolding: Hmm. Can you remove it from the public API? If not, then yes, the analyzer could be improved.

---
In reply to: [482307432](https://github.com/microsoft/sarif-visualstudio-extension/pull/191#discussion_r482307432) [](ancestors = 482307432,482127215)

GabeDeBacker: Nope. Trying to remove it from either place makes the analyzers complain. (Ugh)

---
In reply to: [482377108](https://github.com/microsoft/sarif-visualstudio-extension/pull/191#discussion_r482377108) [](ancestors = 482377108,482307432,482127215)

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
lgolding: But it still says "id" in the comment :)

---
In reply to: [482169104](https://github.com/microsoft/sarif-visualstudio-extension/pull/191#discussion_r482169104) [](ancestors = 482169104)

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
lgolding: Note to self: Does it represent a line, or a single highlight on a line (there might be more than one).

---
In reply to: [482124036](https://github.com/microsoft/sarif-visualstudio-extension/pull/191#discussion_r482124036) [](ancestors = 482124036)

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -28,17 +28,19 @@ public class ResultTextMarker
...
+        private IWpfTextView _wpfTextView;
+        private IVsWindowFrame _vsWindowFrame;
 
         public string FullFilePath { get; set; }
         public string UriBaseId { get; set; }
         public Region Region { get; set; }
         public string Color { get; set; }
 
+        /// <summary>
+        /// Fired when an the text editor caret enters a tagged region.
```
lgolding, line 42: >an the [](start = 23, length = 6)

"an the" => "the"

lgolding:
>         public ResultTextMarker(int runId, Region region, string fullFilePath)

`runIndex`, `_runIndex`

---
Refers to: src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs:49 in c2568f8. [](commit_id = c2568f8096cda6ee6379d405a5409b94df71b27e, deletion_comment = False)

GabeDeBacker:
>         public ResultTextMarker(int runId, Region region, string fullFilePath)

This is going to linger around forever :) I'll clean this one up.

---
In reply to: [685982558](https://github.com/microsoft/sarif-visualstudio-extension/pull/191#issuecomment-685982558) [](ancestors = 685982558)

---
Refers to: src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs:49 in c2568f8. [](commit_id = c2568f8096cda6ee6379d405a5409b94df71b27e, deletion_comment = False)

lgolding:
>             Region = region;

While we're in here, the new "throw expression" syntax is nice and saves 3 lines:
```
Region = region ?? throw new ArgumentNullException(nameof(region));
```

---
Refers to: src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs:57 in c2568f8. [](commit_id = c2568f8096cda6ee6379d405a5409b94df71b27e, deletion_comment = False)

GabeDeBacker:
>             Region = region;

Yep. Agreed. I'll leave it up to you to spelunk through this code and figure out why the file name isn't also thrown. (ugh)  It isn't pretty (and not my fault) :)

---
In reply to: [685984081](https://github.com/microsoft/sarif-visualstudio-extension/pull/191#issuecomment-685984081) [](ancestors = 685984081)

---
Refers to: src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs:57 in c2568f8. [](commit_id = c2568f8096cda6ee6379d405a5409b94df71b27e, deletion_comment = False)

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
GabeDeBacker: How about "LocationTextMarker" ?

---
In reply to: [482421238](https://github.com/microsoft/sarif-visualstudio-extension/pull/191#discussion_r482421238) [](ancestors = 482421238,482124036)

lgolding: Let's talk more about this. I'm beginning to think the name might be ok and the comment might be wrong. A SARIF region can cover a range of colums in a line; it doesn't have to cover a whole line. So if this class represents a highlight for a single region, rather than always representing a whole line, then the comment should be "... an instance of a 'highlighted' region in the  editor", and the class name might be `RegionTextMarker`.

---
In reply to: [482445581](https://github.com/microsoft/sarif-visualstudio-extension/pull/191#discussion_r482445581) [](ancestors = 482445581,482421238,482124036)

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -27,508 +27,303 @@ public class ResultTextMarker
...
+                        return false;
+                    }
                 }
 
-                var sourceLocation = this.GetSourceLocation();
+                if (File.Exists(this.FullFilePath) && Uri.TryCreate(this.FullFilePath, UriKind.Absolute, out Uri uri))
+                {
+                    // Fill out the region's properties
+                    FileRegionsCache regionsCache = CodeAnalysisResultManager.Instance.RunIndexToRunDataCache[_runIndex].FileRegionsCache;
+                    Region = regionsCache.PopulateTextRegionProperties(Region, uri, true);
```
lgolding, line 133: >PopulateTextRegionProperties [](start = 42, length = 28)

Confirmed. After this call, all of `region.startLine`, `.endLine`, `.startColumn`, `.endColumn`, `.charOffset`, and `.charLength` are guaranteed to be populated.

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -27,508 +27,303 @@ public class ResultTextMarker
...
-        /// </returns>
-        public Region GetSourceLocation()
-        {
-            Region sourceLocation = Region.DeepClone();                
-            SaveCurrentTrackingData(sourceLocation);
-            return sourceLocation;
-        }
+                // Alright, navigating again didn't work, so now let's just select
+                // the region.
+                if (!TryCreateTextSpanWithinDocumentFromSourceRegion(this.Region, windowFrame, out TextSpan documentSpan))
```
lgolding, line 151: >if [](start = 16, length = 2)

`OpenDocument` returns `null` if `this.FullFilePath` refers to a non-existent file, so if we get this far, `PopulateTextRegionProperties` has been called (except for the pathological case where the path exists but you can't make a URI out of it; is that even possible?), and the region's text properties are guaranteed to be fully populated.

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -27,508 +27,303 @@ public class ResultTextMarker
...
+        {
+            ThreadHelper.ThrowIfNotOnUIThread();
 
-                var textLineStart = _textView.GetTextViewLineContainingBufferPosition(startPoint);
-                var textLineEnd = _textView.GetTextViewLineContainingBufferPosition(endPoint);
+            // SARIF regions are 1 based, VS is zero based.
+            textSpan.iStartLine = Math.Max(region.StartLine - 1, 0);
+            textSpan.iEndLine = Math.Max(region.EndLine - 1, 0);
+            textSpan.iStartIndex = Math.Max(region.StartColumn - 1, 0);
+            textSpan.iEndIndex = Math.Max(region.EndColumn - 1, 0);
```
lgolding, line 284: >; [](start = 66, length = 1)

Since the region line/column properties are guaranteed to be fully populated, none of `startLine`, `endLine`, `startColumn`, and `endColumn` can possibly be 0, so these checks aren't necessary. Just

```
textSpan.iStartLine = region.StartLine - 1;
```

_etc._ should suffice.

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -27,508 +27,303 @@ public class ResultTextMarker
...
+            ITextSnapshot textSnapshot = wpfTextView.TextSnapshot;
+            ITextSnapshotLine startTextLine = textSnapshot.GetLineFromLineNumber(textSpan.iStartLine);
+            ITextSnapshotLine endTextLine = textSnapshot.GetLineFromLineNumber(textSpan.iEndLine);
 
-        private void UpdateAtCaretPosition(CaretPosition caretPoisition)
-        {
-            // Check if the current caret position is within our region. If it is, raise the RegionSelected event.
-            if (_trackingSpan.GetSpan(_trackingSpan.TextBuffer.CurrentSnapshot).Contains(caretPoisition.Point.GetPoint(_trackingSpan.TextBuffer, PositionAffinity.Predecessor).Value))
+            // If the start column of the start lines is beyond the length of the start line
+            // then we will reset the start column to zero and maybe tag the entire line if the start and end lines are the same.
```
lgolding, line 311: >same [](start = 124, length = 4)

This _should_ violate the SARIF spec -- it should not be legal for `startColumn` to point past the end of the line -- but unfortunately the spec doesn't come right out and say that. I filed [oasis-tcs/sarif-spec#475](https://github.com/oasis-tcs/sarif-spec/issues/475), "Spec doesn't say that region.startColumn can't point past the end of the line".

Having said all that, we should consider treating this as if the spec got it right, and just returning false here.

lgolding (approved):
:shipit:

GabeDeBacker:
This also helps address issue #182

Issue #193, opened 2020-09-07
Occasional exception: can't get file name from text buffer.


lgolding:
I'm running v2.1.14 of the extension. Every now and then VS reports an exception (yellow bar). The log says:

```xml
  <entry>
    <record>919</record>
    <time>2020/09/07 15:15:14.024</time>
    <type>Error</type>
    <source>Editor or Editor Extension</source>
    <description>System.ArgumentException: Always expect to be able to get file name from text buffer.&#x000D;&#x000A;
Parameter name: textBuffer&#x000D;&#x000A;
   at Microsoft.Sarif.Viewer.Tags.SarifLocationTagger..ctor(ITextBuffer textBuffer, IPersistentSpanFactory persistentSpanFactory) in D:\a\1\s\src\Sarif.Viewer.VisualStudio\Tags\SarifLocationTagger.cs:line 116&#x000D;&#x000A;
   at Microsoft.Sarif.Viewer.Tags.SarifLocationTaggerProvider.&lt;&gt;c__DisplayClass4_0.&lt;CreateSarifLocationTaggerInternal&gt;b__0() in D:\a\1\s\src\Sarif.Viewer.VisualStudio\Tags\SarifLocationTaggerProvider.cs:line 81&#x000D;&#x000A;
   at Microsoft.VisualStudio.Utilities.PropertyCollection.GetOrCreateSingletonProperty[T](Object key, Func`1 creator)&#x000D;&#x000A;
   at Microsoft.Sarif.Viewer.Tags.SarifLocationTaggerProvider.CreateSarifLocationTaggerInternal(ITextBuffer textBuffer) in D:\a\1\s\src\Sarif.Viewer.VisualStudio\Tags\SarifLocationTaggerProvider.cs:line 79&#x000D;&#x000A;
   at Microsoft.Sarif.Viewer.Tags.SarifLocationTaggerProvider.CreateTagger[T](ITextView textView, ITextBuffer buffer) in D:\a\1\s\src\Sarif.Viewer.VisualStudio\Tags\SarifLocationTaggerProvider.cs:line 51&#x000D;&#x000A;
   at Microsoft.VisualStudio.Text.Tagging.Implementation.TagAggregator`1.InstantiateViewTagger(ITextView textView, Lazy`2 factory, ITextBuffer textBuffer, IList`1 newTaggers)&#x000D;&#x000A;--- End of stack trace from previous location where exception was thrown ---&#x000D;&#x000A;
[...]

PR #194, opened 2020-09-10
SARIF location tagger does not need to throw on no file-name

GabeDeBacker:
This addresses issue #193 

VS allows a text buffer to be created without file names at any time.

Our tagger requires a file name to be able to associate a SARIF result location with that file name.

So, if we don't have a file name, then don't return a tagger. Visual Studio's tagger aggregator expects and correct handles a return value of null.

This also removes some unnecessary code that for retrieving a tagger from an existing text buffer (VS already has APIs for that on the text buffer).


GabeDeBacker:
Closing this due to finding other issues (unrelated to this) that I need to address

PR #195, opened 2020-09-14
SARIF location tagger does not need to throw on no file-name

GabeDeBacker:
This addresses issue #193 

VS allows a text buffer to be created without file names at any time.

Our tagger requires a file name to be able to associate a SARIF result location with that file name.

So, if we don't have a file name, then don't return a tagger. Visual Studio's tagger aggregator expects and correct handles a return value of null.

This also removes some unnecessary code for retrieving a tagger from an existing text buffer (VS already has APIs for that on the text buffer).

Beyond that, the "result text marker" was caching certain Visual Studio interfaces (such as IWpfTextView, etc.) that cannot be cached and re-used as when a view is closed (users closes a document) they are not longer valid. So the "result text markers" no longer cache any Visual Studio interfaces.

Furthermore, the code flow steps were not being properly highlighted because the code was re-mapping the original region from the SARIF log over and over again via a line like "this.Region = Remap(this.Region)" (oops) so now the remapping is kept separate from the original region.


src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
```diff
@@ -672,36 +678,80 @@ private static string GetCommonSuffix(string firstPath, string secondPath)
...
+            if (string.IsNullOrEmpty(documentName))
+            {
+                return;
+            }
+
+            if (RunIndexToRunDataCache == null)
+            {
+                return;
+            }
+
```
GabeDeBacker, line 718: Move this check up. If we don't have the cache, then there is no sense doing all the rest of this.

GabeDeBacker:
Update version files and readme history

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -27,17 +25,27 @@ public class ResultTextMarker
...
         public string Color { get; set; }
 
+        public Region Region
+        {
+            get => this.region;
+            set
+            {
+                this.region = value;
+                this.remapResult = null;
+                this.mappedRegion = null;
```
GabeDeBacker, line 45: Comment this.

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
```diff
@@ -672,36 +678,80 @@ private static string GetCommonSuffix(string firstPath, string secondPath)
...
+            if (vsTextView.GetBuffer(out IVsTextLines vsTextLines) != VSConstants.S_OK)
+            {
+                return;
+            }
+
+            IComponentModel componentModel = (IComponentModel)Package.GetGlobalService(typeof(SComponentModel));
+            if (componentModel == null)
+            {
+                return;
+            }
```
sherryyshi, line 702: should this throw here?

sherryyshi (approved):
:shipit:

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
```diff
@@ -672,36 +678,80 @@ private static string GetCommonSuffix(string firstPath, string secondPath)
         private void AttachToDocumentChanges(uint docCookie, IVsWindowFrame pFrame)
         {
             ThreadHelper.ThrowIfNotOnUIThread();
+
+            if (RunIndexToRunDataCache == null)
```
roywill, line 682: >== null [](start = 39, length = 7)

is null would be better here.

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
```diff
@@ -672,36 +678,80 @@ private static string GetCommonSuffix(string firstPath, string secondPath)
...
+                return;
+            }
+
+            IVsEditorAdaptersFactoryService editorAdapterFactoryService = componentModel.GetService<IVsEditorAdaptersFactoryService>();
+            if (editorAdapterFactoryService == null)
+            {
+                return;
+            }
+
+            ITextBuffer textBuffer = editorAdapterFactoryService.GetDataBuffer(vsTextLines);
```
roywill, line 710: This should move down to after the if below

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
```diff
@@ -672,36 +678,80 @@ private static string GetCommonSuffix(string firstPath, string secondPath)
...
+
             string documentName = GetDocumentName(docCookie);
 
-            if (!string.IsNullOrEmpty(documentName))
+            if (string.IsNullOrEmpty(documentName))
+            {
+                return;
+            }
+
+            foreach (int key in RunIndexToRunDataCache.Keys)
```
roywill, line 719: >   foreach (int key in RunIndexToRunDataCache.Keys) [](start = 9, length = 51)

We should not get the keys and then reindex.  You should get the keyvalue pair.

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
```diff
@@ -672,36 +678,80 @@ private static string GetCommonSuffix(string firstPath, string secondPath)
...
+                return;
+            }
+
+            string documentName = GetDocumentName(docCookie);
+            if (string.IsNullOrEmpty(documentName))
+            {
+                return;
+            }
+
+            foreach (int key in RunIndexToRunDataCache.Keys)
```
roywill, line 748: Nuts

michaelcfanning (approved):
:shipit:

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
GabeDeBacker: Not in this case, failure is okay.

---
In reply to: [488137585](https://github.com/microsoft/sarif-visualstudio-extension/pull/195#discussion_r488137585) [](ancestors = 488137585)

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
GabeDeBacker: There are "== null" checks everywhere, leaving for consistency

---
In reply to: [489583167](https://github.com/microsoft/sarif-visualstudio-extension/pull/195#discussion_r489583167) [](ancestors = 489583167)

src/build.props
```diff
@@ -19,7 +19,7 @@
     <Company Condition=" '$(Company)' == '' ">Microsoft</Company>
     <Product Condition=" '$(Product)' == '' ">Microsoft SARIF Viewer for Visual Studio</Product>
     <Copyright Condition=" '$(Copyright)' == '' ">© Microsoft Corporation. All rights reserved.</Copyright>
-    <VersionPrefix Condition=" '$(VersionPrefix)' == ''">2.1.14</VersionPrefix>
+    <VersionPrefix Condition=" '$(VersionPrefix)' == ''">2.1.15</VersionPrefix>
```
lgolding, line 22: >15 [](start = 61, length = 2)

Revert: 2.1.14 not yet published.

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -190,90 +193,114 @@ public void AddHighlightMarker(string highlightColor)
...
-                _tag = existingTag;
+                this.tag = existingTag;
             }
 
 
-            _tag.CaretEnteredTag += CaretEnteredTag;
-            _wpfTextView = wpfTextView;
-            _vsWindowFrame = vsWindowFrame;
-            _wpfTextView.Closed += this.TextViewClosed;
+            if (!SdkUIUtilities.TryGetActiveViewForTextBuffer(this.tag.DocumentPersistentSpan.Span.TextBuffer, out IWpfTextView wpfTextView))
```
roywill, line 243: Assume that this cannot be null?

src/build.props
lgolding: Nope: some insiders have already installed it.

---
In reply to: [489624314](https://github.com/microsoft/sarif-visualstudio-extension/pull/195#discussion_r489624314) [](ancestors = 489624314)

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
```diff
@@ -672,36 +678,80 @@ private static string GetCommonSuffix(string firstPath, string secondPath)
...
+
+            ITextBuffer textBuffer = editorAdapterFactoryService.GetDataBuffer(vsTextLines);
+
             string documentName = GetDocumentName(docCookie);
 
-            if (!string.IsNullOrEmpty(documentName))
+            if (string.IsNullOrEmpty(documentName))
+            {
+                return;
+            }
```
GabeDeBacker, line 715: Move this up, this is an early out

roywill (approved):
:shipit:

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
```diff
@@ -672,36 +678,85 @@ private static string GetCommonSuffix(string firstPath, string secondPath)
...
+                    SarifErrors.
+                    Where(sarifError => string.Compare(documentName, sarifError.FileName, StringComparison.OrdinalIgnoreCase) == 0);
+
+                foreach (SarifErrorListItem sarifError in sarifErrorsForDocument)
                 {
-                    foreach (SarifErrorListItem sarifError in RunIndexToRunDataCache[key].SarifErrors)
-                    {
-                        sarifError.DetachFromDocument();
-                    }
+                    sarifError.RemoveMarkers();
```
GabeDeBacker, line 759: >RemoveMarkers [](start = 31, length = 13)

Perhaps rename to "ResetTagColor{

GabeDeBacker:
>             AttachToDocumentChanges(docCookie, pFrame);

Rename to Attach and Detach to something like (TagDocument and ResetTagColors or something appropriate).

---
Refers to: src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs:597 in 7aaf01c. [](commit_id = 7aaf01c2fda26a2ce96c51124594ad9de9c96fca, deletion_comment = False)

GabeDeBacker:
>     public class ResultTextMarker

Add a comment here that explains that this object exists as long as a text buffer that is "tagged" exists. It will outlast any view\frame

---
Refers to: src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs:21 in 7aaf01c. [](commit_id = 7aaf01c2fda26a2ce96c51124594ad9de9c96fca, deletion_comment = False)

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -27,17 +25,47 @@ public class ResultTextMarker
...
+        /// <summary>
+        /// This is the original region from the SARIF log file before
+        /// it is remapped to an open document by the <see cref="TryMapRegion" method./>
+        /// </summary>
+        private Region region;
+
+        /// <summary>
+        /// Contains the region information mapped to a file on disk.
+        /// </summary>
+        private Region mappedRegion;
```
GabeDeBacker, line 37: >mappedRegion [](start = 23, length = 12)

fullyPopulatedRegion

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -27,17 +25,47 @@ public class ResultTextMarker
...
+
+        /// <summary>
+        /// Indicates whether a call to <see cref="TryMapRegion"/> has already occurred and what the result
+        /// of the remap was.
+        /// </summary>
+        private bool? remapResult;
```
GabeDeBacker, line 43: >remapResult [](start = 22, length = 11)

regionIsFullyPopulated

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -61,15 +89,37 @@ public ResultTextMarker(int runIndex, Region region, string fullFilePath)
...
-#pragma warning disable VSTHRD010 // This is an analyzer false positive. 
-            return this.TryNavigateTo(usePreviewPane, retryNaviation: true);
-#pragma warning restore VSTHRD010
-        }
-
-        private bool TryNavigateTo(bool usePreviewPane, bool retryNaviation)
-        {
             ThreadHelper.ThrowIfNotOnUIThread();
 
+            // If the tag doesn't have a persistent span, or it isn't open
```
GabeDeBacker, line 94: >it [](start = 61, length = 2)

it's associated document isn't open

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -61,15 +89,37 @@ public ResultTextMarker(int runIndex, Region region, string fullFilePath)
...
+                this.tag.DocumentPersistentSpan.Span.TextBuffer == null)
+            {
+                if (!File.Exists(this.FullFilePath))
+                {
+                    if (!CodeAnalysisResultManager.Instance.TryRebaselineAllSarifErrors(runIndex, this.UriBaseId, this.FullFilePath))
+                    {
+                        return false;
+                    }
+                }
+                else if (!this.TryMapRegion())
```
GabeDeBacker, line 107: > else [](start = 15, length = 5)

This should not be an else-if..

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -61,15 +89,37 @@ public ResultTextMarker(int runIndex, Region region, string fullFilePath)
...
 
+            // If the tag doesn't have a persistent span, or it isn't open
+            // then this indicates that we need to attempt to open the document.
+            if (this.tag?.DocumentPersistentSpan.Span == null ||
+                !this.tag.DocumentPersistentSpan.IsDocumentOpen ||
+                this.tag.DocumentPersistentSpan.Span.TextBuffer == null)
+            {
+                if (!File.Exists(this.FullFilePath))
+                {
+                    if (!CodeAnalysisResultManager.Instance.TryRebaselineAllSarifErrors(runIndex, this.UriBaseId, this.FullFilePath))
```
GabeDeBacker, line 102: >TryRebaselineAllSarifErrors [](start = 60, length = 27)

Comment the fact that if this function succeeds, this.FullFilePath is reset

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -61,15 +89,37 @@ public ResultTextMarker(int runIndex, Region region, string fullFilePath)
...
+                    return false;
+                }
+
+                IVsWindowFrame vsWindowFrame = SdkUIUtilities.OpenDocument(ServiceProvider.GlobalProvider, this.FullFilePath, usePreviewPane);
+                if (vsWindowFrame == null)
+                {
+                    return false;
+                }
+
+                // This step is super important is it will cause the document to get "tagged" before the next check
```
GabeDeBacker, line 118: >step [](start = 24, length = 4)

window frame must be shown now (at this point) because we need tagging to occur before the rest of this method executes.

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -61,15 +89,37 @@ public ResultTextMarker(int runIndex, Region region, string fullFilePath)
...
-
-        private bool TryNavigateTo(bool usePreviewPane, bool retryNaviation)
-        {
             ThreadHelper.ThrowIfNotOnUIThread();
 
+            // If the tag doesn't have a persistent span, or it isn't open
+            // then this indicates that we need to attempt to open the document.
+            if (this.tag?.DocumentPersistentSpan.Span == null ||
+                !this.tag.DocumentPersistentSpan.IsDocumentOpen ||
+                this.tag.DocumentPersistentSpan.Span.TextBuffer == null)
```
GabeDeBacker, line 98: Add a helper since this check is performed (although inverted) below so the check is consistent

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -82,85 +132,38 @@ private bool TryNavigateTo(bool usePreviewPane, bool retryNaviation)
...
+                this.tag.DocumentPersistentSpan.Span.TextBuffer != null)
             {
-                ITextSnapshot currentSnapshot = this._wpfTextView.TextSnapshot;
+
+                if (!SdkUIUtilities.TryGetActiveViewForTextBuffer(this.tag.DocumentPersistentSpan.Span.TextBuffer, out IWpfTextView wpfTextView))
+                {
+                    return false;
+                }
+
+                ITextSnapshot currentSnapshot = tag.DocumentPersistentSpan.Span.TextBuffer.CurrentSnapshot;
```
GabeDeBacker, line 145: >tag [](start = 48, length = 3)

make these "this.tag" for consistency

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
GabeDeBacker, line 243: >(this.tag.DocumentPersistentSpan.Span.TextBuffer [](start = 61, length = 48)

This can just be the passes in text buffer, do not need to use the one from the persistent span

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -190,90 +193,114 @@ public void AddHighlightMarker(string highlightColor)
...
-            _wpfTextView = wpfTextView;
-            _vsWindowFrame = vsWindowFrame;
-            _wpfTextView.Closed += this.TextViewClosed;
+            if (!SdkUIUtilities.TryGetActiveViewForTextBuffer(this.tag.DocumentPersistentSpan.Span.TextBuffer, out IWpfTextView wpfTextView))
+            {
+                return false;
+            }
+
+            this.tag.CaretEnteredTag += CaretEnteredTag;
+            wpfTextView.Closed += this.TextViewClosed;
```
GabeDeBacker, line 235: Comment what and why this is here.

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
```diff
@@ -190,90 +193,114 @@ public void AddHighlightMarker(string highlightColor)
...
+                this.remapResult = false; 
+                return false;
+            }
+
+            if (File.Exists(this.FullFilePath) &&
+                Uri.TryCreate(this.FullFilePath, UriKind.Absolute, out Uri uri))
+            {
+                // Fill out the region's properties
+                FileRegionsCache regionsCache = CodeAnalysisResultManager.Instance.RunIndexToRunDataCache[runIndex].FileRegionsCache;
+                this.mappedRegion = regionsCache.PopulateTextRegionProperties(this.region, uri, true);
```
GabeDeBacker: >PopulateTextRegionProperties [](start = 49, length = 28)

Call this function twice and open a bug in the SDK if it is modified...

GabeDeBacker: >true [](start = 96, length = 4)

Used named arguments here because what the heck is "true"????

GabeDeBacker: And why do we need to populate code snippets???

---
In reply to: [489786777](https://github.com/microsoft/sarif-visualstudio-extension/pull/195#discussion_r489786777) [](ancestors = 489786777)

lgolding (approved):
:shipit:

src/Sarif.Viewer.VisualStudio/ResultTextMarker.cs
GabeDeBacker: Verified that calling this twice does not result in any bad behavior. I was confused because the incoming region is an insertion point and so the result is an insertion point.

Being an assertion point is corrected below to extend to the entire line.

---
In reply to: [489786511](https://github.com/microsoft/sarif-visualstudio-extension/pull/195#discussion_r489786511) [](ancestors = 489786511)

lgolding (approved):
:shipit:

Issue #205, opened 2020-09-29
Crash occurring when closing and re-opening the same source file multiple times


GabeDeBacker:
When opening and closing the same source file that has SARIF results rendered in it multiple times, a crash occurs in visual studio.

This is related to issue this Windows bug:

https://microsoft.visualstudio.com/DefaultCollection/OS/_workitems/edit/29582350/

mpande98:
@lgolding

mpande98:
In process of validating PR, should be finished today

PR #206, opened 2020-09-29
Stop storing tagger as property

GabeDeBacker:
This addresses issue #205.

A "tagger" in Visual Studio is related to both a text view and a text buffer. Multiple text views can be associated with the same text buffer.

When a view is closed in Visual Studio, it disposes the tagger. (This is key to this fix).

Our tagger implementation maintains static lists of "tags" that are shared between all tagger instances. So every time a new view is created a new tagger instance is created and it attaches itself to the set of tags that are appropriate for it.

All that was fine.

However, what we were doing is storing a singleton tagger instance associated with a property bag given to each Visual Studio text buffer. This was the key to the bug. When a view is closed, Visual Studio disposes the tagger BUT the disposed instance still remained as a value in the text buffer's property bag. (oops).

So this change simply stops associating tagger instances with the text buffer.

There are also minor changes to how the events from the error list table are handled. The previous code was finding the list-box control and asking it for the selected items rather than just using the data that was already present in the event. Our extension shouldn't be "assuming" the XAML implementation of the error table in the first place.

Also, due to the change in the implementation, we can remove one of the "text view creation listeners" thereby increasing VS performance a tiny bit.

### DIFF
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -1,6 +1,6 @@
 # SARIF Viewer Visual Studio extension Release History
 ## **v2.1.17** [SARIF Viewer](https://marketplace.visualstudio.com/items?itemName=WDGIS.MicrosoftSarifViewer)
-* BUGFIX: Fix crash when closing and re-opening the a file with tagged SARIF results. [#205](https://github.com/microsoft/sarif-visualstudio-extension/issues/205)
+* BUGFIX: Fix crash when closing and re-opening a file with tagged SARIF results. [#205](https://github.com/microsoft/sarif-visualstudio-extension/issues/205)
 
 ## **v2.1.16** [SARIF Viewer](https://marketplace.visualstudio.com/items?itemName=WDGIS.MicrosoftSarifViewer)
 * BUGFIX: Offload file reading and parsing from Visual Studio's UI thread. [#160](https://github.com/microsoft/sarif-visualstudio-extension/issues/160)
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -1,4 +1,7 @@
 # SARIF Viewer Visual Studio extension Release History
+## **v2.1.17** [SARIF Viewer](https://marketplace.visualstudio.com/items?itemName=WDGIS.MicrosoftSarifViewer)
+* BUGFIX: Fix crash when closing and re-opening the a file with tagged SARIF results. [#205](https://github.com/microsoft/sarif-visualstudio-extension/issues/205)
+
 ## **v2.1.16** [SARIF Viewer](https://marketplace.visualstudio.com/items?itemName=WDGIS.MicrosoftSarifViewer)
 * BUGFIX: Offload file reading and parsing from Visual Studio's UI thread. [#160](https://github.com/microsoft/sarif-visualstudio-extension/issues/160)
 
--- a/src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs
+++ b/src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs
@@ -6,11 +6,10 @@ using System.ComponentModel;
 using System.IO;
 using Microsoft.CodeAnalysis.Sarif;
 using Microsoft.VisualStudio.Shell;
-using Microsoft.VisualStudio.Text;
 
 namespace Microsoft.Sarif.Viewer
 {
-    public abstract class CodeLocationObject : NotifyPropertyChangedObject
+    internal abstract class CodeLocationObject : NotifyPropertyChangedObject
     {
         private Region _region;
         protected ResultTextMarker _lineMarker;
@@ -27,7 +26,7 @@ namespace Microsoft.Sarif.Viewer
                 // without a region.
                 if (_lineMarker == null && Region != null)
                 {
-                    _lineMarker = new ResultTextMarker(RunId, Region, FilePath);
+                    _lineMarker = new ResultTextMarker(RunId, Region, FilePath, DefaultSourceHighlightColor);
                 }
 
                 // If the UriBaseId was populated before the marker was available, set the
@@ -195,13 +194,7 @@ namespace Microsoft.Sarif.Viewer
 
         public void ApplyDefaultSourceFileHighlighting()
         {
-            ThreadHelper.ThrowIfNotOnUIThread();
-
-            // Remove hover marker
             LineMarker?.RemoveTagHighlight();
-
-            // Add default marker instead
-            LineMarker?.AddTagHighlight(DefaultSourceHighlightColor);
         }
 
         /// <summary>
@@ -209,22 +202,7 @@ namespace Microsoft.Sarif.Viewer
         /// </summary>
         public void ApplySelectionSourceFileHighlighting()
         {
-            ThreadHelper.ThrowIfNotOnUIThread();
-
-            // Remove previous highlighting and replace with hover color
-            LineMarker?.RemoveTagHighlight();
             LineMarker?.AddTagHighlight(SelectedSourceHighlightColor);
-        }
-
-        /// <summary>
-        /// An overridden method for reacting to the event of a document window
-        /// being opened
-        /// </summary>
-        internal void AttachToDocument(ITextBuffer textBuffer)
-        {
-            ThreadHelper.ThrowIfNotOnUIThread();
-
-            LineMarker?.TryTagDocument(textBuffer);
         }
     }
 }
--- a/src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs
+++ b/src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs
@@ -183,6 +183,8 @@ namespace Microsoft.Sarif.Viewer
 
         public void ApplyDefaultSourceFileHighlighting()
         {
+            ThreadHelper.ThrowIfNotOnUIThread();
+
             // Remove hover marker
             LineMarker?.RemoveTagHighlight();
 
@@ -195,6 +197,8 @@ namespace Microsoft.Sarif.Viewer
         /// </summary>
         public void ApplySelectionSourceFileHighlighting()
         {
+            ThreadHelper.ThrowIfNotOnUIThread();
+
             // Remove previous highlighting and replace with hover color
             LineMarker?.RemoveTagHighlight();
             LineMarker?.AddTagHighlight(SelectedSourceHighlightColor);
--- a/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
+++ b/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
@@ -452,6 +452,8 @@ namespace Microsoft.Sarif.Viewer.ErrorList
 
             (dataCache.SarifErrors as List<SarifErrorListItem>).AddRange(sarifErrors);
             SarifTableDataSource.Instance.AddErrors(sarifErrors);
+            SarifLocationTagger.NotifyAllTagsChanged();
+
             return sarifErrors.Count;
         }
 
--- a/src/Sarif.Viewer.VisualStudio/ErrorList/SarifTableControlEventProcessorProvider.cs
+++ b/src/Sarif.Viewer.VisualStudio/ErrorList/SarifTableControlEventProcessorProvider.cs
@@ -46,6 +46,8 @@ namespace Microsoft.Sarif.Viewer.ErrorList
             /// </summary>
             public override void PreprocessSelectionChanged(TableSelectionChangedEventArgs e)
             {
+                ThreadHelper.ThrowIfNotOnUIThread();
+
                 // We only support single selection.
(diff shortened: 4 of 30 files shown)


src/Sarif.Viewer.VisualStudio/ErrorList/SarifTableControlEventProcessorProvider.cs
```diff
@@ -44,42 +45,37 @@ private class EventProcessor : TableControlEventProcessorBase
...
             public override void PreprocessSelectionChanged(TableSelectionChangedEventArgs e)
             {
-                SarifErrorListItem sarifResult;
-                ListView errorList = (ListView)e.SelectionChangedEventArgs.Source;
-
-                if (errorList.SelectedItems.Count != 1)
+                // We only support single selection.
+                // So if there is no selection, or more than one, clear
+                // the SARIF explorer pane (set it's data context to null) and return.
+                ITableEntryHandle selectedTableEntry = e.AddedEntries.SingleOrDefault();
```
roywill, line 51: >SingleOrDefault [](start = 70, length = 15)

The Enumerable implementation of this throws if it is not a singleton or empty.  Is that what happens here?

src/Sarif.Viewer.VisualStudio/ErrorList/SarifTableControlEventProcessorProvider.cs
```diff
@@ -44,42 +45,37 @@ private class EventProcessor : TableControlEventProcessorBase
...
+                SarifViewerPackage.SarifToolWindow.Control.DataContext = null;
+
+                if (sarifErrorListItem.HasDetails)
                 {
-                    SarifViewerPackage.SarifToolWindow.Control.DataContext = null;
+                    SarifViewerPackage.SarifToolWindow.Control.DataContext = sarifErrorListItem;
                 }
 
-                if (sarifResult.Locations?.Count > 0)
+                if (sarifErrorListItem.Locations?.Count > 0)
```
roywill, line 76: Is Count cheap or should you use Any()?

src/Sarif.Viewer.VisualStudio/ErrorList/SarifTableControlEventProcessorProvider.cs
```diff
@@ -44,42 +45,37 @@ private class EventProcessor : TableControlEventProcessorBase
...
                 {
-                    sarifResult.Locations[0].ApplyDefaultSourceFileHighlighting();
+                    sarifErrorListItem.Locations[0].ApplyDefaultSourceFileHighlighting();
```
roywill, line 78: Why not use FirstOrDefault and stop the Count and index operation

src/Sarif.Viewer.VisualStudio/ErrorList/SarifTableControlEventProcessorProvider.cs
```diff
@@ -93,26 +89,25 @@ public override void PreprocessSelectionChanged(TableSelectionChangedEventArgs e
...
                     SarifViewerPackage.SarifToolWindow.Show();
                 }
 
                 // Navigate to the source file of the first location for the defect.
-                if (sarifResult.Locations?.Count > 0)
+                if (sarifErrorListItem.Locations?.Count > 0)
                 {
-                    sarifResult.Locations[0].NavigateTo(false);
-                    sarifResult.Locations[0].ApplyDefaultSourceFileHighlighting();
+                    sarifErrorListItem.Locations[0].NavigateTo(false);
```
roywill, line 109: Indexing twice for the first element is hoping that the optimize will make it go away.  I don't think that does.  Then again FirstOrDefault() like above

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -20,8 +20,10 @@ namespace Microsoft.Sarif.Viewer.Tags
     /// Handles adding, removing, and tagging SARIF locations in a text buffer for Visual Studio integration.
     /// </summary>
     /// <remarks>
-    /// The SARIF location tagger is created per instance of a Visual Studio <see cref="ITextBuffer"/>. The tagger is disposed
-    /// when the text buffer is no longer needed.
+    /// The SARIF location tagger is created per instance of a Visual Studio <see cref="ITextView"/>. The tagger is disposed
+    /// when the text view is no longer used. However the underlying "tags" are associated with a text buffer
+    /// which can outlast an <see cref="ITextView"/>. For instance, a split window view on the same document.
```
roywill, line 25: Use <para> to keep paragraphs

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -214,21 +243,16 @@ public static void RemoveAllTags()
...
+            List<SarifLocationTagger> taggersToDispose;
+            using (SarifTaggersLock.EnterWriteLock())
             {
-                // There is no need to collect the tags from the run Id to SARIF tag dictionary as the
-                // objects exist in both dictionaries and only need to be disposed once.
-                tagsToDispose.AddRange(SourceCodeFileToSarifTags.Values.SelectMany(sarifTags => sarifTags));
-
-                SourceCodeFileToSarifTags.Clear();
-                RunIdToSarifTags.Clear();
+                taggersToDispose = new List<SarifLocationTagger>(SarifTaggers);
```
roywill, line 249: Why duplicate instead of swap out?

roywill (approved):
:shipit:

sherryyshi (approved):
:shipit:

src/ReleaseHistory.md
```diff
@@ -1,4 +1,7 @@
 # SARIF Viewer Visual Studio extension Release History
+## **v2.1.17** [SARIF Viewer](https://marketplace.visualstudio.com/items?itemName=WDGIS.MicrosoftSarifViewer)
+* BUGFIX: Fix crash when closing and re-opening the a file with tagged SARIF results. [#205](https://github.com/microsoft/sarif-visualstudio-extension/issues/205)
```
lgolding, line 3: >the a [](start = 48, length = 5)

"the a" => "a"

lgolding:
>             /// Does not show the tool window if it is not already open. Displaying of the tool window is handed by PreprocessNavigate.

"handed" => "handled", while you're in here.

---
Refers to: src/Sarif.Viewer.VisualStudio/ErrorList/SarifTableControlEventProcessorProvider.cs:44 in 52dc167. [](commit_id = 52dc167bbc53c3e7134782860e5d9ec8212f37d6, deletion_comment = False)

src/Sarif.Viewer.VisualStudio/ErrorList/SarifTableControlEventProcessorProvider.cs
lgolding: +1 on this. `FirstOrDefault` would be safer (unless you want to find out the hard way on the day we start supporting multiple selection).

---
In reply to: [497147372](https://github.com/microsoft/sarif-visualstudio-extension/pull/206#discussion_r497147372) [](ancestors = 497147372)

src/Sarif.Viewer.VisualStudio/ErrorList/SarifTableControlEventProcessorProvider.cs
```diff
@@ -44,42 +45,37 @@ private class EventProcessor : TableControlEventProcessorBase
...
-
-                if (!TryGetSarifResult(entryHandle, out sarifResult))
+                if (!TryGetSarifResult(selectedTableEntry, out SarifErrorListItem sarifErrorListItem))
                 {
                     // The selected item is not a SARIF result. Clear the SARIF Explorer.
                     SarifViewerPackage.SarifToolWindow.Control.DataContext = null;
                     return;
                 }
 
                 // Set the current sarif error in the manager so we track code locations.
```
lgolding, line 65: >sarif [](start = 35, length = 5)

"SARIF", while you're in here.

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTaggerProvider.cs
```diff
@@ -44,52 +47,33 @@ internal class SarifLocationTaggerProvider : IViewTaggerProvider, ITextViewCreat
...
+                return null;
             }
 
-            sarifLocationTagger = textBuffer.Properties.GetOrCreateSingletonProperty(delegate
-            {
-                return new SarifLocationTagger(textBuffer, this.PersistentSpanFactory);
-            });
+            SarifLocationTagger tagger = new SarifLocationTagger(textView, textBuffer, this.PersistentSpanFactory);
+
+            // Text view creation comes before the ask to create a tagger.
```
lgolding, line 71: >ask [](start = 51, length = 3)

Ultra-nit: "ask" => "request"

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -20,8 +20,10 @@ namespace Microsoft.Sarif.Viewer.Tags
     /// Handles adding, removing, and tagging SARIF locations in a text buffer for Visual Studio integration.
     /// </summary>
     /// <remarks>
-    /// The SARIF location tagger is created per instance of a Visual Studio <see cref="ITextBuffer"/>. The tagger is disposed
-    /// when the text buffer is no longer needed.
+    /// The SARIF location tagger is created per instance of a Visual Studio <see cref="ITextView"/>. The tagger is disposed
+    /// when the text view is no longer used. However the underlying "tags" are associated with a text buffer
```
lgolding, line 24: >r [](start = 108, length = 1)

Add period.

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
```diff
@@ -20,8 +20,10 @@ namespace Microsoft.Sarif.Viewer.Tags
...
+    /// which can outlast an <see cref="ITextView"/>. For instance, a split window view on the same document.
+    /// That results in 2 <see cref="ITextView"/> instances against 1 underlying <see cref="ITextBuffer"/>.
```
lgolding, line 26: >2 [](start = 24, length = 1)

2 => "multiple"

lgolding (approved):
:shipit:

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTaggerProvider.cs
GabeDeBacker: This call didn't even need to exist anymore. I'm glad you commented this or we would have been listening to caret events twice!

---
In reply to: [497831993](https://github.com/microsoft/sarif-visualstudio-extension/pull/206#discussion_r497831993) [](ancestors = 497831993)

src/Sarif.Viewer.VisualStudio/Tags/SarifLocationTagger.cs
GabeDeBacker: The number of taggers is going to be "about" equal to the number of documents the user has opened. So that number is small, so I'd prefer to leave this as a static readonly list.

---
In reply to: [497151946](https://github.com/microsoft/sarif-visualstudio-extension/pull/206#discussion_r497151946) [](ancestors = 497151946)

src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs
```diff
@@ -195,6 +195,8 @@ public void NavigateTo(bool usePreviewPane = true)
 
         public void ApplyDefaultSourceFileHighlighting()
         {
+            ThreadHelper.ThrowIfNotOnUIThread();
```
jeffersonking, line 198: `ThreadHelper.ThrowIfNotOnUIThread()` is also present at least 2 frames down the call graph (and also one up). What is the purpose of the redundancy?

GabeDeBacker: The analyzers pop that this must be called. I would also prefer to not have the redundancy.

---
In reply to: [499329815](https://github.com/microsoft/sarif-visualstudio-extension/pull/206#discussion_r499329815) [](ancestors = 499329815)

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -452,6 +452,8 @@ private int WriteRunToErrorList(Run run, string logFilePath)
 
             (dataCache.SarifErrors as List<SarifErrorListItem>).AddRange(sarifErrors);
             SarifTableDataSource.Instance.AddErrors(sarifErrors);
+            SarifLocationTagger.NotifyAllTagsChanged();
```
GabeDeBacker, line 455: Comment why this is here..

This causes already open "text views" to be tagged when SARIF logs are processed after a view is opened.

GabeDeBacker:
Abandoning this in favor of a cleaner approach.

PR #207, opened 2020-09-29
Fix #201: Double-clicking result does not navigate to source file

### LINKED ISSUE #201: Double-clicking result does not navigate to source file
A SARIF log file contains
```
      "originalUriBaseIds": {
        "TARGET": {
          "uri": "file:///c:/MyProject/"
        }
      },
      ...
      "results: [
        {
          "locations": [
            {
              "physicalLocation": {
                "artifactLocation": {
                  "uri": "/Client/Example.cs",
                  "uriBaseId": "TARGET"
                },
```

The file c:\MyProject\Client\Example.cs exists on my machine. But double-clicking the result does not open the file.

In the VS Code viewer, double-clicking the result _does_ open the file and highlight the correct line.

[I will provide a minimal repro for this. The file in which I observed this contains sensitive data.]

@michaelcfanning FYI

lgolding:
The proximate cause is that `uriBaseId` was null in a call to `CodeAnalysisResultManager.TryRebaselineAllSarifErrors`. I don't know the root cause (that is, I don't know the last time this feature worked, or which code change broke it). But it's straightforward to supply the `uriBaseId`.

Also:
- Fix a backwards logical condition that prevented `CodeLocationObject.UriBaseId` from being set. That bug didn't seem to cause any observable problems (in particular, it was not the cause of _this_ bug). I don't know if that property used to percolate down to the `CodeAnalysisResultManager`, but as the code stands today, it does not.
- Clean up a few IDE suggestions.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs
+++ b/src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs
@@ -22,9 +22,21 @@ namespace Microsoft.Sarif.Viewer
             get
             {
                 // Not all locations have regions. Don't try to mark the locations that don't.
+                //
+                // PROBLEM: This means we can't double-click to open a file containing a result
+                // without a region.
                 if (_lineMarker == null && Region != null)
                 {
                     _lineMarker = new ResultTextMarker(RunId, Region, FilePath);
+                }
+
+                // If the UriBaseId was populated before the marker was available, set the
+                // marker's UriBaseId property now. The marker's UriBaseId is used to resolve
+                // relative paths to absolute paths when the user double-clicks a result in
+                // the Error List window.
+                if (_lineMarker != null && UriBaseId != null)
+                {
+                    _lineMarker.UriBaseId = UriBaseId;
                 }
 
                 return _lineMarker;
@@ -94,7 +106,7 @@ namespace Microsoft.Sarif.Viewer
             }
             set
             {
-                if (string.Equals(value, this._uriBaseId))
+                if (!string.Equals(value, this._uriBaseId))
                 {
                     this._uriBaseId = value;
 
--- a/src/Sarif.Viewer.VisualStudio/Models/SarifErrorListItem.cs
+++ b/src/Sarif.Viewer.VisualStudio/Models/SarifErrorListItem.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Sarif.Viewer
 {
     public class SarifErrorListItem : NotifyPropertyChangedObject
     {
-        private int _runId;
+        private readonly int _runId;
         private string _fileName;
         private ToolModel _tool;
         private RuleModel _rule;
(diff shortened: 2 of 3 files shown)


src/Sarif.Viewer.VisualStudio/CodeLocationObject.cs
```diff
@@ -94,7 +106,7 @@ internal virtual string UriBaseId
             }
             set
             {
-                if (string.Equals(value, this._uriBaseId))
+                if (!string.Equals(value, this._uriBaseId))
```
lgolding, line 109: >! [](start = 20, length = 1)

This backwards test prevented `CodeLocationObject.UriBaseId` from being set, and that _is_ a regression. But this isn't the `UriBaseId` whose null-ness caused the bug.

src/Sarif.Viewer.VisualStudio/Models/SarifErrorListItem.cs
```diff
@@ -490,6 +489,10 @@ public ResultTextMarker LineMarker
                 if (_lineMarker == null && Region != null && Region.StartLine > 0)
                 {
                     _lineMarker = new ResultTextMarker(_runId, Region, FileName);
+                    if (Locations.Count > 0)
+                    {
+                        _lineMarker.UriBaseId = Locations[0].UriBaseId;
```
lgolding, line 494: >UriBaseId [](start = 36, length = 9)

_This_ is the fix.

jeffersonking: Prefer:
```
_lineMarker = new ResultTextMarker(_runId, Region, FileName) {
    UriBaseId = Locations[0]?.UriBaseId,
};
```

lgolding: That doesn't actually work. `[0]?` doesn't mean "return `null` if the 0th element isn't there". It means "return `null` if `[0]` is `null`". If there is no 0th element, this throws "index out of range". I've made this mistake before :-)

---
In reply to: [498665970](https://github.com/microsoft/sarif-visualstudio-extension/pull/207#discussion_r498665970) [](ancestors = 498665970)

jeffersonking: Thanks. Obvious after you point it out. What are your thoughts on:
```
_lineMarker = new ResultTextMarker(_runId, Region, FileName) {
    UriBaseId = Locations.FirstOrDefault()?.UriBaseId,
};
```

lgolding: Sure, that's fine. It would have to be `Locations?.FirstOrDefault()?.UriBaseId` but yes.

---
In reply to: [499018240](https://github.com/microsoft/sarif-visualstudio-extension/pull/207#discussion_r499018240) [](ancestors = 499018240)

jeffersonking: Ah yes. Likewise the original code would have been `Locations?.Count` also right?

lgolding: LOL yes.

---
In reply to: [499028535](https://github.com/microsoft/sarif-visualstudio-extension/pull/207#discussion_r499028535) [](ancestors = 499028535)

GabeDeBacker (approved):

Issue #218, opened 2020-10-12
Selection of result location in SARIF Explorer window is off by one column


lgolding:
- Open a SARIF log file.
- Double click on a result to navigate to the result location in the editor window.
- The SARIF Explorer displays the result location, but does not highlight it (this is #xxx to be filed by @GabeDeBacker).
- Now click on the line with the error, but a few columns before the start of the error.
- Right arrow one column at a time.

Expected: When the caret reaches the left-hand edge of the error location, the location should highlight in the SARIF Explorer.

Actual: The location highlights in the SARIF explorer one column _before_ the left-hand edge of the error location.

PR #219, opened 2020-10-13
Address caret tag entry (and SARIF explorer window selection) being off by one column

GabeDeBacker:
This fix addresses issue #218 where the tags in the documents representing SARIF locations is a bit to eager (off by one column).

The issue was rather simple. A caret's insertion point is a zero-length span, not a span of one character. The code was incorrectly adding one to the caret position to look for intersection with tags rather than just the "insertion point".

### DIFF
--- a/src/ReleaseHistory.md
+++ b/src/ReleaseHistory.md
@@ -1,7 +1,4 @@
 # SARIF Viewer Visual Studio extension Release History
-## **v2.1.19** [SARIF Viewer](https://marketplace.visualstudio.com/items?itemName=WDGIS.MicrosoftSarifViewer)
-* BUGFIX: Fix tag highlighting (and selection in SARIF explorer window) being off by one column. [#218](https://github.com/microsoft/sarif-visualstudio-extension/issues/218)
-
 ## **v2.1.18** [SARIF Viewer](https://marketplace.visualstudio.com/items?itemName=WDGIS.MicrosoftSarifViewer)
 * BUGFIX: Fix crash when closing and re-opening a file with tagged SARIF results. [#205](https://github.com/microsoft/sarif-visualstudio-extension/issues/205)
 * BUGFIX: Ensure that SARIF explorer properly selects code-flow location when navigating caret in editor. [#198](https://github.com/microsoft/sarif-visualstudio-extension/issues/198)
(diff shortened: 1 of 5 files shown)


src/build.props
```diff
@@ -19,7 +19,7 @@
     <Company Condition=" '$(Company)' == '' ">Microsoft</Company>
     <Product Condition=" '$(Product)' == '' ">Microsoft SARIF Viewer for Visual Studio</Product>
     <Copyright Condition=" '$(Copyright)' == '' ">© Microsoft Corporation. All rights reserved.</Copyright>
-    <VersionPrefix Condition=" '$(VersionPrefix)' == ''">2.1.18</VersionPrefix>
+    <VersionPrefix Condition=" '$(VersionPrefix)' == ''">2.1.19</VersionPrefix>
```
lgolding, line 22: >19 [](start = 61, length = 2)

We never published 2.18 even to insiders gallery so no need for this.

src/ReleaseHistory.md
```diff
@@ -1,4 +1,7 @@
 # SARIF Viewer Visual Studio extension Release History
+## **v2.1.19** [SARIF Viewer](https://marketplace.visualstudio.com/items?itemName=WDGIS.MicrosoftSarifViewer)
+* BUGFIX: Fix tag highlighting (and selection in SARIF explorer window) being off by one column. [#218](https://github.com/microsoft/sarif-visualstudio-extension/issues/218)
```
lgolding, line 3: >column [](start = 89, length = 6)

Again, since we didn't publish 2.18 (yet) noone saw this, so no need to mention it.

src/Sarif.Viewer.VisualStudio/Tags/TextViewCaretListener.cs
```diff
@@ -64,13 +64,7 @@ private void UpdateAtCaretPosition(CaretPosition caretPosition)
...
             SnapshotPoint caretSnapshotPoint = caretPosition.BufferPosition;
 
-            // If the caret is at the end of the file, then the caret cannot be inside a tag.
-            if (caretSnapshotPoint.Position + 1 >= this.textView.TextBuffer.CurrentSnapshot.Length)
-            {
-                return;
-            }
-
-            NormalizedSnapshotSpanCollection normalizedSnapshotSpanCollection = new NormalizedSnapshotSpanCollection(new SnapshotSpan(caretSnapshotPoint, 1));
+            NormalizedSnapshotSpanCollection normalizedSnapshotSpanCollection = new NormalizedSnapshotSpanCollection(new SnapshotSpan(caretSnapshotPoint, caretSnapshotPoint));
```
lgolding, line 67: >caretSnapshotPoint [](start = 154, length = 18)

Ah, without the ", 1"

lgolding (approved):
:shipit:

src/Sarif.Viewer.VisualStudio/Tags/TextViewCaretListener.cs
GabeDeBacker: And since it's not clear from the overloaded constructor... This is a start, end not a length.

So I'll used named parameters.

---
In reply to: [503598671](https://github.com/microsoft/sarif-visualstudio-extension/pull/219#discussion_r503598671) [](ancestors = 503598671)

Issue #220, opened 2020-10-13
First time navigation to a "location" (that is not a code-flow) does not show selection in SARIF explorer


GabeDeBacker:
Code flows have special logic in their XAML implementation (callTreeView.cs) for ensuring that there is at least one selected item in the explorer view.

This same logic does not exist for locations or related locations.

PR #221, opened 2020-10-13
Send an initial update of the caret position for newly created text views

GabeDeBacker:
This address issue #220 .

When a new text view is created, and a tagger is requested, the extension creates a new text-view listener for it. Consumers of that listener (such as the SARIF explorer window) use the events from the text-view listener to properly show selected items in their views.

During initial creation of the text view (and its corresponding listener), an initial set of events was not being sent for the caret position in the text view. Due to this, the SARIF explorer XAML views could not properly update their initial selection.

The fix was to send an initial events for the starting caret location for the text view.

While testing the fix, I also realized that when you clear the SARIF errors from the list, or do anything else that results in a SARIF result not being valid (such as closing a SARIF log file that was directly opened, etc.), that the SARIF explorer continued to show content for a SARIF result that has been disposed. That's not so good. This was also addressed.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/Tags/TextViewCaretListener.cs
+++ b/src/Sarif.Viewer.VisualStudio/Tags/TextViewCaretListener.cs
@@ -57,7 +57,7 @@ namespace Microsoft.Sarif.Viewer.Tags
             // is the main UI thread, then it runs this task synchronously.
             // What we want is for the constructor above to return and the consumer
             // of the new object to be able to subscribe to events before the initial
-            // caret position update (and it's corresponding events) are sent.
+            // caret position update (and its corresponding events) are sent.
             // So we yield to allow the constructor call stack to unwind before sending
             // the initial caret position events.
             await System.Threading.Tasks.Task.Yield();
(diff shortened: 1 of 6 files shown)


src/Sarif.Viewer.VisualStudio/Tags/TextViewCaretListener.cs
```diff
@@ -41,6 +51,20 @@ public TextViewCaretListener(ITextView textView, ITagger<T> tagger)
         /// </summary>
         public event EventHandler<TagInCaretChangedEventArgs> CaretLeftTag;
 
+        private async System.Threading.Tasks.Task UpdateInitialCaretPositionAsync()
+        {
+            // When ThreadHelper.JoinableTaskFactory.RunAsync is invoked, if the current thread
+            // is the main UI thread, then it runs this task synchronously.
+            // What we want is for the constructor above to return and the consumer
+            // of the new object to be able to subscribe to events before the initial
+            // caret position update (and it's corresponding events) are sent.
```
lgolding, line 60: >it's [](start = 42, length = 4)

"its"

src/Sarif.Viewer.VisualStudio/Tags/TextViewCaretListener.cs
```diff
@@ -60,16 +84,28 @@ private void Caret_PositionChanged(object sender, CaretPositionChangedEventArgs
...
 
-            List<ISarifLocationTag> tagsCaretIsCurrentlyIn = this.tagger.GetTags(normalizedSnapshotSpanCollection).
+            // Handling the aggregate focus allows tags to have their highlights removed when the focus moves from
+            // one document to another rather than having a bunch of views with highlights that don't correspond
+            // with what's being selected in the solution explorer.
+            List<ISarifLocationTag> tagsCaretIsCurrentlyIn = (this.textView.HasAggregateFocus ? this.tagger.GetTags(normalizedSnapshotSpanCollection).
                 Where(tag => tag.Tag is ISarifLocationTag).
-                Select(tag => tag.Tag as ISarifLocationTag).
-                ToList();
+                Select(tag => tag.Tag as ISarifLocationTag) : Enumerable.Empty<ISarifLocationTag>()).ToList();
```
lgolding, line 108: >; [](start = 109, length = 1)

I had trouble following this because the choice of line breaks. Consider:
```
            List<ISarifLocationTag> tagsCaretIsCurrentlyIn = (this.textView.HasAggregateFocus
                ? this.tagger.GetTags(normalizedSnapshotSpanCollection)
                    Where(tag => tag.Tag is ISarifLocationTag).
                    Select(tag => tag.Tag as ISarifLocationTag)
                : Enumerable.Empty<ISarifLocationTag>()).ToList();
```

lgolding (approved):
:shipit:

Issue #222, opened 2020-10-13
While SARIF logs are being processed via the API, focus keeps getting set to the error list (stealing focus)


GabeDeBacker:
When the extension's open log APIs is being invoked, and several log files are being processed (which can take a long time depending on the number of files), focus is repeatedly set back to the error list.

This is incredibly annoying as a user because they cannot continue to use the editor, solution explorer, etc. while the log files are being processed.

It's okay to "show" the error list, but focus should not be "set" to it. Keyboard focus should be left where it is.

PR #223, opened 2020-10-13
Don't activate the error list when parsing error logs, only show it. …

GabeDeBacker:
This fix address issue #222 which was causing keyboard focus to be taken away from the user when an extension called the open log APIs.

…It is annoying to have keyboard focus yanked away from you while log files are processing in the background.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
+++ b/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
@@ -377,7 +377,7 @@ namespace Microsoft.Sarif.Viewer.ErrorList
 
             if (hasResults)
             {
-                SdkUIUtilities.ShowToolWindowAsync(new Guid(ToolWindowGuids80.ErrorList), activate: false).FileAndForget("Microsoft/SARIF/Viewer/ShowErrorList");
+                SdkUIUtilities.ShowToolWindowAsync(new Guid(ToolWindowGuids80.ErrorList), activate: false).FileAndForget(Constants.FileAndForgetFaultEventNames.ShowErrorList);
             }
             else if (showMessageOnNoResults)
             {
--- a/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
+++ b/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
@@ -215,12 +215,6 @@ namespace Microsoft.Sarif.Viewer.ErrorList
             }
 
             await ProcessSarifLogAsync(log, outputPath, showMessageOnNoResults: promptOnLogConversions, cleanErrors: cleanErrors).ConfigureAwait(continueOnCapturedContext: false);
-
-            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
-            if (AsyncPackage.GetGlobalService(typeof(DTE)) is DTE2 dte)
-            {
-                dte.ExecuteCommand("View.ErrorList");
-            }
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
```diff
@@ -388,6 +388,40 @@ public static bool TryGetActiveViewForTextBuffer(ITextBuffer textBuffer, out IWp
...
+        /// Locates a tool window and request that it be shown.
+        /// </summary>
+        /// <param name="toolWindowGuid">A tool window identifier, typically one from <see cref="ToolWindowGuids80"/> such as the error list.</param>
+        /// <param name="activate">Indicates whether to activate (take keyboard focus) when showing the tool window.</param>
+        /// <returns>Returns a task that when completed indicates if the tool window was shown.</returns>
+        public static async System.Threading.Tasks.Task<bool> ShowToolWindowAsync(Guid toolWindowGuid, bool activate)
+        {
+            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+            IVsUIShell uiShell = Microsoft.VisualStudio.Shell.Package.GetGlobalService(typeof(SVsUIShell)) as IVsUIShell;
+            if (uiShell == null)
```
jeffersonking, line 401: Use null-conditional below to skip this check?

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -215,12 +215,6 @@ public static async System.Threading.Tasks.Task ProcessLogFileAsync(string fileP
             }
 
             await ProcessSarifLogAsync(log, outputPath, showMessageOnNoResults: promptOnLogConversions, cleanErrors: cleanErrors).ConfigureAwait(continueOnCapturedContext: false);
-
-            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
-            if (AsyncPackage.GetGlobalService(typeof(DTE)) is DTE2 dte)
-            {
-                dte.ExecuteCommand("View.ErrorList");
```
lgolding, line 222: >ErrorList [](start = 41, length = 9)

So if we don't do this, when does the viewer see the error list? I understand you don't want to pull focus to it, but if (1) if it wasn't previously opened you do want to open it, and (2) if it's open but not the currently selected tab in -- in whatever you call that collection of windows at the bottom -- then you do want to make that the selected tab. IMO.

lgolding: I see your PR description explicitly says "It's ok to "show" the Error list"... so I will keep looking for the code that does that.

---
In reply to: [504119046](https://github.com/microsoft/sarif-visualstudio-extension/pull/223#discussion_r504119046) [](ancestors = 504119046)

lgolding (approved):
:shipit:

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -381,9 +375,12 @@ internal static async System.Threading.Tasks.Task ProcessSarifLogAsync(SarifLog
...
             }
 
-            if (!hasResults && showMessageOnNoResults)
+            if (hasResults)
+            {
+                SdkUIUtilities.ShowToolWindowAsync(new Guid(ToolWindowGuids80.ErrorList), activate: false).FileAndForget("Microsoft/SARIF/Viewer/ShowErrorList");
+            }
+            else if (showMessageOnNoResults)
             {
-                await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
```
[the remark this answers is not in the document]
GabeDeBacker: No, it is already being done at the top of this function.

---
In reply to: [504117780](https://github.com/microsoft/sarif-visualstudio-extension/pull/223#discussion_r504117780) [](ancestors = 504117780)

src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
GabeDeBacker: This was lifted right from VS's codebase from their utility class.

Cleaned it up.

---
In reply to: [504106375](https://github.com/microsoft/sarif-visualstudio-extension/pull/223#discussion_r504106375) [](ancestors = 504106375)

jeffersonking (approved):

PR #225, opened 2020-10-14
Trying to show the error list during unit-testing is not necessary

GabeDeBacker:
And causes a null-deref since we have no joinable task factory and no VS shell.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
+++ b/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
@@ -377,7 +377,10 @@ namespace Microsoft.Sarif.Viewer.ErrorList
 
             if (hasResults)
             {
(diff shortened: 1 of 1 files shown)


src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -377,7 +377,10 @@ internal static async System.Threading.Tasks.Task ProcessSarifLogAsync(SarifLog
 
             if (hasResults)
             {
-                SdkUIUtilities.ShowToolWindowAsync(new Guid(ToolWindowGuids80.ErrorList), activate: false).FileAndForget(Constants.FileAndForgetFaultEventNames.ShowErrorList);
+                if (!SarifViewerPackage.IsUnitTesting) // We cannot show UI during unit-tests.
```
lgolding, line 380: >tests [](start = 88, length = 5)

Yup. In future we'd want to fire an event that some UI service could consume, but this is just fine for now.

lgolding (approved):
:shipit:

Issue #216, opened 2020-10-09
Telemetry drops events if package not initialized


lgolding:
Our extension initializes the telemetry provider when VS initializes (calls `InitializeAsync` on) the package. But if you do **File**, **Open** on a SARIF file without previously having done anything that would _cause_ VS to initialize the package (for example, by invoking one of the extensions menu items on the **Tools** menu), then the "open file" telemetry event is not sent because the telemetry provider hasn't been initialized.

Call stack:
```
Microsoft.Sarif.Viewer.dll!Microsoft.Sarif.Viewer.TelemetryProvider.WriteEvent(Microsoft.Sarif.Viewer.TelemetryEvent eventType, System.Collections.Generic.Dictionary<string, string> properties) Line 115 C#
Microsoft.Sarif.Viewer.dll!Microsoft.Sarif.Viewer.TelemetryProvider.WriteEvent(Microsoft.Sarif.Viewer.TelemetryEvent eventType, System.Collections.Generic.KeyValuePair<string, string>[] pairs) Line 105 C#
Microsoft.Sarif.Viewer.dll!Microsoft.Sarif.Viewer.ErrorList.ErrorListService.ProcessSarifLogAsync(Microsoft.CodeAnalysis.Sarif.SarifLog sarifLog, string logFilePath, bool showMessageOnNoResults, bool cleanErrors) Line 376 C#
```

Note that ??? is MEF-exported... TODO FINISH THIS

mpande98:
**Medium priority**: User can interact with services without loading extension itself. Extension not initialized, so telemetry not initialized. 

**High priority**: Where does the data in app insights go? Where does the telemetry go??

PR #226, opened 2020-10-15
Ensure telemetry provider is initialized regardless of package initialization status

GabeDeBacker:
This change address issue #216 and ensure that telemetry is properly sent from the SARIF extension.

The implementation uses MEF to export a tiny telemetry provider type\service that is constructed on demand when the first telemetry event is fired. The reason MEF is used is that it provides a way to "flush" and "dispose" the telemetry client on Visual Studio shutdown (confirmed with debugger and Visual Studio developers).

The implementation provides an easy and consistent way to add new telemetry points in a way that allows for easy "auditing" of the telemetry as well as consistent naming.

How verified:
Obviously stepped through the debugger 👍
Verified that the extension telemetry was flowing to Azure and able to be queried from the Azure portal.


lgolding:
>             /// Used when showing the error list fails.

Can we append "/Failed" to each of the appropriate names. In case there's something else we want to say about opening the error list (/Succeeded, /InProgress...?)

---
Refers to: src/Sarif.Viewer.VisualStudio/Constants.cs:13 in ab7995a. [](commit_id = ab7995a4f64bd6bccfed771e4ec623838dbddc08, deletion_comment = False)

src/Sarif.Viewer.VisualStudio/SarifViewerPackage.cs
```diff
@@ -1,32 +1,24 @@
 ﻿// Copyright (c) Microsoft. All rights reserved. 
 // Licensed under the MIT license. See LICENSE file in the project root for full license information. 
 
+using Microsoft.Sarif.Viewer.ErrorList;
+using Microsoft.Sarif.Viewer.Services;
+using Microsoft.Sarif.Viewer.Tags;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Text.Tagging;
```
lgolding, line 8: Ordering

src/Sarif.Viewer.VisualStudio/Services/LoadSarifLogService.cs
```diff
@@ -49,6 +53,8 @@ public void LoadSarifLogs(IEnumerable<string> paths)
         /// <inheritdoc/>
         public void LoadSarifLogs(IEnumerable<string> paths, bool promptOnSchemaUpgrade)
         {
+            Telemetry.Events.LoadSarifLogApiInvoked();
```
lgolding, line 56: >LoadSarifLogApiInvoked [](start = 29, length = 22)

Shouldn't this have arguments, at least the `path`?

lgolding: Come to think of it `path` for the first two overloads, `paths` (perhaps string-joined?) for this one.

---
In reply to: [505741594](https://github.com/microsoft/sarif-visualstudio-extension/pull/226#discussion_r505741594) [](ancestors = 505741594)

src/Sarif.Viewer.VisualStudio/Telemetry/Events.cs
```diff
@@ -0,0 +1,76 @@
...
+    internal class Events
+    {
+        /// <summary>
+        /// Does nothing.
+        /// </summary>
+        /// <remarks>
+        /// The reason we call track event directly here is that
+        /// we are in the middle of MEF composition\creation <see cref="TelemetryProvider.TelemetryProviderService"/>, if we called
+        /// <see cref="TelemetryProvider.TrackEvent(string)"/> that would end up being a recursive MEF composition call (and a
+        /// MEF composition exception). This method is purely used for event naming purposes.
```
lgolding, line 27: >This method is purely used for event naming purposes. [](start = 40, length = 53)

Note to self: Understand what this means.

src/Sarif.Viewer.VisualStudio/Telemetry/Events.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// </remarks>
+        public static void ExtensionLoaded() => throw new NotImplementedException();
+
+        /// <summary>
+        /// Tracks an event indicating that a SARF log file was run was loaded into the extension.
```
lgolding, line 32: >was run was loaded [](start = 60, length = 18)

"was run was loaded" => "was loaded"

src/Sarif.Viewer.VisualStudio/Telemetry/Events.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// </summary>
+        /// <param name="toolName">The name of the tool that provided the SARIF log.</param>
+        public static void LogFileRunCreatedByToolName(string toolName) =>
+            TelemetryProvider.TrackEvent<Events>(new Dictionary<string, string> { { nameof(toolName), toolName } } );
+
+        /// <summary>
+        /// Tracks an event indicating that a document was opened as a result of navigation either through
+        /// a call to <see cref="ResultTextMarker.NavigateTo(bool, bool)"/> or <see cref="CodeLocationObject.NavigateTo(bool, bool)"/>.
+        /// </summary>
+        public static void TaskItemDocumentOpened() =>
```
lgolding, line 42: >TaskItem [](start = 27, length = 8)

What does "TaskItem" mean in this name?

src/Sarif.Viewer.VisualStudio/Telemetry/Events.cs
```diff
@@ -0,0 +1,76 @@
...
+            TelemetryProvider.TrackEvent<Events>();
+
+        /// <summary>
+        /// Indicates that a SARIF log file was opened (imported) through the tools menu.
```
lgolding, line 46: >menu [](start = 84, length = 4)

We need one for **File**, **Open** as well.

src/Sarif.Viewer.VisualStudio/Telemetry/Events.cs
```diff
@@ -0,0 +1,76 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// Contains the methods for firing telemetry events from the SARIF viewer extension.
+    /// </summary>
+    /// <remarks>
+    /// The documentation in this class is super useful for understanding where this telemetry is fired.
+    /// The documentation should also contain "why" the telemetry is being fired, i.e., why is this data needed?
```
jeffersonking, line 14: You mention "the why", but I don't see it being adhered to below.

lgolding: Now I understand, it's because you're using `[CallerMemberName]`.

---
In reply to: [505749749](https://github.com/microsoft/sarif-visualstudio-extension/pull/226#discussion_r505749749) [](ancestors = 505749749)

src/Sarif.Viewer.VisualStudio/Telemetry/Events.cs
lgolding: Now I understand, it's because you're using `[CallerMemberName]`.

---
In reply to: [505746342](https://github.com/microsoft/sarif-visualstudio-extension/pull/226#discussion_r505746342) [](ancestors = 505746342)

src/Sarif.Viewer.VisualStudio/Telemetry/Events.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// <summary>
+        /// Does nothing.
+        /// </summary>
+        /// <remarks>
+        /// The reason we call track event directly here is that
+        /// we are in the middle of MEF composition\creation <see cref="TelemetryProvider.TelemetryProviderService"/>, if we called
+        /// <see cref="TelemetryProvider.TrackEvent(string)"/> that would end up being a recursive MEF composition call (and a
+        /// MEF composition exception). This method is purely used for event naming purposes.
+        /// </remarks>
+        public static void ExtensionLoaded() => throw new NotImplementedException();
```
lgolding, line 29: >throw new NotImplementedException(); [](start = 48, length = 36)

BUG: Presumably you meant to call `TelemetryClient.TrackEvent` directly here, and the comment suggests.

GabeDeBacker: No, actually I did not... This should not be called, it is here for naming purposes only (see comment)

GabeDeBacker: No, actually I did not... This should not be called, it is here for naming purposes only (see comment)

---
In reply to: [505751525](https://github.com/microsoft/sarif-visualstudio-extension/pull/226#discussion_r505751525) [](ancestors = 505751525)

src/Sarif.Viewer.VisualStudio/Telemetry/Events.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// Indicates that a SARIF log file was opened (imported) through the tools menu.
+        /// </summary>
+        /// <param name="toolName">The name of the tool that produced the original static analysis file.</param>
+        public static void LogFileOpenedByMenuCommand(string toolName) =>
+            TelemetryProvider.TrackEvent<Events>(new Dictionary<string, string> { { nameof(toolName), toolName } });
+
+        /// <summary>
+        /// Called when a call to <see cref="Services.CloseSarifLogService.CloseAllSarifLogs"/> is invoked.
+        /// </summary>
+        public static void CloseAllSarifLogsApiInvoked() =>
```
jeffersonking, line 55: I am concerned about scaling this boilerplate pattern. The majority of the members here are already like this. The comment doesn't tell me something I don't know. It also creates an additional member that needs to be named that isn't meaningfully different than the caller.

src/Sarif.Viewer.VisualStudio/Telemetry/Events.cs
lgolding: The comment says "The reason we call track event directly here..." but you don't call track event here at all.

---
In reply to: [505752538](https://github.com/microsoft/sarif-visualstudio-extension/pull/226#discussion_r505752538) [](ancestors = 505752538)

src/Sarif.Viewer.VisualStudio/Telemetry/Events.cs
GabeDeBacker: All I can say is this is the technique we use in most of our extensions and it works very well. The comment should include "where" it is invoked and "why" is is invoked.

Right now I would say that there are no "whys" for any of these.

We shouldn't be firing telemetry for data we will never consume.

So, we have the option to check in with NO telemetry points until we understand "why" we want this data.

For example, it isn't interesting to know that someone called an API, or that something was executed unless you plan to take action based on that data. Windows (for example) takes a hard stance on this since telemetry takes not only user bandwidth but storage space on Azure.



---
In reply to: [505753537](https://github.com/microsoft/sarif-visualstudio-extension/pull/226#discussion_r505753537) [](ancestors = 505753537)

src/Sarif.Viewer.VisualStudio/Telemetry/Events.cs
GabeDeBacker: You are correct. And I will expand the comment and probably remove ALL telemetry points.

Right now I would say that there are no "whys" for any of these.

We shouldn't be firing telemetry for data we will never consume.

So, we have the option to check in with NO telemetry points until we understand "why" we want this data.

For example, it isn't interesting to know that someone called an API, or that something was executed unless you plan to take action based on that data. Windows (for example) takes a hard stance on this since telemetry takes not only user bandwidth but storage space on Azure.

---
In reply to: [505749749](https://github.com/microsoft/sarif-visualstudio-extension/pull/226#discussion_r505749749) [](ancestors = 505749749)

src/Sarif.Viewer.VisualStudio/Services/LoadSarifLogService.cs
GabeDeBacker: No, we absolutely cannot send the path as it will almost always have a userid in it.

---
In reply to: [505742115](https://github.com/microsoft/sarif-visualstudio-extension/pull/226#discussion_r505742115) [](ancestors = 505742115,505741594)

src/Sarif.Viewer.VisualStudio/Telemetry/Events.cs
lgolding: Ok, I see that the call is in `TelemetryProvider` itself. I can't see creating a throw-only method just so you can use its _name_ somewhere else (`nameof(ExtensionLoaded)`).

If you are concerned about hard-coding a name into TelemetryProvider, and then somebody accidentally creates a "real" event with the same name in this class, then just defined
```
internal const string ExtensionLoaded = nameof(ExtensionLoaded);
```

---
In reply to: [505753641](https://github.com/microsoft/sarif-visualstudio-extension/pull/226#discussion_r505753641) [](ancestors = 505753641,505752538)

GabeDeBacker: I'll add the work "don't"

---
In reply to: [505753641](https://github.com/microsoft/sarif-visualstudio-extension/pull/226#discussion_r505753641) [](ancestors = 505753641,505752538)

lgolding: IMO adding the word "don't" to a comment that explains why you should never call a method that always throws won't do. If you need a string, declare a string. You don't need a method.

---
In reply to: [505756813](https://github.com/microsoft/sarif-visualstudio-extension/pull/226#discussion_r505756813) [](ancestors = 505756813,505753641,505752538)

jeffersonking (approved):

roywill (approved):
:shipit:

PR #227, opened 2020-10-17
Migrate to .NET Fx 4.7.2

lgolding:


jeffersonking (approved):

lgolding:
Since @jeffersonking approved this, I had to make some large changes to get everything building again.

The first hint something was wrong was an unresolvable conflict between versions of Newtonsoft.Json. It turned out that some of the VSIX's NuGet dependencies wanted one version, and some of the Interop assembly's dependencies wanted another. It seemed to be related to the fact that the VSIX was a .NET Framework project and the Interop assembly was a .NET Core project. Since we've previously discussed that there was no good reason for the Interop to be .NET Core (other than that it _could_ be) because it was always consumed from a VSIX project, I took the opportunity to replace Interop's .NET Core project file with an old-style class library project file.

This in turn allowed us to greatly simplify the build scripts, which previously had to account for a mixture of .NET Framework and .NET Core projects.

Finally, because you can't just "dotnet pack" an old-style project, I had to craft a .nuspec for the Interop project.

And, since I wanted to use `<license>` rather than `<licenseUrl>` in the .nuspec, I tool the opportunity to upgrade the copy of NuGet.exe in this enlistment.

jeffersonking:
@lgolding Re-approving these new changes.

lgolding:
Thanks!

---
In reply to: [711080417](https://github.com/microsoft/sarif-visualstudio-extension/pull/227#issuecomment-711080417) [](ancestors = 711080417)

PR #234, opened 2020-10-29
Demonstrate proof of concept: push test data to SARIF Viewer

lgolding:
We read a SARIF log from an embedded resource string, write it to a temp file, and then call into the SARIF Viewer interop assembly to cause the results to appear in the viewer.

To repro:

1. Ensure you have the SARIF Viewer extension installed (we don't yet declare an explicit dependency on it).
1. Pull and build the branch `users/lgolding/sarifer-poc`.
1. Set `Sarif.Sarifer` as the startup project.
1. `CTRL-F5`
1. Select menu item **Tools**, **Generate SARIF Test Data**:

image

A single SARIF result appears in the Error List Window:

image


src/Sarif.Viewer.VisualStudio.Interop/SarifViewerInterop.cs
```diff
@@ -111,7 +111,7 @@ public Task<bool> OpenSarifLogAsync(string path)
 
             return this.CallServiceApiAsync(ViewerLoadServiceInterfaceName, (service) =>
             {
-                service.LoadSarifLogs(path);
```
lgolding, line 114: >LoadSarifLogs [](start = 24, length = 13)

This was a misspelling, missed because `service` is `dynamic` and this method was (apparently) not tested. There is no overload of `ILoadSarifLogService.LoadSarifLogs` that takes a single `string` argument. This call caused a runtime error.

src/Sarif.Sarifer/TestData/proof-of-concept.sarif
```diff
@@ -0,0 +1,43 @@
+{
+  "$schema": "https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.5.json",
+  "version": "2.1.0",
+  "runs": [
+    {
+      "tool": {
+        "driver": {
+          "name": "SpellChecker",
```
michaelcfanning, line 8: >SpellChecker [](start = 19, length = 12)

Can you please embed Example.cs into this log file as text? Using the 'artifacts' table. Add an artifact index as well, of course.

lgolding: Sure.

---
In reply to: [514515148](https://github.com/microsoft/sarif-visualstudio-extension/pull/234#discussion_r514515148) [](ancestors = 514515148)

src/Sarif.Sarifer/GenerateTestDataCommand.cs
```diff
@@ -0,0 +1,64 @@
...
+            {
+                s_viewerInterop.LoadViewerExtension();
+            }
+
+            await s_viewerInterop.OpenSarifLogAsync(testDataFilePath);
+        }
+
+        private static async Task<string> CreateTestDataFilePathAsync()
+        {
+            using (Stream testDataResourceStream = Assembly.GetExecutingAssembly().GetManifestResourceStream("TestData.proof-of-concept.sarif"))
```
michaelcfanning, line 52: >proof-of-concept [](start = 119, length = 16)

Nit, can you rename this? Generally, we don't have hyphens in our TestData persisted file names, can we just pascal case it?

lgolding: Will do.

---
In reply to: [514516094](https://github.com/microsoft/sarif-visualstudio-extension/pull/234#discussion_r514516094) [](ancestors = 514516094)

michaelcfanning (approved):
:shipit:

src/Sarif.Viewer.VisualStudio.Interop/SarifViewerInterop.cs
```diff
@@ -111,7 +111,7 @@ public Task<bool> OpenSarifLogAsync(string path)
...
+                service.LoadSarifLog(path);
```
michaelcfanning, line 114: >LoadSarifLog [](start = 24, length = 12)

we're going to shortly need an API that accepts an in-memory SARIF log (because we won't want to require that a background checker persist logs to disk, for perf reasons). Does this API exist currently?

src/Sarif.Viewer.VisualStudio.Interop/SarifViewerInterop.cs
michaelcfanning: Not entirely clear here, is there a test gap? If so, where is it?

---
In reply to: [514514633](https://github.com/microsoft/sarif-visualstudio-extension/pull/234#discussion_r514514633) [](ancestors = 514514633)

src/Sarif.Viewer.VisualStudio.Interop/SarifViewerInterop.cs
lgolding: It doesn't, but now's a fine time to add it (next PR).

---
In reply to: [514521066](https://github.com/microsoft/sarif-visualstudio-extension/pull/234#discussion_r514521066) [](ancestors = 514521066)

src/Sarif.Viewer.VisualStudio.Interop/SarifViewerInterop.cs
lgolding: A test gap. There are no tests at all of the interop assembly (they'd have to be integration tests because it expects the viewer at the other end).

---
In reply to: [514524509](https://github.com/microsoft/sarif-visualstudio-extension/pull/234#discussion_r514524509) [](ancestors = 514524509,514514633)

src/Sarif.Sarifer/GenerateTestDataCommand.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.Threading.Tasks;
+using Microsoft.Sarif.Viewer.Interop;
+using Microsoft.VisualStudio.Shell;
+using Microsoft.VisualStudio.Shell.Interop;
+
+namespace Microsoft.CodeAnalysis.Sarif.Sarifer
+{
+    internal class GenerateTestDataCommand : MenuCommand
+    {
+        private static SarifViewerInterop s_viewerInterop;
```
GabeDeBacker, line 17: >static [](start = 16, length = 6)

Does this really need to be static?

src/Sarif.Sarifer/GenerateTestDataCommand.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        private static async Task<string> CreateTestDataFileAsync()
+        {
+            using (Stream testDataResourceStream = Assembly.GetExecutingAssembly().GetManifestResourceStream("TestData.ProofOfConcept.sarif"))
+            using (TextReader reader = new StreamReader(testDataResourceStream))
+            {
+                string testDataFileContents = await reader.ReadToEndAsync();
+
+                string testDataFilePath = Path.GetTempFileName();
+                File.WriteAllText(testDataFilePath, testDataFileContents); // Pity there's no async version.
```
GabeDeBacker, line 58: >File.WriteAllText(testDataFilePath, testDataFileContents); // Pity there's no async version. [](start = 16, length = 92)


You can use StreamWriter to do async writes (if you are so inclined).

That would move this off the UI thread at least.

src/Sarif.Sarifer/GenerateTestDataCommand.cs
lgolding: I didn't know how to get an instance method passed in as the "event handler" in the call to the base class ctor on Line 21. Would love to change that.

---
In reply to: [514612188](https://github.com/microsoft/sarif-visualstudio-extension/pull/234#discussion_r514612188) [](ancestors = 514612188)

src/Sarif.Sarifer/GenerateTestDataCommand.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            await s_viewerInterop.OpenSarifLogAsync(testDataFilePath);
+        }
+
+        private static async Task<string> CreateTestDataFileAsync()
+        {
+            using (Stream testDataResourceStream = Assembly.GetExecutingAssembly().GetManifestResourceStream("TestData.ProofOfConcept.sarif"))
+            using (TextReader reader = new StreamReader(testDataResourceStream))
+            {
+                string testDataFileContents = await reader.ReadToEndAsync();
```
GabeDeBacker, line 55: Here is a case where using ConfigureAwait(continueOnCapturedContext: false) would be a good thing. You don't need to continue the rest of this method on the UI thread. (FxCop analyzers will tell you this :))

src/Sarif.Sarifer/GenerateTestDataCommand.cs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        /// <summary>
+        /// Event handler called when the user selects the Generate SARIF Test Data command.
+        /// </summary>
+        /// <remarks>
+        /// Since this is a menu item callback, it must return void.
+        /// </remarks>
+#pragma warning disable VSTHRD100 // Avoid async void methods
+        private static async void MenuCommandCallback(object caller, EventArgs args)
```
GabeDeBacker, line 34: >static async [](start = 16, length = 12)

This doesn't need to be static or async :)

Use VS's FileAndForget method for calling OpenSarifLogsasync().FileAndForget() for starting the task.

GabeDeBacker:
>   <ItemGroup>

New project, suggest adding FxCop and StyleCop for it

---
Refers to: src/Sarif.Sarifer/Sarif.Sarifer.csproj:26 in aa52639. [](commit_id = aa52639e7d010a1c1e9804c69a9b6fa8c96219db, deletion_comment = False)

src/Sarif.Sarifer/GenerateTestDataCommand.cs
lgolding: Fixed.

---
In reply to: [514613468](https://github.com/microsoft/sarif-visualstudio-extension/pull/234#discussion_r514613468) [](ancestors = 514613468)

src/Sarif.Sarifer/GenerateTestDataCommand.cs
lgolding: Fixed, using your suggestion to aggregate rather than inherit from `MenuCommand`.

---
In reply to: [514613555](https://github.com/microsoft/sarif-visualstudio-extension/pull/234#discussion_r514613555) [](ancestors = 514613555,514612188)

src/Sarif.Sarifer/GenerateTestDataCommand.cs
lgolding: Fixed. The reason it wasn't compiling was I needed a reference to separate NuGet package `Microsoft.VisualStudio.Shell.15.0`.

---
In reply to: [514614472](https://github.com/microsoft/sarif-visualstudio-extension/pull/234#discussion_r514614472) [](ancestors = 514614472)

lgolding:
>   <ItemGroup>

Added FxCop analyzers. Will do StyleCop in a separate PR.

---
In reply to: [719070623](https://github.com/microsoft/sarif-visualstudio-extension/pull/234#issuecomment-719070623) [](ancestors = 719070623)

---
Refers to: src/Sarif.Sarifer/Sarif.Sarifer.csproj:26 in aa52639. [](commit_id = aa52639e7d010a1c1e9804c69a9b6fa8c96219db, deletion_comment = False)

Issue #237, opened 2020-10-30
Apply uniform style guidelines to repository


lgolding:
... while it's still small. Let's add it to the interop project while we're at it.

mpande98:
Sarifer = another VSIX which gives menu items

lgolding:
Let's bite the bullet and do it for all the projects.

michaelcfanning:
@eddynaka, can you work your .editorconfig magic on this repo?

eddynaka:
@lgolding @michaelcfanning , I just created a PR with this.

Should we use the same rules that we have in sarif-sdk? For example, should we separate in groups (that's the only rule that I didn't enable here yet).

lgolding:
IMO yes, we should use the same rules as the SARIF SDK.

PR #238, opened 2020-10-30
Address Gabe's feedback from previous PR

lgolding:
- Use `StreamWriter.WriteAsync` instead of `File.WriteAllText`.
- Aggregate, don't inherit from, `MenuCommand`. This avoids the need for static methods and data.
- `MenuItemCallback` doesn't need to be `async`: use `FileAndForget` instead of `await`.
- Add FxCop analyzers to Sarifer and Interop projects.

Gabe had also asked to add StyleCop analyzers to the new project while it was still small. I'll do that in a separate PR. I filed #237.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,4 @@
-image: Visual Studio 2017
+image: Visual Studio 2019
 
 init:
   - git config --global core.autocrlf true
--- a/src/Sarif.Sarifer/GenerateTestDataCommand.cs
+++ b/src/Sarif.Sarifer/GenerateTestDataCommand.cs
@@ -55,7 +55,10 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer
                 string testDataFileContents = await reader.ReadToEndAsync();
 
                 string testDataFilePath = Path.GetTempFileName();
-                File.WriteAllText(testDataFilePath, testDataFileContents); // Pity there's no async version.
+                using (var writer = new StreamWriter(testDataFilePath))
+                {
+                    await writer.WriteAsync(testDataFileContents);
+                }
 
                 return testDataFilePath;
             }
--- a/src/Sarif.Sarifer/GenerateTestDataCommand.cs
+++ b/src/Sarif.Sarifer/GenerateTestDataCommand.cs
@@ -13,8 +13,10 @@ using Task = System.Threading.Tasks.Task;
 
 namespace Microsoft.CodeAnalysis.Sarif.Sarifer
 {
-    internal class GenerateTestDataCommand
+    internal class GenerateTestDataCommand : SariferCommandBase
     {
+        private const string SendDataToViewerFailureEventName = "SendDataToViewer/Failure";
+
         private readonly SarifViewerInterop viewerInterop;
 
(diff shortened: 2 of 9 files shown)


src/Sarif.Sarifer/GenerateTestDataCommand.cs
```diff
@@ -9,53 +9,61 @@
...
+        private void MenuCommandCallback(object caller, EventArgs args)
         {
-            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+            this.SendDataToViewerAsync().FileAndForget(GetFileAndForgetEventName(SendDataToViewerFailureEventName));
+        }
 
-            string testDataFilePath = await CreateTestDataFileAsync();
+        private async Task SendDataToViewerAsync()
+        {
+            string testDataFilePath = await CreateTestDataFileAsync().ConfigureAwait(continueOnCapturedContext: true);
```
GabeDeBacker, line 43: >true [](start = 112, length = 4)

This is interesting. One would thing that the SARIF viewer code would switch to the UI thread if it needed to :)

Can't think of a reason why either of these need to be true.. (There may be one, I just don't know it)

lgolding: I'm convinced. Changed.

---
In reply to: [514645936](https://github.com/microsoft/sarif-visualstudio-extension/pull/238#discussion_r514645936) [](ancestors = 514645936)

src/Sarif.Sarifer/SariferPackage.cs
```diff
@@ -53,13 +53,10 @@ protected override async Task InitializeAsync(CancellationToken cancellationToke
 
             // The OleCommandService object provided by the MPF is responsible for managing the set
             // of commands implemented by the package.
-            if (await GetServiceAsync(typeof(IMenuCommandService)) is OleMenuCommandService mcs &&
-                await GetServiceAsync(typeof(SVsShell)) is IVsShell vsShell)
+            if (await GetServiceAsync(typeof(IMenuCommandService)).ConfigureAwait(continueOnCapturedContext: true) is OleMenuCommandService mcs &&
+                await GetServiceAsync(typeof(SVsShell)).ConfigureAwait(continueOnCapturedContext: true) is IVsShell vsShell)
             {
```
GabeDeBacker, line 58: nit: This "could" be done in a static method of GenerateTestDataCommand.

Something like inernal static void GenerateTestDataCommand.InitAsync(AsyncPcakcage pacakge)

GabeDeBacker (approved):
:shipit:

src/Sarif.Sarifer/SariferPackage.cs
lgolding: Personal preference to inject the dependencies, making the dependencies explicit.

---
In reply to: [514646623](https://github.com/microsoft/sarif-visualstudio-extension/pull/238#discussion_r514646623) [](ancestors = 514646623)

appveyor.yml
```diff
@@ -1,4 +1,4 @@
-image: Visual Studio 2017
+image: Visual Studio 2019
```
lgolding, line 1: >2019 [](start = 21, length = 4)

Because otherwise the FxCop rules assembly doesn't load (can't be found)

PR #259, opened 2020-11-06
Fix #236: Provide an interop API to accept a SARIF log from a stream

### LINKED ISSUE #236: Provide an interop API to accept a SARIF log from a stream
That way background checkers don't have to write their results to disk first.

@michaelcfanning @GabeDeBacker

### LINKED ISSUE #254: Refactors in preparation for stream-driven SARIF transfer API
The existing code tightly intertwines file handling logic with core display logic. This makes it difficult to provide an in-memory file transfer interop API (#251). We propose to perform the following refactors to (1) make it practical to implement the API, and (2) improve unit testability:

**TODO** This is a partial list. I'll be adding more details to it later, as I continue review my conversation with @GabeDeBacker on this topic.

- [ ] Make `public` the currently `internal` overload of `ErrorListService.ProcessLogFileAsync` that takes a `SarifLog` argument. Note that it still takes the file path as an argument.
- [ ] In `CodeAnalysisResultsManager.PromptForResolvedPath`, in the initialization of `OpenFileDialog.InitialDirectory`, allow `errorListItem.LogFilePath` to be `null`.

@michaelcfanning FYI

lgolding:
This turns out to require only a small change to `ErrorListService`. We introduce a "shim" API `ErrorListService.ProcessSarifLogAsync(Stream)`, which deserializes the stream into a `SarifLog` and then calls an existing private API `ErrorListService.ProcessSarifLogAsync(SarifLog)`.

The fact that there is no file name available turns out not to matter:
- The file name never appears in the UI.
- At the moment, there is no gesture to re-read the file, or to re-persist it, so the lack of a backing file is moot.
- The extension does use the log file path, when prompting the user to resolve a source file path, as the "initial directory" in the File Open dialog. But if the path is null, it just starts in the current working directory.

The only other thing the extension uses the log file path for is as a key into a dictionary that associates sets of results with log files. This allows the extension to remove those results from the Error List when the user closes a log file. Without a file name, the results from a streamed in file can't be collectively removed from the error list.

But the scenario for using the stream-based APIs is background analysis, and there's no gesture today to remove (for example) all IDE warnings from the error list, or all warnings from the StyleCop or FxCop analyzer. So our proposed SARIF-based background analysis is no worse off in this regard than existing background analyses.

In fact, the stream-based API is no worse off than the file-based API in this regard. When another extension sends a SARIF file to the viewer through the file-based API, the viewer does _not_ display the file -- so again, there is no "close the file and remove its results from the Error List" gesture.

Because this turns out to "just work", we don't need to refactor, so we will close #254, "Refactors in preparation for stream-driven SARIF transfer API", as `resolved-wont-fix`.

### DIFF
--- a/src/Sarif.Sarifer/GenerateTestDataCommand.cs
+++ b/src/Sarif.Sarifer/GenerateTestDataCommand.cs
@@ -42,7 +42,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer
 
         private async Task SendDataToViewerAsync()
         {
-            string testDataFilePath = await CreateTestDataFileAsync().ConfigureAwait(continueOnCapturedContext: true);
+            Stream testDataStream = GetTestDataStream();
 
             // TODO: Why does this never return true?
             if (!viewerInterop.IsViewerExtensionLoaded)
@@ -50,25 +50,10 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer
                 this.viewerInterop.LoadViewerExtension();
             }
 
-            await this.viewerInterop.OpenSarifLogAsync(testDataFilePath).ConfigureAwait(continueOnCapturedContext: false);
+            await this.viewerInterop.OpenSarifLogAsync(testDataStream).ConfigureAwait(continueOnCapturedContext: false);
         }
 
-        private static async Task<string> CreateTestDataFileAsync()
-        {
-            using (Stream testDataResourceStream = Assembly.GetExecutingAssembly().GetManifestResourceStream("TestData.ProofOfConcept.sarif"))
-            using (TextReader reader = new StreamReader(testDataResourceStream))
-            {
-                // No need to continue on the UI thread because we're just doing file I/O.
-                string testDataFileContents = await reader.ReadToEndAsync().ConfigureAwait(continueOnCapturedContext: false);
-
-                string testDataFilePath = Path.GetTempFileName();
-                using (var writer = new StreamWriter(testDataFilePath))
-                {
-                    await writer.WriteAsync(testDataFileContents).ConfigureAwait(continueOnCapturedContext: false);
-                }
-
-                return testDataFilePath;
-            }
-        }
+        private static Stream GetTestDataStream() =>
+            Assembly.GetExecutingAssembly().GetManifestResourceStream("TestData.ProofOfConcept.sarif");
     }
 }
--- a/src/Sarif.Sarifer/GenerateTestDataCommand.cs
+++ b/src/Sarif.Sarifer/GenerateTestDataCommand.cs
@@ -17,6 +17,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer
 {
     internal class GenerateTestDataCommand
     {
+        private const string ProofOfConceptResourceName = "TestData.ProofOfConcept.sarif";
         private const string SendDataToViewerFailureEventName = "SendDataToViewer/Failure";
 
         private readonly SarifViewerInterop viewerInterop;
@@ -37,13 +38,13 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer
         /// </summary>
         private void MenuCommandCallback(object caller, EventArgs args)
         {
-            this.SendDataToViewerAsync().FileAndForget(FileAndForget.EventName(SendDataToViewerFailureEventName));
+            Stream testDataStream = Assembly.GetExecutingAssembly().GetManifestResourceStream(ProofOfConceptResourceName);
+
+            this.SendDataToViewerAsync(testDataStream).FileAndForget(FileAndForget.EventName(SendDataToViewerFailureEventName));
         }
 
-        private async Task SendDataToViewerAsync()
+        private async Task SendDataToViewerAsync(Stream testDataStream)
         {
-            Stream testDataStream = GetTestDataStream();
-
             // TODO: Why does this never return true?
             if (!viewerInterop.IsViewerExtensionLoaded)
             {
@@ -52,8 +53,5 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer
 
             await this.viewerInterop.OpenSarifLogAsync(testDataStream).ConfigureAwait(continueOnCapturedContext: false);
         }
-
-        private static Stream GetTestDataStream() =>
-            Assembly.GetExecutingAssembly().GetManifestResourceStream("TestData.ProofOfConcept.sarif");
     }
 }
--- a/src/Sarif.Sarifer/GenerateTestDataCommand.cs
+++ b/src/Sarif.Sarifer/GenerateTestDataCommand.cs
@@ -5,13 +5,10 @@ using System;
 using System.ComponentModel.Design;
 using System.IO;
 using System.Reflection;
-using System.Threading.Tasks;
 
 using Microsoft.Sarif.Viewer.Interop;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
-
-using Task = System.Threading.Tasks.Task;
 
 namespace Microsoft.CodeAnalysis.Sarif.Sarifer
 {
@@ -26,7 +23,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer
         {
             this.viewerInterop = new SarifViewerInterop(vsShell);
 
-            MenuCommand menuCommand = new MenuCommand(
+            var menuCommand = new MenuCommand(
                 new EventHandler(this.MenuCommandCallback),
                 new CommandID(Guids.SariferCommandSet, SariferPackageCommandIds.GenerateTestData));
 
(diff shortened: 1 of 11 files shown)


src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -334,7 +336,15 @@ private static async System.Threading.Tasks.Task SaveLogFileAsync(string filePat
             }
         }
 
-        internal static async System.Threading.Tasks.Task ProcessSarifLogAsync(SarifLog sarifLog, string logFilePath, bool showMessageOnNoResults, bool cleanErrors, bool openInEditor)
+        internal static async Task ProcessSarifLogAsync(Stream stream, bool showMessageOnNoResults, bool cleanErrors, bool openInEditor)
+        {
+            SarifLog sarifLog = SarifLog.Load(stream);
+            string logFilePath = Path.GetTempFileName();
```
GabeDeBacker, line 342: Interesting... I'd like to understand the ramifications of this.

GabeDeBacker: vs allowing it to be null

---
In reply to: [519462499](https://github.com/microsoft/sarif-visualstudio-extension/pull/259#discussion_r519462499) [](ancestors = 519462499)

src/Sarif.Viewer.VisualStudio/OpenLogFileCommands.cs
```diff
@@ -271,7 +271,7 @@ private async System.Threading.Tasks.Task MenuItemCallbackAsync(object sender, E
             }
         }
 
-        string ConvertSarifProtocol(string inputUrl)
+        private string ConvertSarifProtocol(string inputUrl)
```
GabeDeBacker, line 274: static?

src/Sarif.Viewer.VisualStudio/OpenLogFileCommands.cs
```diff
@@ -298,7 +298,7 @@ string ConvertSarifProtocol(string inputUrl)
             return newUrl;
         }
 
-        bool TryDownloadFile(string inputUrl, out string downloadedFilePath)
+        private bool TryDownloadFile(string inputUrl, out string downloadedFilePath)
```
GabeDeBacker, line 301: static?

src/Sarif.Viewer.VisualStudio/OpenLogFileCommands.cs
GabeDeBacker: I would argue that both of these methods (this one and the one below) should be in some utilities class. Am I mistaken?

---
In reply to: [519462594](https://github.com/microsoft/sarif-visualstudio-extension/pull/259#discussion_r519462594) [](ancestors = 519462594)

src/Sarif.Sarifer/GenerateTestDataCommand.cs
```diff
@@ -37,38 +35,15 @@ public GenerateTestDataCommand(IVsShell vsShell, IMenuCommandService menuCommand
...
-
-                string testDataFilePath = Path.GetTempFileName();
-                using (var writer = new StreamWriter(testDataFilePath))
-                {
-                    await writer.WriteAsync(testDataFileContents).ConfigureAwait(continueOnCapturedContext: false);
-                }
-
-                return testDataFilePath;
-            }
+            this.viewerInterop.OpenSarifLogAsync(testDataStream).FileAndForget(FileAndForget.EventName(SendDataToViewerFailureEventName));
```
GabeDeBacker, line 46: >testDataStream [](start = 49, length = 14)

Nice!

GabeDeBacker (approved):
:shipit:

src/Sarif.Viewer.VisualStudio/OpenLogFileCommands.cs
lgolding: Yes, both static.

"Mistaken" is putting it a bit strongly. But both of these are strongly related to opening files, and there isn't another client at the moment. I'd leave them here.

---
In reply to: [519462594](https://github.com/microsoft/sarif-visualstudio-extension/pull/259#discussion_r519462594) [](ancestors = 519462594)

src/Sarif.Viewer.VisualStudio/OpenLogFileCommands.cs
lgolding: yes

---
In reply to: [519462625](https://github.com/microsoft/sarif-visualstudio-extension/pull/259#discussion_r519462625) [](ancestors = 519462625)

src/Sarif.Sarifer/GenerateTestDataCommand.cs
lgolding: Yeah, right? :-)

---
In reply to: [519462799](https://github.com/microsoft/sarif-visualstudio-extension/pull/259#discussion_r519462799) [](ancestors = 519462799)

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
lgolding: Hmm. In the PR description I mentioned a choice between a "real" file and a random (GUID) string, but I didn't consider leaving it blank. Let's try that...

> Since the file isn't really used, I considered just using GUID string as the file path, without actually creating the empty file. I chose to create the file since if in future, we introduce gestures that actually make use of the file (for example, writing a modified set of results back to it), it will be good to have it around. Otherwise, we'd be writing GUID-named files to whatever the current directory is.

---
In reply to: [519462520](https://github.com/microsoft/sarif-visualstudio-extension/pull/259#discussion_r519462520) [](ancestors = 519462520,519462499)

lgolding: You are right. It works just fine without a log file path name.

---
In reply to: [519464163](https://github.com/microsoft/sarif-visualstudio-extension/pull/259#discussion_r519464163) [](ancestors = 519464163,519462520,519462499)

src/Sarif.Viewer.VisualStudio/Services/ILoadSarifLogService2.cs
```diff
@@ -5,6 +5,9 @@
 
 namespace Microsoft.Sarif.Viewer.Services
 {
+    /// <summary>
+    /// Interface for loading SARIF logs from multiple files into the viewer.
+    /// </summary>
     internal interface ILoadSarifLogService2
```
michaelcfanning, line 11: >ILoadSarifLogService2 [](start = 23, length = 21)

I don't understand the strategy of evolving these interfaces. For one thing, we own most of this architecture outright, why are we being so precious about maintenance? I'm not aware of a significant customer of these interfaces. Second, in the COM world a '2' interface would typically inherit from and extend the original version.

Can we simply converge these into a single interface? Yes, I know this is a very COM unfriendly thing to do, back in 1996 it certainly wouldn't fly. Today (in this specific context), I'm not sure I care.

lgolding: Gabe and I discussed both points, but happy to reopen the discussion.

1. Re: numbered interfaces extending earlier interfaces: VS is a mixed bag here. For example, `IVsWindowFrameEvents2` extends `IVsWindowFrameEvents`, but `IVsWindowFrameNotify2` does not extend `IVsWindowFrameNotify`. I have no objection to adding the extensions, subject to point #2.

2. Re: COM unfriendliness, two considerations:
- As old as COM is, VS still uses it under the covers.
- Regardless of COM, adding a method to a "published" (more on that below) interface is a breaking change.

Then the question becomes, are there any customers of these interfaces. The only customer I'm aware of is WAVE -- but! they access the interfaces through the Viewer Interop API -- which uses `dynamic`, and so is _not_ a compile time customer of the interface.

It comes down to whether we think anyone has every taken a build-time dependency on the Viewer's DLL (which is where those interfaces are defined), and then defined a type that implements `ILoadSarifLogService`.



---
In reply to: [520680157](https://github.com/microsoft/sarif-visualstudio-extension/pull/259#discussion_r520680157) [](ancestors = 520680157)

michaelcfanning: The point is not VS general use of COM, the point is whether its use of COM requires us to maintain the level of care around COM interfaces that are required in some scenarios.

My guidance is to simply converge everything into ILoadSarifLogService. I'm not aware of any utilization we would break. If we do break someone, I'd be glad to help fix whatever is required.

---
In reply to: [520702295](https://github.com/microsoft/sarif-visualstudio-extension/pull/259#discussion_r520702295) [](ancestors = 520702295,520680157)

lgolding: Ok, I will merge the interfaces as part of my current PR (which BTW is around displaying the info bar if we receive invalid JSON).

---
In reply to: [520725161](https://github.com/microsoft/sarif-visualstudio-extension/pull/259#discussion_r520725161) [](ancestors = 520725161,520702295,520680157)

michaelcfanning: Please fix the interfaces here, so this change is coherent. i.e., i'd prefer that we not introduce a 3rd interface here then remove it in an arbitrary future PR. The adjustment is small. Let me know when it's in and I'll merge this otherwise approved PR.

---
In reply to: [520726644](https://github.com/microsoft/sarif-visualstudio-extension/pull/259#discussion_r520726644) [](ancestors = 520726644,520725161,520702295,520680157)

lgolding: As discussed offline: This change is already merged, and we agree it's not worth the effort to have a separate "merge interfaces" PR. We will add this to the next PR.

---
In reply to: [520729099](https://github.com/microsoft/sarif-visualstudio-extension/pull/259#discussion_r520729099) [](ancestors = 520729099,520726644,520725161,520702295,520680157)

PR #260, opened 2020-11-07
Introduce framework for background analysis; implement POC analyzer

lgolding:
**PULLING THIS BACK TO DRAFT WHILE DISCUSSING DESIGN WITH GABE**

NOTE: I'm not going to merge this until the corresponding UI tests are signed off. At the moment, I'm waiting for sign-off on the UI tests of the previous feature (the stream-based interop API, on which _this_ feature depends). Once that's in, I'll add the UI tests for this feature.


src/Sarif.Sarifer/FileAndForget.cs
```diff
@@ -1,12 +0,0 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
-
-namespace Microsoft.CodeAnalysis.Sarif.Sarifer
-{
-    internal static class FileAndForget
```
lgolding, line 6: >FileAndForget [](start = 26, length = 13)

Replaced with `FileAndForgetEventNames`.

lgolding:
> Copyright (c) Microsoft. All rights reserved. 

Formatter wanted a white space change.

---
Refers to: src/Sarif.Sarifer/source.extension.vsixmanifest:3 in 11f7052. [](commit_id = 11f7052a59bf8ff0e6ffba709e1fb2a3222b6656, deletion_comment = False)

src/Sarif.Sarifer/GenerateTestDataCommand.cs
```diff
@@ -37,7 +35,7 @@ public GenerateTestDataCommand(IVsShell vsShell, IMenuCommandService menuCommand
         /// </summary>
         private void MenuCommandCallback(object caller, EventArgs args)
         {
-            this.SendDataToViewerAsync().FileAndForget(FileAndForget.EventName(SendDataToViewerFailureEventName));
+            this.SendDataToViewerAsync().FileAndForget(FileAndForgetEventName.SendDataToViewerFailure);
```
lgolding, line 38: >e [](start = 100, length = 1)

Smoother to read than the old way.

src/Sarif.Viewer.VisualStudio/Services/LoadSarifLogService.cs
```diff
@@ -31,14 +31,14 @@ public void LoadSarifLog(string path, bool promptOnSchemaUpgrade = true)
...
-        public void LoadSarifLog(string path)
+        public void LoadSarifLog(string path, bool promptOnLogConversions, bool cleanErrors, bool openInEditor)
         {
             if (string.IsNullOrWhiteSpace(path))
             {
                 return;
             }
 
-            ErrorListService.ProcessLogFile(path, ToolFormat.None, promptOnLogConversions: true, cleanErrors: true, openInEditor: false);
+            ErrorListService.ProcessLogFile(path, ToolFormat.None, promptOnLogConversions, cleanErrors, openInEditor);
```
GabeDeBacker, line 41: >ProcessLogFile [](start = 29, length = 14)

I am 100% on board with not having default parameters for ProcessLogFile :)

src/Sarif.Sarifer/BackgroundAnalysisService.cs
```diff
@@ -0,0 +1,37 @@
...
+#pragma warning disable IDE0044 // Assigned by MEF
+        [ImportMany]
+        private IEnumerable<IBackgroundAnalyzer> analyzers;
+#pragma warning restore IDE0044
+#pragma warning restore CS0649
+
+        /// <inheritdoc/>
+        public void StartAnalysis(string path, string text)
+        {
+            if (this.analyzers.Any() == true)
```
GabeDeBacker, line 28: >== true [](start = 37, length = 7)

Not needed

GabeDeBacker: Can the import be null? (Honestly don't know, what if there are none?)

---
In reply to: [519917088](https://github.com/microsoft/sarif-visualstudio-extension/pull/260#discussion_r519917088) [](ancestors = 519917088)

src/Sarif.Sarifer/BackgroundAnalysisService.cs
```diff
@@ -0,0 +1,37 @@
...
+            {
+                foreach (IBackgroundAnalyzer analyzer in this.analyzers)
```
GabeDeBacker, line 30: >this.analyzers [](start = 57, length = 14)

If this.analyzers cannot be null, then the any isn't needed at all right?

src/Sarif.Sarifer/BackgroundAnalysisService.cs
```diff
@@ -0,0 +1,37 @@
...
+                {
+                    analyzer.StartAnalysis(path, text);
```
GabeDeBacker, line 32: >StartAnalysis [](start = 29, length = 13)

I would think that would be StartAnalysisAsync and that this class would keep an array of running analysis tasks. (should likely also integrate with VS task center with some sort of progress so the user knows something is happening in the background)

src/Sarif.Sarifer/IBackgroundAnalyzer.cs
```diff
@@ -0,0 +1,19 @@
...
+    /// </summary>
+    public interface IBackgroundAnalyzer
+    {
+        /// <summary>
+        /// Performs a single analysis on the specified text.
+        /// </summary>
+        /// <param name="text">
+        /// The text to analyze
+        /// </param>
+        void StartAnalysis(string path, string text);
```
GabeDeBacker, line 17: >StartAnalysis [](start = 13, length = 13)

Perhaps Task StartAnalysisAsync(string path, string text, CancellationToken cancellationToken)?

GabeDeBacker, line 17: >StartAnalysis [](start = 13, length = 13)

We should probably be consistent and have a stream version as well

src/Sarif.Sarifer/BackgroundAnalysisTextViewCreationListener.cs
```diff
@@ -0,0 +1,76 @@
...
+                return null;
+            }
+
+            if (!(vsTextLines is IPersistFileFormat persistFile))
+            {
+                return null;
+            }
+
+            return persistFile.GetCurFile(out string path, out _) == VSConstants.S_OK
+                ? path
```
GabeDeBacker, line 72: >path [](start = 18, length = 4)

Remember that this can return S_OK with null (so make sure the calling code handles this). Perhaps this should be TryGetPathFromTextView?

src/Sarif.Sarifer/BackgroundAnalysisTextViewCreationListener.cs
```diff
@@ -0,0 +1,76 @@
...
+#pragma warning restore CS0649
+
+        /// <inheritdoc/>
+        public void TextViewCreated(ITextView textView)
+        {
+            ThreadHelper.ThrowIfNotOnUIThread();
+
+            textView = textView ?? throw new ArgumentNullException(nameof(textView));
+
+            string text = textView.TextBuffer.CurrentSnapshot.GetText();
```
GabeDeBacker, line 44: >textView [](start = 26, length = 8)

Something to consider, when a text view is closed, could the buffer text become invalid? (not sure exactly how this should be handled)

src/Sarif.Sarifer/BackgroundAnalysisTextViewCreationListener.cs
```diff
@@ -0,0 +1,76 @@
...
+            string path = GetPathFromTextView(textView);
+
+            this.backgroundAnalysisService.Value.StartAnalysis(path, text);
```
GabeDeBacker, line 47: >StartAnalysis [](start = 49, length = 13)

So when the text view is closed, do we want to stop analysis? (Should this be async and take a cancellation token?)

src/Sarif.Sarifer/BackgroundAnalyzerBase.cs
```diff
@@ -0,0 +1,86 @@
...
+        /// </returns>
+        protected abstract void CreateSarifLog(string path, string text, TextWriter writer);
+
+        private async Task AnalyzeAsync(string path, string text)
+        {
+            using (Stream stream = new MemoryStream())
+            using (TextWriter writer = new StreamWriter(stream, Encoding.UTF8))
+            {
+                CreateSarifLog(path, text, writer);
+                await writer.FlushAsync().ConfigureAwait(continueOnCapturedContext: true);
```
GabeDeBacker, line 76: >true [](start = 84, length = 4)

Think you mean "false" here and below since you say this can run on a background thread.

src/Sarif.Sarifer/ProofOfConceptBackgroundAnalyzer.cs
```diff
@@ -0,0 +1,116 @@
...
+                    ["default"] = new MultiformatMessageString
+                    {
+                        Text = "Public class should be internal."
+                    }
+                }
+            };
+
+            int targetStringIndex = 0;
+            while (targetStringIndex < text.Length)
+            {
```
GabeDeBacker, line 71: I know that in this same this is a rather trivial analyzers, but a more in-depth analyzer could probably spend some time looking through the text. I would make this Task GenerateReultsAsync() with a cancellation token

src/Sarif.Sarifer/SarifViewerBackgroundAnalysisSink.cs
```diff
@@ -0,0 +1,53 @@
...
+        private SarifViewerInterop sarifViewerInterop;
+
+        /// <inheritdoc/>
+        public async Task ReceiveAsync(Stream logStream)
+        {
+            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+
+            SarifViewerInterop sarifViewerInterop = await GetInteropObjectAsync().ConfigureAwait(continueOnCapturedContext: true);
+
+            await sarifViewerInterop.OpenSarifLogAsync(logStream).ConfigureAwait(continueOnCapturedContext: false);
```
GabeDeBacker, line 34: >logStream [](start = 55, length = 9)

Want to discuss this with you. Mentally I have two pictures in my mind. The first is a completed "stream" versus "streaming results"

lgolding:
Abandoned while we're discussing.

src/Sarif.Sarifer/BackgroundAnalysisService.cs
lgolding: I don't think it can. In my experience, whenever I've made a mistake that failed to satisfy an import, the class was never instantiated.

---
In reply to: [519917408](https://github.com/microsoft/sarif-visualstudio-extension/pull/260#discussion_r519917408) [](ancestors = 519917408,519917088)

src/Sarif.Sarifer/BackgroundAnalysisService.cs
lgolding: Right. The worst it can be is empty, so you're right, the check isn't needed.

---
In reply to: [519917807](https://github.com/microsoft/sarif-visualstudio-extension/pull/260#discussion_r519917807) [](ancestors = 519917807)

src/Sarif.Sarifer/BackgroundAnalysisService.cs
lgolding: Yes, we do need to keep track of what's running and what's been done. I propose to accept the small/easy fixes in this PR, and then we'll use it as the basis for future work. Much of that work will be done by @eddynaka, per @michaelcfanning.

---
In reply to: [519919012](https://github.com/microsoft/sarif-visualstudio-extension/pull/260#discussion_r519919012) [](ancestors = 519919012)

src/Sarif.Sarifer/ProofOfConceptBackgroundAnalyzer.cs
lgolding: Yes, this really is just a placeholder for a more sophisticated analyzer that Eddy will provide. We might very well even include a cancellation token in the public interface.

---
In reply to: [519929252](https://github.com/microsoft/sarif-visualstudio-extension/pull/260#discussion_r519929252) [](ancestors = 519929252)

src/Sarif.Sarifer/BackgroundAnalyzerBase.cs
lgolding: Right.

---
In reply to: [519924943](https://github.com/microsoft/sarif-visualstudio-extension/pull/260#discussion_r519924943) [](ancestors = 519924943)

src/Sarif.Sarifer/BackgroundAnalysisTextViewCreationListener.cs
lgolding: Absolutely. Future work based on this PR as a starting point.

---
In reply to: [519924051](https://github.com/microsoft/sarif-visualstudio-extension/pull/260#discussion_r519924051) [](ancestors = 519924051)

src/Sarif.Sarifer/BackgroundAnalysisTextViewCreationListener.cs
lgolding: The calling code does indeed handle `null` (as it must, since the return code might _not_ be `S_OK`.

---
In reply to: [519921251](https://github.com/microsoft/sarif-visualstudio-extension/pull/260#discussion_r519921251) [](ancestors = 519921251)

src/Sarif.Sarifer/SarifViewerBackgroundAnalysisSink.cs
lgolding: Indeed. We've had this discussion offline and we'll migrate towards the model we discussed, using a tagger to populate the error list, and streaming out individual results (each as a self-contained SARIF file) rather than waiting to analyze the whole buffer.

---
In reply to: [519930411](https://github.com/microsoft/sarif-visualstudio-extension/pull/260#discussion_r519930411) [](ancestors = 519930411)

src/Sarif.Sarifer/IBackgroundAnalyzer.cs
lgolding: I don't think so. Remember that the `path` here is the path to a source file we just opened in the editor (or, per @michaelfanning's ultimate goal, one of the source files in the solution that you just right-clicked and selected "Perform [redacted] analysis"). We won't be streaming content into this API.

---
In reply to: [519920052](https://github.com/microsoft/sarif-visualstudio-extension/pull/260#discussion_r519920052) [](ancestors = 519920052)

PR #266, opened 2020-11-11
Introduce framework for background analysis; introduce POC analyzer

lgolding:
We introduce a `BackgroundAnalysisTextViewCreationListener`. This class implements, and MEF exports, the `ITextViewCreationListener` interface. VS invokes the listener's `TextViewCreated` method when the user opens any file.

The listener invokes the `BackgroundAnalysisService` (which it MEF imports through the `IBackgroundAnalysisService` interface) to start the analysis in the background. The introduction of this and other interfaces proposed here facilitate unit testing.

The `BackgroundAnalysisService` in turn MEF imports any number of objects that implement `IBackgroundAnalyzer` (note the `[ImportMany]`. This means that anyone can plug into the background analysis services by writing an extension containing a class that MEF exports the `IBackgroundAnalyzer` interface.

Our one and only implementation of `IBackgroundAnalyzer` is the `ProofOfConceptBackgroundAnalyzer`, which offers to replace `internal class` with `public class`.

Every analyzer derives from `BackgroundAnalyzerBase`. The base class takes care of stream management and output creation. Derived classes just override its abstract `CreateSarifLog` method, which takes the string of text to analyze, the path to the file, and a `TextWriter` to which the analyzer can write the SARIF log. This works nicely if the analyzer uses the `SarifLogger` (which can be constructed from a `TextWriter`) to create the SARIF. The POC analyzer does just that.

I mentioned "output creation". The base class does a MEF `[ImportMany]` of objects that implement `IBackgroundAnalysisSink`. At the moment, the sole implementer is the `SarifViewerBackgroundAnalysisSink`, which sends the log file to the SARIF viewer through the viewer's interop API (using `LoadSarifLog(Stream)`). But unit tests can replace the sink with (for example) one that writes to a `StringBuilder`, and one could imagine other destinations for the analysis in future (a database, for example, or a telemetry provider).

This implementation has several functional holes and design flaws:

- If you close a file, the error list items resulting from the background analysis of that file remain.
- If you close a file and the background analysis is not complete, the analysis continues. Similarly, if you then quickly reopen the file, multiple copies of the same analysis will run at the same time.
- If you edit the file, the background analysis is not repeated.
- The background operations are not cancelable.
- There's no indication of the background operations in the status center.

Some of these (for example, status center presence) could easily be added to the current design. Most of the other deficiencies stem from the fact that we do not (yet) track of the identity or lifetime of a background analysis and its associated results.

Gabe believes we can evolve the code in this PR to the desired state (which is why I raise it for your inspection).


michaelcfanning:
@lgolding, status on this? can we update this branch from master? is there some baseline work we should merge/build on or do you need to iterate?

let us know if @eddynaka can jump in.

lgolding:
No need for further iteration. I will merge this as soon as you give the go ahead. I've walked @eddynaka through the design and he understands where to plug in his analyzer.

---
In reply to: [728214435](https://github.com/microsoft/sarif-visualstudio-extension/pull/266#issuecomment-728214435) [](ancestors = 728214435)

eddynaka (approved):

lgolding:
This branch is fully up to date with master and ready to merge.

---
In reply to: [728233936](https://github.com/microsoft/sarif-visualstudio-extension/pull/266#issuecomment-728233936) [](ancestors = 728233936,728214435)

src/Sarif.Sarifer/BackgroundAnalysisService.cs
```diff
@@ -0,0 +1,34 @@
...
+#pragma warning restore CS0649
+
+        /// <inheritdoc/>
+        public async Task StartAnalysisAsync(string path, string text)
+        {
+            foreach (IBackgroundAnalyzer analyzer in this.analyzers)
+            {
+                await analyzer.StartAnalysisAsync(path, text).ConfigureAwait(continueOnCapturedContext: false);
+            }
+        }
```
GabeDeBacker, line 32: You should collect these into an array and then return Task.All

eddynaka: I can make that change since i will doing some stuff with the .spam

src/Sarif.Sarifer/IBackgroundAnalysisService.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// The absolute path of the file being analyzed, or null if the text came from a VS text
+        /// buffer that was not attached to a file.
+        /// </param>
+        /// <param name="text">
+        /// The text to analyze.
+        /// </param>
+        /// <returns>
+        /// A <see cref="Task"/> that represents the completion of the background analysis.
+        /// </returns>
+        Task StartAnalysisAsync(string path, string text);
```
GabeDeBacker, line 26: Should probably take a CancellationToken

GabeDeBacker: Same comment applies to all Async methods unless it really cannot be cancelled.

---
In reply to: [524491510](https://github.com/microsoft/sarif-visualstudio-extension/pull/266#discussion_r524491510) [](ancestors = 524491510)

src/Sarif.Sarifer/BackgroundAnalyzerBase.cs
```diff
@@ -0,0 +1,72 @@
...
+            using (Stream stream = new MemoryStream())
+            using (TextWriter writer = new StreamWriter(stream, Encoding.UTF8))
+            {
+                CreateSarifLog(path, text, writer);
+                await writer.FlushAsync().ConfigureAwait(continueOnCapturedContext: false);
+
+                foreach (IBackgroundAnalysisSink sink in sinks)
+                {
+                    stream.Seek(0L, SeekOrigin.Begin);
+                    await sink.ReceiveAsync(stream).ConfigureAwait(continueOnCapturedContext: false);
```
GabeDeBacker, line 47: >sink.ReceiveAsync(stream) [](start = 26, length = 25)

Hmm.. seeking the stream to zero won't necessary help if an implementation of ReceiveAsync does something on a completely background task or, for example, another thread.

You should consider new StreamReader(memoryStream) and pass StreamReader.BaseStream to ReceiveAsync.

That will also allow you to start all the ReceiveAsync in parallel and return Task.All()

src/Sarif.Sarifer/BackgroundAnalysisService.cs
lgolding: I'm taking care of it in a new PR where I'm also handling cancellation.

---
In reply to: [524490434](https://github.com/microsoft/sarif-visualstudio-extension/pull/266#discussion_r524490434) [](ancestors = 524490434)

src/Sarif.Sarifer/IBackgroundAnalysisService.cs
lgolding: Cancellation is now plumbed from text view creation listener all the way down into the derived analyzer classes. Will also hook it up to the task center.

---
In reply to: [524491811](https://github.com/microsoft/sarif-visualstudio-extension/pull/266#discussion_r524491811) [](ancestors = 524491811,524491510)

PR #268, opened 2020-11-11
Expose info bars for testing.

lgolding:
The Apex OM doesn't expose the info bars in a way that enables our tests, so we do it ourselves.


GabeDeBacker:
>             {

Question: Should the info-bar close when the user clicks on it? (Should that be an option)?

---
Refers to: src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs:167 in 642b07a. [](commit_id = 642b07a7993404d73015a0928576c860cd83e6a7, deletion_comment = False)

src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
```diff
@@ -19,6 +20,14 @@ namespace Microsoft.Sarif.Viewer.Controls
     /// </summary>
     public class InfoBar : IVsInfoBarUIEvents
     {
+        /// <summary>
+        /// Gets a list of the active info bar instances created by this class.
+        /// </summary>
+        /// <remarks>
+        /// Exposed for testing purposes.
+        /// </remarks>
+        public static List<InfoBar> InfoBars { get; } = new List<InfoBar>();
```
GabeDeBacker, line 29: >public [](start = 8, length = 6)

The fact that this is public and is public from the assembly means that anyone else (not just our test code) can easily get access to these.

What exactly are we going to do with this in the tests?

lgolding:
>             {

1. Good question. At the moment we don't have any examples of action items in the info bars. Maybe we don't need this API at all. YAGNI.

2. In general I think the answer is "No". The info bar has an "x" to close. I'll have to see how some of the usual VS error bars behave.

---
In reply to: [725677752](https://github.com/microsoft/sarif-visualstudio-extension/pull/268#issuecomment-725677752) [](ancestors = 725677752)

---
Refers to: src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs:167 in 642b07a. [](commit_id = 642b07a7993404d73015a0928576c860cd83e6a7, deletion_comment = False)

src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
lgolding: I no longer have this member. Instead I now have `Shown` and `Closed` events that both product code and test code can subscribe to. See the test PR for usage.

---
In reply to: [521657580](https://github.com/microsoft/sarif-visualstudio-extension/pull/268#discussion_r521657580) [](ancestors = 521657580)

lgolding: It is now internal. I tried making it public to overcome a problem that turned out to be pilot error.

---
In reply to: [521657580](https://github.com/microsoft/sarif-visualstudio-extension/pull/268#discussion_r521657580) [](ancestors = 521657580)

michaelcfanning (approved):
:shipit:

Issue #230, opened 2020-10-26
Developers should get confirmation that code analysis has run but no defects were found


michaelcfanning:
VS XP for successful analysis should keep in sync with this issue:

https://github.com/microsoft/sarif-vscode-extension/issues/365

Once PR #272: "Handle additional error bar scenarios" is merged, this will be the status for handling log files received through the stream-based interop API (which is used in the background analysis scenario):

- [x] Scenario <span>#</span>1: No log file is received. This isn't the viewer's responsibility to handle.
- [x] Scenario <span>#</span>2: Malformed file: Implemented with info bar with an red "error" icon.
- [x] Scenario <span>#</span>3: Catastrophic tool failure. In error list + summary in info bar with red "error" icon.
- [x] Scenario <span>#</span>4: Catastrophic configuration failure. In error list + summary in info bar with red "error" icon.
- [x] Scenario <span>#</span>5: No results: Will implement with info bar with a blue "information" icon.
- [x] Scenario <span>#</span>6: Analysis succeeds and there were results. Was already handled.
- [x] Scenario <span>#</span>7: Mixture of previous scenarios. No additional work implied.

We also need to...

- [x] Integrate exceptional condition handling into the file-based interop API (which is used in the WAVE scenario).
- [ ] Related: #167: "Opening multiple SARIF files will prompt the user to select "ok" for every empty SARIF file."
- [ ] Related: #279: Stream-based and file-based APIs must be at parity for downlevel SARIF and native tool output

mpande98:
Banner-driven model is possible. 

In Visual Studio, this is called the info bar. More details here: https://microsoft.visualstudio.com/DefaultCollection/EngSys/_git/devprod.wave.VsExtension?path=%2FWave.VS%2FUtilities%2FInfoBar.cs&_a=contents&version=GBtsunami

mpande98:
Scenarios:

Log file fails to materialize (no log file)
Bad log files (malformed and unexpected file exceptions) **
Log file internally indicates a catastrophic tool failure
Log file indicates a catastrophic configuration failure
Analysis succeeded but there's no results
Analysis succeeded and there are results // Goal scenario
Mixture of 6&everything before it (some results, some failures)

mpande98:
Spec linked: https://microsoft-my.sharepoint.com/:w:/p/malpande/ERFv-DPh2UJJrO_MLRQzd2YBXhson5ZBU1CEl4_wmki7WQ?e=S2XZoK

[...]

PR #272, opened 2020-11-13
Handle additional info bar scenarios

lgolding:
This is addresses #230, Scenario <span>#</span>5.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
+++ b/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
@@ -357,7 +357,8 @@ namespace Microsoft.Sarif.Viewer.ErrorList
             if (sarifLog != null)
             {
                 await ProcessSarifLogAsync(sarifLog, logFilePath: null, showMessageOnNoResults: showMessageOnNoResults, cleanErrors: cleanErrors, openInEditor: openInEditor);
-                if (sarifLog.HasNoResults())
+
+                if (!sarifLog.HasResults())
                 {
                     new InfoBar(Resources.InfoNoResultsInLog, imageMoniker: KnownMonikers.StatusInformation).ShowAsync().FileAndForget(FileAndForgetEventName.InfoBarOpenFailure);
                 }
--- a/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
+++ b/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
@@ -238,8 +238,13 @@ namespace Microsoft.Sarif.Viewer.ErrorList
 
             foreach (string logFile in logFiles)
             {
+                // The null conditional operator in the Where clause is necessary because log files
+                // that come in through the API ILoadSarifLogService.LoadSarifLog(Stream) don't have
+                // a file name. The good news is, we never close such a log file. If in future we
+                // do need to close such a log file, we'll need to synthesize a log file name so we
+                // know which runs belong to that file.
                 runIdsToClear.AddRange(CodeAnalysisResultManager.Instance.RunIndexToRunDataCache.
-                    Where(runDataCacheKvp => runDataCacheKvp.Value.LogFilePath.Equals(logFile, StringComparison.OrdinalIgnoreCase)).
+                    Where(runDataCacheKvp => runDataCacheKvp.Value.LogFilePath?.Equals(logFile, StringComparison.OrdinalIgnoreCase) == true).
                     Select(runDataCacheKvp => runDataCacheKvp.Key));
             }
 
@@ -355,6 +360,16 @@ namespace Microsoft.Sarif.Viewer.ErrorList
                 if (sarifLog.HasNoResults())
                 {
                     new InfoBar(Resources.InfoNoResultsInLog, imageMoniker: KnownMonikers.StatusInformation).ShowAsync().FileAndForget(FileAndForgetEventName.InfoBarOpenFailure);
+                }
+
+                if (sarifLog.HasErrorLevelToolConfigurationNotifications())
+                {
+                    new InfoBar(Resources.ErrorLogHasErrorLevelToolConfigurationNotifications).ShowAsync().FileAndForget(FileAndForgetEventName.InfoBarOpenFailure);
+                }
+
+                if (sarifLog.HasErrorLevelToolExecutionNotifications())
+                {
+                    new InfoBar(Resources.ErrorLogHasErrorLevelToolExecutionNotifications).ShowAsync().FileAndForget(FileAndForgetEventName.InfoBarOpenFailure);
                 }
             }
             else
--- a/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
+++ b/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
@@ -27,6 +27,7 @@ using Microsoft.Sarif.Viewer.Controls;
 using Microsoft.Sarif.Viewer.Models;
 using Microsoft.Sarif.Viewer.Sarif;
 using Microsoft.Sarif.Viewer.Tags;
+using Microsoft.VisualStudio.Imaging;
 using Microsoft.VisualStudio.PlatformUI;
 using Microsoft.VisualStudio.Shell;
 using Microsoft.VisualStudio.Shell.Interop;
@@ -351,6 +352,10 @@ namespace Microsoft.Sarif.Viewer.ErrorList
             if (sarifLog != null)
             {
                 await ProcessSarifLogAsync(sarifLog, logFilePath: null, showMessageOnNoResults: showMessageOnNoResults, cleanErrors: cleanErrors, openInEditor: openInEditor);
+                if (sarifLog.HasNoResults())
+                {
+                    new InfoBar(Resources.InfoNoResultsInLog, imageMoniker: KnownMonikers.StatusInformation).ShowAsync().FileAndForget(FileAndForgetEventName.InfoBarOpenFailure);
+                }
             }
             else
             {
--- a/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
+++ b/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
@@ -341,7 +341,7 @@ namespace Microsoft.Sarif.Viewer.ErrorList
             }
         }
 
-        internal static async Task<ExceptionalConditions> ProcessSarifLogAsync(Stream stream, bool showMessageOnNoResults, bool cleanErrors, bool openInEditor)
+        internal static async Task<ExceptionalConditions> ProcessSarifLogAsync(Stream stream, string logId, bool showMessageOnNoResults, bool cleanErrors, bool openInEditor)
         {
             SarifLog sarifLog = null;
             try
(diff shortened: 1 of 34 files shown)


src/Sarif.Viewer.VisualStudio/Sarif.Viewer.VisualStudio.csproj
```diff
@@ -6,8 +6,7 @@
   </PropertyGroup>
   <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
   <!-- Hack: https://github.com/dotnet/Nerdbank.GitVersioning/issues/404, since its a vsix with pages -->
-  <Import Project="$(MSBuildProjectExtensionsPath)$(_TargetAssemblyProjectName)$(MSBuildProjectExtension).nuget.g.props"
-    Condition=" '$(_TargetAssemblyProjectName)' != '' and '$(ImportProjectExtensionProps)' != 'false' and exists('$(MSBuildProjectExtensionsPath)$(_TargetAssemblyProjectName)$(MSBuildProjectExtension).nuget.g.props')" />
+  <Import Project="$(MSBuildProjectExtensionsPath)$(_TargetAssemblyProjectName)$(MSBuildProjectExtension).nuget.g.props" Condition=" '$(_TargetAssemblyProjectName)' != '' and '$(ImportProjectExtensionProps)' != 'false' and exists('$(MSBuildProjectExtensionsPath)$(_TargetAssemblyProjectName)$(MSBuildProjectExtension).nuget.g.props')" />
```
lgolding, line 9: >Import [](start = 3, length = 6)

Eddy, is this your change? I didn't intentionally change this.

src/Sarif.Viewer.VisualStudio/Sarif/Run.extensions.cs
```diff
@@ -12,12 +12,12 @@
 
 namespace Microsoft.Sarif.Viewer.Sarif
 {
-    static class RunExtensions
+    internal static class RunExtensions
```
lgolding, line 15: >internal [](start = 4, length = 8)

.editorconfig changes.

src/Sarif.Viewer.VisualStudio/Sarif.Viewer.VisualStudio.csproj
eddynaka: yes. this is needed to nerdbank.

src/Sarif.Viewer.VisualStudio/Sarif/SarifLog.extensions.cs
```diff
@@ -0,0 +1,31 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.CodeAnalysis.Sarif;
+
+namespace Microsoft.Sarif.Viewer.Sarif
+{
+    /// <summary>
+    /// Extension methods for the <see cref="SarifLog"/> class used by the viewer.
+    /// </summary>
+    internal static class SarifLogExtensions
```
lgolding, line 11: >SarifLogExtensions [](start = 26, length = 18)

I know that our usual practice is to put all extension methods in a single file. At the moment there are more than a dozen per-class extension method classes, and I'm following the pattern of the existing code. It should be cleaned up all at once.

src/Sarif.Viewer.VisualStudio/Sarif/SarifLog.extensions.cs
```diff
@@ -0,0 +1,31 @@
...
+        static internal bool HasNoResults(this SarifLog sarifLog)
+        {
+            if (sarifLog.Runs == null)
+            {
+                return false;
+            }
+
+            foreach (Run run in sarifLog.Runs)
+            {
+                if (!run.HasNoResults())
```
lgolding, line 22: >!run.HasNoResults() [](start = 20, length = 19)

BEHAVIORAL QUESTION: Do we want to show the bar if _any_ run in the log lacks results, or only if _all_ of them lack results (which is what I'm doing here)?

IMO the behavior here is correct. If any log has a result, then _something_ will show up in the error list, and the user won't wonder if the analysis finished. The idea is to show give the user explicit feedback in case _nothing_ shows up in the error list.

@mikefan

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -351,6 +357,20 @@ internal static async Task ProcessSarifLogAsync(Stream stream, bool showMessageO
...
+                }
+
+                if (sarifLog.HasErrorLevelToolConfigurationNotifications())
+                {
+                    new InfoBar(Resources.ErrorLogHasErrorLevelToolConfigurationNotifications).ShowAsync().FileAndForget(FileAndForgetEventName.InfoBarOpenFailure);
+                }
+
+                if (sarifLog.HasErrorLevelToolExecutionNotifications())
+                {
+                    new InfoBar(Resources.ErrorLogHasErrorLevelToolExecutionNotifications).ShowAsync().FileAndForget(FileAndForgetEventName.InfoBarOpenFailure);
```
eddynaka, line 372: Is it possible to have a sarif log with no results and the tool configuration contains an error?
Why: if that happen, we will show many issues in the `InfoBar`. Is that expected?

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -351,6 +357,20 @@ internal static async Task ProcessSarifLogAsync(Stream stream, bool showMessageO
             if (sarifLog != null)
             {
                 await ProcessSarifLogAsync(sarifLog, logFilePath: null, showMessageOnNoResults: showMessageOnNoResults, cleanErrors: cleanErrors, openInEditor: openInEditor);
+                if (sarifLog.HasNoResults())
```
GabeDeBacker, line 360: This is an example of where I would expect ProcessSarifLogAsync to return something like Task<ProcessLogResults> which would contain information that is being inspected here. The caller of this should be doing UI work.  I.E Let's not do UI stuff directly in the error service

src/Sarif.Viewer.VisualStudio/Sarif/SarifLog.extensions.cs
```diff
@@ -0,0 +1,67 @@
...
+using Microsoft.CodeAnalysis.Sarif;
+
+namespace Microsoft.Sarif.Viewer.Sarif
+{
+    /// <summary>
+    /// Extension methods for the <see cref="SarifLog"/> class used by the viewer.
+    /// </summary>
+    internal static class SarifLogExtensions
+    {
+        static internal bool HasNoResults(this SarifLog sarifLog)
```
GabeDeBacker, line 13: So having functions that return negative flags is a bit confusing when they are combined with boolean operators.

I.e. 
if (!HasResults()) is a lot easier to understand mentally then if (!HasNoResults())

can we perhaps change this function to return a positive name rather than a negative one?

(I had to re-read line 22 like 10 times to make sure I understood it)

src/Sarif.Viewer.VisualStudio/Sarif/Run.extensions.cs
```diff
@@ -98,5 +98,43 @@ public static string GetToolName(this Run run)
 
             return run.Tool.Driver.FullName ?? run.Tool.Driver.Name;
         }
+
+        public static bool HasNoResults(this Run run) => run.Results?.Count == 0;
```
GabeDeBacker, line 102: >Count [](start = 70, length = 5)

I love .net using signed values everywhere... so if .Count < 0 this returns true?

src/Sarif.Viewer.VisualStudio/Sarif/SarifLog.extensions.cs
lgolding: +100. Fixed.

---
In reply to: [523438956](https://github.com/microsoft/sarif-visualstudio-extension/pull/272#discussion_r523438956) [](ancestors = 523438956)

src/Sarif.Viewer.VisualStudio/Sarif/Run.extensions.cs
lgolding: Reversed the logic as you suggested in another comment, so the logic is now `run.Results?.Count > 0`.

---
In reply to: [523439091](https://github.com/microsoft/sarif-visualstudio-extension/pull/272#discussion_r523439091) [](ancestors = 523439091)

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
lgolding: Agreed. The actual logic for deciding which bars to show (both in the single-log and multiple-log case) is non-trivial, and easily separable and unit testable, so I'm going to refactor to that. This will help address Eddy's question (can you have no results _and_ a configuration error? the answer is "yes") in a way that will be easy to see.

I propose:
- If a given log file has a catastrophic error, do _not_ display the "no results" bar, regardless of whether the log contains results.
- Across multiple logs, display each "kind" of bar only once.

The result is that you will see at most one _each_ of "no results", "catastrophic configuration error", "catastrophic execution error", and "invalid SARIF file" across any set of logs.

---
In reply to: [523438595](https://github.com/microsoft/sarif-visualstudio-extension/pull/272#discussion_r523438595) [](ancestors = 523438595)

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
lgolding: Yes, it is possible to have such a log. In fact it is likely: If the tool detects a catastrophic configuration error, it probably won't produce results. And you're right, we don't want multiple bars in that case.

But if we are sent two logs, and one has "no results" and the other has "catastrophic configuration error", then IMO we _should_ show one of each kind of bar. What we don't want is 50 "no results" bars or 50 "catastrophic configuration error" bars. That's why the messages talk about "one or more SARIF logs".

See also my answer to Gabe. Your two comments, taken together, suggest some additional logic, and a refactoring to separate out that logic and make it unit testable, which I will now do.

---
In reply to: [523400552](https://github.com/microsoft/sarif-visualstudio-extension/pull/272#discussion_r523400552) [](ancestors = 523400552)

lgolding: This is now done.

---
In reply to: [523445875](https://github.com/microsoft/sarif-visualstudio-extension/pull/272#discussion_r523445875) [](ancestors = 523445875,523400552)

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
lgolding: This is now done.

---
In reply to: [523445509](https://github.com/microsoft/sarif-visualstudio-extension/pull/272#discussion_r523445509) [](ancestors = 523445509,523438595)

GabeDeBacker (approved):
:shipit:

michaelcfanning:
> // Copyright (c) Microsoft. All rights reserved.

The casing of this file name is wrong. I can fix it or have Eddy do it later.

---
Refers to: src/Sarif.Viewer.VisualStudio/Sarif/SarifLog.extensions.cs:1 in 74a3d5b. [](commit_id = 74a3d5b63c3c6a8583694499cce90a1f3e071b0b, deletion_comment = False)

michaelcfanning (changes requested):
:clock1:

michaelcfanning (approved):
:shipit:

lgolding:
> // Copyright (c) Microsoft. All rights reserved.

Agreed. These are existing files. I'd appreciate it if Eddy could do it later, or -- as you may have seen in my other comment -- we might instead want to gather all the extensions into a single file `SarifExtensions` rather than having one file per extended class. This is something you've requested in the past in other contexts.

---
In reply to: [728296872](https://github.com/microsoft/sarif-visualstudio-extension/pull/272#issuecomment-728296872) [](ancestors = 728296872)

---
Refers to: src/Sarif.Viewer.VisualStudio/Sarif/SarifLog.extensions.cs:1 in 74a3d5b. [](commit_id = 74a3d5b63c3c6a8583694499cce90a1f3e071b0b, deletion_comment = False)

PR #274, opened 2020-11-17
Cancellation token plumbing

lgolding:


src/Sarif.Sarifer/BackgroundAnalysisService.cs
```diff
@@ -23,12 +25,14 @@ internal class BackgroundAnalysisService : IBackgroundAnalysisService
...
         /// <inheritdoc/>
-        public async Task StartAnalysisAsync(string path, string text)
+        public async Task StartAnalysisAsync(string path, string text, CancellationToken cancellationToken)
         {
-            foreach (IBackgroundAnalyzer analyzer in this.analyzers)
-            {
-                await analyzer.StartAnalysisAsync(path, text).ConfigureAwait(continueOnCapturedContext: false);
-            }
+            Task[] tasks = this.analyzers
+                .Select(async a => await a.StartAnalysisAsync(path, text, cancellationToken).ConfigureAwait(continueOnCapturedContext: false))
```
GabeDeBacker, line 31: >async [](start = 24, length = 5)

remove async, and await and configurate await from this line.

this should really be just

return Task.WhenAll(this.analyzers.Select(analyzer => analyzer.StartAnalysisAsnc(path, text cancellationToken);

short sweet one-liner

src/Sarif.Sarifer/FirstViewAddedEventArgs.cs
```diff
@@ -24,10 +25,15 @@ public class FirstViewAddedEventArgs : EventArgs
         /// <param name="text">
         /// The contents of the text buffer being viewed.
         /// </param>
-        public FirstViewAddedEventArgs(string path, string text)
+        /// <param name="cancellationToken">
+        /// A <see cref="CancellationToken"/> that can be used to cancel analysis of the buffer
+        /// being viewed.
+        /// </param>
+        public FirstViewAddedEventArgs(string path, string text, CancellationToken cancellationToken)
```
GabeDeBacker, line 32: >, CancellationToken cancellationToken [](start = 63, length = 37)

This doesn't need to be here.

src/Sarif.Sarifer/TextBufferViewTracker.cs
```diff
@@ -43,7 +43,7 @@ public void AddTextView(ITextView textView, string path, string text)
             trackingInformation.Add(textView);
             if (first)
             {
-                FirstViewAdded?.Invoke(this, new FirstViewAddedEventArgs(path, text));
+                FirstViewAdded?.Invoke(this, new FirstViewAddedEventArgs(path, text, trackingInformation.CancellationToken));
```
GabeDeBacker, line 46: >, trackingInformation.CancellationToken) [](start = 83, length = 40)

It's not standard to pass a cancellation token in an event.. I would remove it here.

src/Sarif.Sarifer/TextBufferViewTrackingInformation.cs
```diff
@@ -22,6 +23,7 @@ internal TextBufferViewTrackingInformation()
         {
             Views = new List<ITextView>();
             LogId = Guid.NewGuid().ToString();
+            CancellationToken = new CancellationToken();
```
GabeDeBacker, line 26: >CancellationToken = new CancellationToken(); [](start = 12, length = 44)

Here you would create a CancellationTokenSource. Usually you would use CancellationTokenSource<bool> since it doesn't support void (sadness).

src/Sarif.Sarifer/TextBufferViewTracker.cs
GabeDeBacker: On second thought. :) If this event is used internally to track views :) I would pass the whole trackingInformation.

---
In reply to: [524853366](https://github.com/microsoft/sarif-visualstudio-extension/pull/274#discussion_r524853366) [](ancestors = 524853366)

src/Sarif.Sarifer/BackgroundAnalysisTextViewCreationListener.cs
```diff
@@ -77,7 +83,8 @@ public void TextViewCreated(ITextView textView)
 
         private void TextBufferViewTracker_FirstViewAdded(object sender, FirstViewAddedEventArgs e)
         {
-            this.backgroundAnalysisService.Value.StartAnalysisAsync(e.Path, e.Text).FileAndForget(FileAndForgetEventName.BackgroundAnalysisFailure);
+            Task task = this.backgroundAnalysisService.Value.StartAnalysisAsync(e.Path, e.Text, e.CancellationToken);
```
GabeDeBacker, line 86: > e.CancellationToken); [](start = 95, length = 22)

FirstViewAddedEventArgs should contain the "tracking information" and what should be passed in for the CancellationToken would be something like TextBufferViewTrackingInformation.TokenSource.Token.

src/Sarif.Sarifer/TextBufferViewTrackingInformation.cs
```diff
@@ -38,6 +40,12 @@ internal TextBufferViewTrackingInformation()
         /// </remarks>
         internal string LogId { get; }
 
+        /// <summary>
+        /// Gets a <see cref="CancellationToken"/> that can be used to cancel analysis of this
+        /// <see cref="ITextBuffer"/>.
+        /// </summary>
+        internal CancellationToken CancellationToken { get; }
```
GabeDeBacker, line 47: >CancellationToken [](start = 17, length = 17)

CancellationTokenSource :)

GabeDeBacker:
>     public class FirstViewAddedEventArgs : EventArgs

Was thinking about this a bit.. Not sure you need two different event arg classes do you? Do they carry the same payload?

---
Refers to: src/Sarif.Sarifer/FirstViewAddedEventArgs.cs:16 in 377db57. [](commit_id = 377db579deb4f00a18131bcd14f2e8edc0d1edb7, deletion_comment = False)

GabeDeBacker:
>         {

In this event handler, assuming that LastViewRemovedEventArgs contains the buffer tracking information, then you could call Cancel on the CancellationTokenSource contained within it.

(Of course the tracking information may not need to be exposed in the event args at all and then buffer tracker can call Cancel before it fires this event)

---
Refers to: src/Sarif.Sarifer/BackgroundAnalysisTextViewCreationListener.cs:98 in 377db57. [](commit_id = 377db579deb4f00a18131bcd14f2e8edc0d1edb7, deletion_comment = False)

eddynaka:
Closing this issue, since we merged this: https://github.com/microsoft/sarif-visualstudio-extension/pull/291

PR #275, opened 2020-11-17
Creating Spam background analyzer

eddynaka:


src/Sarif.Sarifer/SpamBackgroundAnalyzer.cs
```diff
@@ -0,0 +1,163 @@
...
+using System.ComponentModel.Composition;
+using System.IO;
+using System.Text.RegularExpressions;
+
+using Microsoft.CodeAnalysis.Sarif.Writers;
+
+namespace Microsoft.CodeAnalysis.Sarif.Sarifer
+{
+    [Export(typeof(IBackgroundAnalyzer))]
+    internal class SpamBackgroundAnalyzer : BackgroundAnalyzerBase, IBackgroundAnalyzer
```
lgolding, line 15: >IBackgroundAnalyzer [](start = 68, length = 19)

I know that my POC analyzer did this, but it's not necessary since `BackgroundAnalyzerBase` implements `IBackgroundAnalyzer`. I suggest you remove it from both concrete analyzers, and just make sure that the MEF export still works even if the interface is implemented in a base class. By "make sure the MEF export still works" I mean "make sure the analysis still runs."

src/Sarif.Sarifer/Rule.cs
```diff
@@ -0,0 +1,28 @@
...
+            this.SearchPattern = searchPattern;
+            this.ReplacePattern = replacePattern;
+            this.Description = description;
+            this.Message = message;
+        }
+
+        public Rule()
+        {
+
+        }
```
lgolding, line 26: Default ctor not necessary

src/Sarif.Sarifer/Rule.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.CodeAnalysis.Sarif.Sarifer
+{
+    internal class Rule
```
lgolding, line 6: >Rule [](start = 19, length = 4)

Let's call it `SpamRule`. This isn't really a SARIF rule (SARIF rules are actually `ReportingDescriptor`s), and I don't want there to be confusion.

src/Sarif.Sarifer/Rule.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        public string Id { get; set; }
```
lgolding, line 8: >set;  [](start = 32, length = 5)

Since these are all initialized in the ctor and we don't expect them to change, make them `get`-only.

src/Sarif.Sarifer/SpamBackgroundAnalyzer.cs
lgolding, line 15: >SpamBackgroundAnalyzer [](start = 19, length = 22)

You know, this is so similar to the POC analyzer -- what do you think of the idea of just replacing the POC analyzer? Just renaming it to `SpamBackgroundAnalyzer`, and making the few small changes you made here? We don't really need both. The POC analyzer was just that -- a proof of concept. The concept is now proven :-)

lgolding: (I'm genuinely asking your opinion here.)

---
In reply to: [528043775](https://github.com/microsoft/sarif-visualstudio-extension/pull/275#discussion_r528043775) [](ancestors = 528043775)

eddynaka: If we remove `IBackgroundAnalyzer` from `BackgroundAnalyerBase`, the service won't find the analyzer. 

---
In reply to: [528042824](https://github.com/microsoft/sarif-visualstudio-extension/pull/275#discussion_r528042824) [](ancestors = 528042824)

lgolding: Sorry, I was suggesting that you remove the explicit declaration of the interface from this file, `SpamBackgroundAnalyzer` -- not from `BackgroundAnalyzerBase`. Assuming that works, I am ok either way. Remove it or leave it alone, that's fine.

---
In reply to: [528229483](https://github.com/microsoft/sarif-visualstudio-extension/pull/275#discussion_r528229483) [](ancestors = 528229483,528042824)

lgolding (approved):
:shipit:

Issue #114, opened 2020-02-28
Viewer must filter out 'absent' and 'suppressed' issues by default


lgolding:
For now, we should remove those issues entirely. In future, we can consider a UI affordance to display them.

michaelcfanning:
We should make this issue a priority. For both of the 'absent' and 'suppressed' states, our behavior s/be to determine, at load time, whether the SARIF file contains either of these classes of issues. The SARIF spec states that all results must have these optional properties (either `result.baselineState` or `result.Suppressions`) or none of them will (so the viewer only has to examine a single result to determine whether these properties are in play). If the VS column relevant to the property (suppressions or whatever column governs the absent state) is *not* visible, it should be made visible and the column s/be configured to remove the suppressed or absent items from the view. We show the column as a UI clue that there are results which have been filtered away (increasing discoverability in cases where the user would like to check relevant filters to see absent/suppressed results).

In the case where this column is already visible, it should remain visible. We should investigate to see whether it's appropriate to retain previous column settings (if the user has enable viewing of suppressed/absent).

There is an established column for suppressed items in recent VS. Historically, the VS viewer has overloaded the 'Category' column for baseline state, but we might want to reconsider this.

For suppressed items, there is the additional wrinkle that many suppressions can be applied to a result, so we need to define a default behavior in this case. As a start, we could consider regarding a result as suppressed in cases where this is an active suppression, which is either in 'under review' or 'rejected' states (or if no state is given).

mpande98:
- Don't want to overwhelm user with multiple suppressions
- Show column that allows you to view filtered items ` 
- For absent scenario, would we consider a slider paradigm in the SARIF viewer tab? 

More discussion required.

lgolding:
@michaelcfanning wrote:

> As a start, we could consider regarding a result as suppressed in cases where this is an active suppression, which is either in 'under review' or 'rejected' states (or if no state is given).

Why would a "rejected" suppression cause us to regard the result as "suppressed"?

The [SARIF Tutorials](https://github.com/microsoft/sarif-tutorials) article [Displaying results in a viewer](https://github.com/microsoft/sarif-tutorials/blob/master/docs/Displaying-results-in-a-viewer.md) suggests:

> If the status of _any_ of the suppressions is `"underReview"` or `"rejected"`, then the result should _not_ be considered suppressed. Otherwise, the result _should_ be considered suppressed.

PR #276, opened 2020-11-17
First step in #114: Provide helpers to calculate absence and suppression state.

lgolding:


src/Sarif.Viewer.VisualStudio/Sarif/Run.extensions.cs
```diff
@@ -136,5 +136,48 @@ public static bool HasErrorLevelToolExecutionNotifications(this Run run)
 
             return false;
         }
+
+        /// <summary>
```
lgolding, line 140: >/// [](start = 8, length = 3)

Would these helpers be better off in the SDK?

GabeDeBacker (approved):
:shipit:

src/Sarif.Viewer.VisualStudio/Sarif/Result.extensions.cs
```diff
@@ -2,6 +2,9 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information. 
 
 using System;
```
michaelcfanning, line 4: >System [](start = 6, length = 6)

Can you please get these files renamed properly to Result.Extensions.cs, etc., in your next change?

src/Sarif.Viewer.VisualStudio/Sarif/Run.extensions.cs
michaelcfanning: Agreed, nice helpers. Create an SDK issue?

---
In reply to: [525564884](https://github.com/microsoft/sarif-visualstudio-extension/pull/276#discussion_r525564884) [](ancestors = 525564884)

src/Sarif.Viewer.VisualStudio/Sarif/Result.extensions.cs
lgolding: You're looking for just the case change 'e' => 'E'? Sure.

src/Sarif.Viewer.VisualStudio/Sarif/Run.extensions.cs
lgolding: [#2157](https://github.com/microsoft/sarif-sdk/issues/2157): Migrate suppressed and absent issue helpers from VS extension to SDK

---
In reply to: [527881875](https://github.com/microsoft/sarif-visualstudio-extension/pull/276#discussion_r527881875) [](ancestors = 527881875,525564884)

PR #278, opened 2020-11-18
Fix #230: Provide info bars through file-based APIs

### LINKED ISSUE #167: Opening multiple SARIF files will prompt user to select "okay" for every empty SARIF file
Unfortunately as a part of the opening process...for every file that does not have SARIF issues in a multi-file-open, it prompts you to select "Okay"
image

### LINKED ISSUE #230: Developers should get confirmation that code analysis has run but no defects were found

lgolding:
Clients of the file-based viewer interop API now see info bars now appear for the same "exceptional conditions" as clients of the stream-based API.

The basic change that makes this possible is to have the `ErrorListService` raise an event when it finishes processing a log, rather than for `ErrorListService` APIs to _return_ the set of exceptional conditions. This is necessary because the file-based APIs are run in the background with "file and forget", so we never get a chance to see a return value.

Also:
- Fix #167: Opening multiple SARIF files will prompt user to select "ok" for every empty SARIF file.
    This actually no longer happened, because there was no code path that popped the dialog. But the code to show the dialog was still there, and I removed it.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
+++ b/src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
@@ -63,37 +63,34 @@ namespace Microsoft.Sarif.Viewer.Controls
         /// </param>
         internal static async Task CreateInfoBarsForExceptionalConditionsAsync(ExceptionalConditions conditions)
         {
-            using (s_infoBarLock.EnterWriteLock())
-            {
-                // The most recently shown bar is displayed at the bottom, so show the bars in order
-                // of decreasing severity. Note that this only works within the info bars shown for
-                // a single SARIF log. Across logs, the "most recent" rule means that the bars are
-                // show in the order the logs are processed. So if the first log has an info-level bar
-                // and the second has an error-level bar, the info-level bar will be on top.
-                await AddInfoBarIfRequiredAsync(
-                    conditions,
-                    ExceptionalConditions.InvalidJson,
-                    Resources.ErrorInvalidSarifStream,
-                    KnownMonikers.StatusError);
+            // The most recently shown bar is displayed at the bottom, so show the bars in order
+            // of decreasing severity. Note that this only works within the info bars shown for
+            // a single SARIF log. Across logs, the "most recent" rule means that the bars are
+            // show in the order the logs are processed. So if the first log has an info-level bar
+            // and the second has an error-level bar, the info-level bar will be on top.
+            await AddInfoBarIfRequiredAsync(
+                conditions,
+                ExceptionalConditions.InvalidJson,
+                Resources.ErrorInvalidSarifStream,
+                KnownMonikers.StatusError);
 
-                await AddInfoBarIfRequiredAsync(
-                    conditions,
-                    ExceptionalConditions.ExecutionError,
-                    Resources.ErrorLogHasErrorLevelToolExecutionNotifications,
-                    KnownMonikers.StatusError);
+            await AddInfoBarIfRequiredAsync(
+                conditions,
+                ExceptionalConditions.ExecutionError,
+                Resources.ErrorLogHasErrorLevelToolExecutionNotifications,
+                KnownMonikers.StatusError);
 
-                await AddInfoBarIfRequiredAsync(
-                    conditions,
-                    ExceptionalConditions.ConfigurationError,
-                    Resources.ErrorLogHasErrorLevelToolConfigurationNotifications,
-                    KnownMonikers.StatusError);
+            await AddInfoBarIfRequiredAsync(
+                conditions,
+                ExceptionalConditions.ConfigurationError,
+                Resources.ErrorLogHasErrorLevelToolConfigurationNotifications,
+                KnownMonikers.StatusError);
 
-                await AddInfoBarIfRequiredAsync(
-                    conditions,
-                    ExceptionalConditions.NoResults,
-                    Resources.InfoNoResultsInLog,
-                    KnownMonikers.StatusInformation);
-            }
+            await AddInfoBarIfRequiredAsync(
+                conditions,
+                ExceptionalConditions.NoResults,
+                Resources.InfoNoResultsInLog,
+                KnownMonikers.StatusInformation);
         }
 
         /// <summary>
@@ -230,13 +227,23 @@ namespace Microsoft.Sarif.Viewer.Controls
             string message,
             ImageMoniker imageMoniker)
         {
-            // It's safe to access and update the dictionary because this method is called inside
-            // the write lock.
-            if ((detectedConditions & individualCondition) == individualCondition && !s_infoBarToConditionDictionary.Values.Contains(individualCondition))
+            InfoBar infoBar = null;
+
+            if ((detectedConditions & individualCondition) == individualCondition)
             {
-                var infoBar = new InfoBar(message, imageMoniker: imageMoniker);
+                using (s_infoBarLock.EnterWriteLock())
+                {
+                    if (!s_infoBarToConditionDictionary.Values.Contains(individualCondition))
+                    {
+                        infoBar = new InfoBar(message, imageMoniker: imageMoniker);
+                        s_infoBarToConditionDictionary.Add(infoBar, individualCondition);
+                    }
+                }
+            }
+
+            if (infoBar != null)
+            {
                 await infoBar.ShowAsync();
-                s_infoBarToConditionDictionary.Add(infoBar, individualCondition);
             }
         }
     }
--- a/src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
+++ b/src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Threading;
 
@@ -47,7 +48,7 @@ namespace Microsoft.Sarif.Viewer.Controls
         private uint eventCookie;
 
         private static readonly ReaderWriterLockSlimWrapper s_infoBarLock = new ReaderWriterLockSlimWrapper(new ReaderWriterLockSlim());
-        private static readonly IDictionary<InfoBar, ExceptionalConditions> s_infoBarToConditionDictionary = new Dictionary<InfoBar, ExceptionalConditions>();
+        private static readonly IDictionary<InfoBar, ExceptionalConditions> s_infoBarToConditionDictionary = new ConcurrentDictionary<InfoBar, ExceptionalConditions>();
 
         /// <summary>
         /// Display info bars appropriate to the specified set of "exceptional conditions."
--- a/src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
+++ b/src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
@@ -226,7 +226,7 @@ namespace Microsoft.Sarif.Viewer.Controls
         {
             // It's safe to access and update the dictionary because this method is called inside
             // the write lock.
-            if ((detectedConditions & individualCondition) != 0 && !s_infoBarToConditionDictionary.Values.Contains(individualCondition))
+            if ((detectedConditions & individualCondition) == individualCondition && !s_infoBarToConditionDictionary.Values.Contains(individualCondition))
             {
                 var infoBar = new InfoBar(message, imageMoniker: imageMoniker);
                 await infoBar.ShowAsync();
--- a/src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
+++ b/src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
@@ -83,8 +83,8 @@ namespace Microsoft.Sarif.Viewer.Controls
                     KnownMonikers.StatusError);
 
                 await AddInfoBarIfRequiredAsync(
-                    ExceptionalConditions.NoResults,
                     conditions,
+                    ExceptionalConditions.NoResults,
                     Resources.InfoNoResultsInLog,
                     KnownMonikers.StatusInformation);
             }
--- a/src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
+++ b/src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
@@ -48,6 +48,11 @@ namespace Microsoft.Sarif.Viewer.Controls
         private uint eventCookie;
 
         private static readonly ReaderWriterLockSlimWrapper s_infoBarLock = new ReaderWriterLockSlimWrapper(new ReaderWriterLockSlim());
+
+        // It might seem more natural to use the condition as the dictionary key, and the InfoBar
+        // as the value. We wrote it this way because when the user closes an InfoBar, it's easier
+        // to remove it from the dictionary if the InfoBar itself is the key -- rather than having
+        // to look up the KeyValuePair that has that InfoBar as its value. See CloseAsync below.
         private static readonly IDictionary<InfoBar, ExceptionalConditions> s_infoBarToConditionDictionary = new ConcurrentDictionary<InfoBar, ExceptionalConditions>();
 
         /// <summary>
--- a/src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
+++ b/src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
@@ -177,8 +177,11 @@ namespace Microsoft.Sarif.Viewer.Controls
                 // Close the info bar to correctly send OnClosed() event to the Shell
                 this.uiElement.Close();
 
-                InfoBars.Remove(this.infoBarModel);
-                s_infoBarToConditionDictionary.Remove(this);
+                using (s_infoBarLock.EnterWriteLock())
+                {
+                    InfoBars.Remove(this.infoBarModel);
+                    s_infoBarToConditionDictionary.Remove(this);
+                }
 
                 this.uiElement = null;
             }
@@ -216,6 +219,8 @@ namespace Microsoft.Sarif.Viewer.Controls
             string message,
             ImageMoniker imageMoniker)
         {
+            // It's safe to access and update the dictionary because this method is called inside
+            // the write lock.
             if ((detectedConditions & individualCondition) != 0 && !s_infoBarToConditionDictionary.Values.Contains(individualCondition))
             {
                 var infoBar = new InfoBar(message, imageMoniker: imageMoniker);
--- a/src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
+++ b/src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
@@ -55,29 +55,34 @@ namespace Microsoft.Sarif.Viewer.Controls
         /// <param name="conditions">
         /// The conditions that require an info bar to be shown.
         /// </param>
-        internal static void CreateInfoBarsForExceptionalConditions(ExceptionalConditions conditions)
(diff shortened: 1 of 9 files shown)


src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -350,14 +364,13 @@ internal static async Task<ExceptionalConditions> ProcessSarifLogAsync(Stream st
...
             {
+                RaiseLogProcessed(ExceptionalConditions.InvalidJson);
             }
 
             if (sarifLog != null)
             {
                 await ProcessSarifLogAsync(sarifLog, logFilePath: logId, showMessageOnNoResults: showMessageOnNoResults, cleanErrors: cleanErrors, openInEditor: openInEditor);
             }
-
-            return ExceptionalConditionsCalculator.Calculate(sarifLog);
```
lgolding, line 360: >ExceptionalConditionsCalculator [](start = 19, length = 31)

Moved into `ProcessSarifLogAsync(SarifLog)` overload. That's what makes it work with the file-based APIs, because everything ultimately calls into the `SarifLog` overload.

src/Sarif.Viewer.VisualStudio/LogProcessedEventArgs.cs
```diff
@@ -0,0 +1,19 @@
...
+    internal class LogProcessedEventArgs
+    {
+        internal LogProcessedEventArgs(ExceptionalConditions exceptionalConditions)
+        {
+            ExceptionalConditions = exceptionalConditions;
+        }
+
+        // Gets any exceptional conditions (for example, an error-level tool execution
+        // failure) that occurred during the processing of the log file.
+        internal ExceptionalConditions ExceptionalConditions { get; }
```
GabeDeBacker, line 17: >internal [](start = 8, length = 8)

The coding style is really up to you, but double internal is something we typically don't do. The class can be internal, and this method is public. (in my opinioin)

Great article on this

[https://ericlippert.com/2014/09/15/internal-or-public/](https://ericlippert.com/2014/09/15/internal-or-public/)

And the "accepted answer" on stack-overflow
[https://stackoverflow.com/questions/9302236/why-use-a-public-method-in-an-internal-class#:~:text=If%20the%20class%20is%20internal%2C%20it%20doesn%27t%20matter,as%20part%20of%20the%20description%20of%20the%20method](https://stackoverflow.com/questions/9302236/why-use-a-public-method-in-an-internal-class#:~:text=If%20the%20class%20is%20internal%2C%20it%20doesn%27t%20matter,as%20part%20of%20the%20description%20of%20the%20method).

michaelcfanning: Haven't seen this issue come up in a while! :) I agree with Eric's point: internal classes should organize their methods in a way that clarifies what is 'public' surface area (given the constraint that the containing scope is restricted to the assembly and any 'internals visible to' attribute. The way I frame this is imagine that the class itself is appropriately made public (i.e., cross-assembly) in the future: the accessibility modifiers should still make sense. Note that this *does* lead to the possibility of internal/internal classification, for things like test helpers that would never be a good candidate for cross-assembly access.

---
In reply to: [527123309](https://github.com/microsoft/sarif-visualstudio-extension/pull/278#discussion_r527123309) [](ancestors = 527123309)

lgolding: Ok, `internal`/`public` makes sense. Note that either (1) When you make the class public, you have to write the XML doc comments then, or (2) You write XML doc comment up front on anything marked `public`. I assume @mikefan@microsoft.com wants #1.

---
In reply to: [527127708](https://github.com/microsoft/sarif-visualstudio-extension/pull/278#discussion_r527127708) [](ancestors = 527127708,527123309)

eddynaka: i think that only applies if the class is public...but i might be wrong..

lgolding: You are right, the _requirement_ for XML comments is only if the class is public. And as long as the team understands that the price of making a class public includes writing the comments at the time the decision is made, then all is well.

---
In reply to: [527130151](https://github.com/microsoft/sarif-visualstudio-extension/pull/278#discussion_r527130151) [](ancestors = 527130151)

src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
```diff
@@ -197,5 +217,21 @@ public void OnActionItemClicked(IVsInfoBarUIElement infoBarUIElement, IVsInfoBar
...
+
+        private static async Task AddInfoBarIfRequiredAsync(
+            ExceptionalConditions detectedConditions,
+            ExceptionalConditions individualCondition,
+            string message,
+            ImageMoniker imageMoniker)
+        {
+            // It's safe to access and update the dictionary because this method is called inside
+            // the write lock.
+            if ((detectedConditions & individualCondition) == individualCondition && !s_infoBarToConditionDictionary.Values.Contains(individualCondition))
```
xiangyuMS, line 229: I have a feeling that the ExceptionalConditions value should be unique in this Dictionary. Don't you think the key should be an ExceptionalConditions value, instead of a InforBar object?
Feel free to correct me if I am wrong :)

xiangyuMS, line 229: Why is it not just "detectedConditions == individualCondition "? In c#,  enum type values can be comparied with "==".

xiangyuMS: With ExceptionalConditions value as key, if an InfoBar already exists, updating it's message, and reloading/refreshing the infoBar display probably is a way for multiple similar categories of information to display.

src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
```diff
@@ -45,32 +46,47 @@ public class InfoBar : IVsInfoBarUIEvents
         private IVsInfoBarUIElement uiElement;
         private uint eventCookie;
 
+        private static readonly ReaderWriterLockSlimWrapper s_infoBarLock = new ReaderWriterLockSlimWrapper(new ReaderWriterLockSlim());
+        private static readonly IDictionary<InfoBar, ExceptionalConditions> s_infoBarToConditionDictionary = new Dictionary<InfoBar, ExceptionalConditions>();
```
xiangyuMS, line 50: In a multi-thread environment, ConcurrentDictionary<> may be better to use.

src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
lgolding: As the code stands, the logic enforces that there's only one `InfoBar` per condition value. I actually did it that way for a reason, but since it's not clear the the reader I'll add a "why" comment explaining why.

---
In reply to: [527201705](https://github.com/microsoft/sarif-visualstudio-extension/pull/278#discussion_r527201705) [](ancestors = 527201705)

lgolding: Because `detectedConditions` is a `Flags` enum -- the same log can exhibit multiple exceptional conditions. I need to know if this particular bit is turned on, regardless of what other bits might be turned one.

---
In reply to: [527204867](https://github.com/microsoft/sarif-visualstudio-extension/pull/278#discussion_r527204867) [](ancestors = 527204867)

lgolding: You're right, it does seem natural to use the condition as the key. I'll add a comment explaining why I did it this way.

And is so often the case when you write a comment, I might discover that there's a better way :-)

---
In reply to: [527210678](https://github.com/microsoft/sarif-visualstudio-extension/pull/278#discussion_r527210678) [](ancestors = 527210678)

src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
lgolding: I didn't think of that. Thanks for suggesting it. I'm not sure it will work, though.

Notice the four calls to `AddInfoBarIfRequiredAsync`  below. I want the whole _set_ of operations to be atomic. Also, within each call to `AddInfoBarIfRequiredAsync`, I want the logic that checks whether the specified bit is set to be atomic with the dictionary update.

Does this make sense? Given that, do you think I can replace the lock with a `ConcurrentDictionary`?

---
In reply to: [527217732](https://github.com/microsoft/sarif-visualstudio-extension/pull/278#discussion_r527217732) [](ancestors = 527217732)

src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
xiangyuMS: > Because `detectedConditions` is a `Flags` enum -- the same log can exhibit multiple exceptional conditions. I need to know if this particular bit is turned on, regardless of what other bits might be turned one.
> 
> In reply to: [527204867](https://github.com/microsoft/sarif-visualstudio-extension/pull/278#discussion_r527204867) [](ancestors = 527204867)

Get it. Didn't notice this. Thanks for the info.

src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
lgolding: Come to think of it, the sequence of four calls does _not_ have to be atomic. If I add one error bar, and _you_ add the one I was _about_ to add before I get around to it, then no harm is done. When I try to add it, I will find that it's already there.

Let me look more closely at the internals of `AddInfoBarIfRequiredAsync` to see if its entire body needs to be atomic, or whether I really can get away with using `ConcurrentDictionary`.

---
In reply to: [527233531](https://github.com/microsoft/sarif-visualstudio-extension/pull/278#discussion_r527233531) [](ancestors = 527233531,527217732)

xiangyuMS: > Come to think of it, the sequence of four calls does _not_ have to be atomic. If I add one error bar, and _you_ add the one I was _about_ to add before I get around to it, then no harm is done. When I try to add it, I will find that it's already there.
> 
> Let me look more closely at the internals of `AddInfoBarIfRequiredAsync` to see if its entire body needs to be atomic, or whether I really can get away with using `ConcurrentDictionary`.
> 
> In reply to: [527233531](https://github.com/microsoft/sarif-visualstudio-extension/pull/278#discussion_r527233531) [](ancestors = 527233531,527217732)

IMO, this is not required, but a preferred data type to use. Thanks for taking this seriously. 👍

xiangyuMS (approved):

src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
lgolding: I always take PR feedback seriously :-) So let me think out load for a moment.

Suppose it turns out that the dictionary accesses need to be encapsulated in a higher-level lock (for example, locking the entire body of `AddInfoBarIfRequiredAsync`, or locking an entire _sequence_ of calls to that method, which is what I'm currently doing. In that case, is there any benefit to using `ConcurrentDictionary`? It would seem that you incur the perf overhead of a concurrent collection without getting any additional safety back.

---
In reply to: [527238898](https://github.com/microsoft/sarif-visualstudio-extension/pull/278#discussion_r527238898) [](ancestors = 527238898)

xiangyuMS: You do come with a good point about  two options here, i.e., using ConcurrentDictionary vs executing locks, and they can bring different performance metrics. And this is a VS extension, which cares more about user experience than some server services.

My original thought is since this is a multi-thread environment, why not using a safer data type to avoid issue caused by a  later change, which slips out of the lock.

So, I agree that the current implementation doesn't need ConcurrentDictionary<> type. For potential perf issues, we don't have to change.  :)

Looks this is a very useful discussion.

lgolding: Ok, you've convinced me. Perf is not an issue here, because we are

(a) performing a small number of operations (there are only a few info bars), and
(b) performing them on a human time scale (whenever a build finishes, for example).

So it comes down to safety, and you are right: On the day somebody performs a dictionary operation _outside_ of the lock -- wrong though that may be -- we will find ourselves with a hard to diagnose bug.

So I will make the change.

---
In reply to: [527253972](https://github.com/microsoft/sarif-visualstudio-extension/pull/278#discussion_r527253972) [](ancestors = 527253972)

xiangyuMS: > Ok, you've convinced me. Perf is not an issue here, because we are
> 
> (a) performing a small number of operations (there are only a few info bars), and
> (b) performing them on a human time scale (whenever a build finishes, for example).
> 
> So it comes down to safety, and you are right: On the day somebody performs a dictionary operation _outside_ of the lock -- wrong though that may be -- we will find ourselves with a hard to diagnose bug.
> 
> So I will make the change.
> 
> In reply to: [527253972](https://github.com/microsoft/sarif-visualstudio-extension/pull/278#discussion_r527253972) [](ancestors = 527253972)

👍 Good talk! Thanks.

lgolding: Fixed.

---
In reply to: [527263211](https://github.com/microsoft/sarif-visualstudio-extension/pull/278#discussion_r527263211) [](ancestors = 527263211)

src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
lgolding: Ok, I tried coding it the other way and it actually makes the code a bit more complicated. So I've done the due diligence and I'm now just going to add the comment that explains why it is as it is.

---
In reply to: [527229642](https://github.com/microsoft/sarif-visualstudio-extension/pull/278#discussion_r527229642) [](ancestors = 527229642,527210678)

lgolding: Done.

---
In reply to: [527281811](https://github.com/microsoft/sarif-visualstudio-extension/pull/278#discussion_r527281811) [](ancestors = 527281811,527229642,527210678)

src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
```diff
@@ -197,5 +223,21 @@ public void OnActionItemClicked(IVsInfoBarUIElement infoBarUIElement, IVsInfoBar
...
+            string message,
+            ImageMoniker imageMoniker)
+        {
+            // It's safe to access and update the dictionary because this method is called inside
+            // the write lock.
+            if ((detectedConditions & individualCondition) == individualCondition && !s_infoBarToConditionDictionary.Values.Contains(individualCondition))
+            {
+                var infoBar = new InfoBar(message, imageMoniker: imageMoniker);
+                await infoBar.ShowAsync();
+                s_infoBarToConditionDictionary.Add(infoBar, individualCondition);
```
GabeDeBacker, line 239: What's the implications of the ordering of these statements? Does it matter?

lgolding: It matters a tiny bit, and to the extent that it does, I have it wrong. If somehow the users closed the InfoBar that was just shown _before_ it was added to the list of tracked InfoBars (between lines 238 and 239), then no `InfoBar` for this `individualCondition` would every be shown again -- because every time we tried to show one, we'd see that it was already in the dictionary, and there'd be no way to remove it because it was already closed.

So I'll reverse 238 and 239.

---
In reply to: [527327149](https://github.com/microsoft/sarif-visualstudio-extension/pull/278#discussion_r527327149) [](ancestors = 527327149)

src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
```diff
@@ -45,32 +47,52 @@ public class InfoBar : IVsInfoBarUIEvents
...
-                new InfoBar(Resources.ErrorLogHasErrorLevelToolConfigurationNotifications).ShowAsync().FileAndForget(FileAndForgetEventName.InfoBarOpenFailure);
-            }
-
-            if ((conditions & ExceptionalConditions.ExecutionError) != 0)
-            {
-                new InfoBar(Resources.ErrorLogHasErrorLevelToolExecutionNotifications).ShowAsync().FileAndForget(FileAndForgetEventName.InfoBarOpenFailure);
-            }
-
-            if ((conditions & ExceptionalConditions.NoResults) != 0)
+            using (s_infoBarLock.EnterWriteLock())
```
GabeDeBacker, line 66: >s_infoBarLock.EnterWriteLock() [](start = 19, length = 30)

So this is really holding a lock while calling into "ShowAsync" below.. Any concerns at all about that? (Like ShowAsync taking a while, or even worse, ending up back here?)

lgolding: Yes, and I'll change that. See my chat with Xiang Yu -- the sequence of four calls to `AddInfoBarIfRequiredAsync` doesn't have to be atomic. I only need to hold the lock _within_ each call, to atomicize the logic that determines that I need to access the dictionary and the dictionary access itself. I'll move the `ShowAsync` call outside the lock. Good catch, thanks for upping my `async` game.

---
In reply to: [527329760](https://github.com/microsoft/sarif-visualstudio-extension/pull/278#discussion_r527329760) [](ancestors = 527329760)

src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
xiangyuMS, line 239: Refer this [link ](https://docs.microsoft.com/en-us/dotnet/standard/collections/thread-safe/how-to-add-and-remove-items)for ConcurrentDictionary<>, TryAdd() is prefered:
if ( !s_infoBarToConditionDictionary.Add(xxxx, xxxx))
{
      throw new Exception("");
}

GabeDeBacker (approved):
:shipit:

src/Sarif.Viewer.VisualStudio/Controls/InfoBar.cs
lgolding: Fixed, thanks.

---
In reply to: [527332697](https://github.com/microsoft/sarif-visualstudio-extension/pull/278#discussion_r527332697) [](ancestors = 527332697)

PR #281, opened 2020-11-20
#114: Preliminary UI with info bars

lgolding:
Also:
- Standardize the names of the extension method files (`.Extensions.cs` instead of `.extensions.cs`).


lgolding:
These changes are merged into the feature branch `feature\spam`.

PR #284, opened 2020-11-23
Fix #283: Closing a source file doesn't remove background analysis error list items

### LINKED ISSUE #283: Closing a source file doesn't remove background analysis error list items
That's because `BackgroundAnalyzerBase.StartAnalysisAsync` does not provide a `logId` as a second argument to `IBackgroundAnalysisSync.ReceiveAsync`. As a result, the `ErrorListService` does not make an entry for this file in `RunIndexToDataCached`.

lgolding:
The bug was that `BackgroundAnalyzerBase.StartAnalysisAsync` did not provide a `logId` as a second argument to `IBackgroundAnalysisSync.ReceiveAsync`. As a result, the `ErrorListService` did not make an entry for this file in `RunIndexToDataCached`, so they were not removed when the file was closed.

Also:
- Fix casing in the the names of the SARIF extension method files.


src/Sarif.Sarifer/IBackgroundAnalysisSink.cs
```diff
@@ -25,7 +25,9 @@ public interface IBackgroundAnalysisSink
         /// <returns>
```
eddynaka, line 25: should we add the new param to the docs?

eddynaka (approved):
added a minor comment. Besides that, LGTM

src/Sarif.Sarifer/IBackgroundAnalysisSink.cs
lgolding: Sorry, I put it in the wrong place :-) See below. I will fix it.

---
In reply to: [528859565](https://github.com/microsoft/sarif-visualstudio-extension/pull/284#discussion_r528859565) [](ancestors = 528859565)

Issue #282, opened 2020-11-23
Background analysis: Pattern matching


lgolding:
Items marked with __*__ explicitly mentioned by @michaelcfanning; other requirements inferred:

- [x] Implement right-click "Analyze solution" menu item:
    - [x] Clear all background-analysis-generated items.
    - [x] Stop any running per-solution analysis.
    - [x] Stop all running per-project analyses.
    - [x] __*__ Read `.spam` folder.
    - [x] __*__ Start new per-solution analysis.
- [x] On completion of per-solution analysis:
    - [x] Remove existing per-solution error list entries.
    - [x] Remove existing per-project error list entries for all projects.
    - [x] Insert new per-solution error list entries.
- [x] Implement right-click "Analyze project" menu item:
    - [x] Clear all background-analysis-generated items.
    - [x] Stop per-project analysis for this project.
    - [x] __*__ Read `.spam` folder.
    - [x] __*__ Start new per-project analysis.
- [x] On completion of per-project analysis:
    - [x] Remove existing per-project error list entries for this project.
    - [x] Insert new per-project error list entries.
- [x] When user closes solution:
    - [x] Stop per-project analyses.
    - [x] Remove per-project error list items.
    - [x] Stop per-solution analysis.
    - [x] Remove per-solution error list items.
- [x] __*__ Do not duplicate Error List items between per-project, per-solution, and individually opened files.
- [x] Do not analyze an individually opened file while a solution-wide analysis is in flight.

eddynaka:
Hi @michaelcfanning / @lgolding , this issue is complete.
Can you download the main branch and test it?

Thank you :)

PR #285, opened 2020-11-24
Implement analyze project/solution gestures

lgolding:
@eddynaka, this is ready for you to look at. It contains the first cut at project and solution analyses. It does _not_ yet do any of the following:

- Detect duplicate results.
- Stop one background analysis if you start another.
- Removing results when solution is closed.

See #282 for a detailed list. Note that exactly two items are checked off:

- [x] Start new per-solution analysis.
- [x] Start new per-project analysis.

Also:
- I removed the incorrect line/column calculation. We can add your correction back in a subsequent PR (or even in this PR, if you like), but I'd like to talk to you about caching first. Since the viewer does the caching already, I'm inclined not to repeat it in Sarifer.


src/Sarif.Sarifer/BackgroundAnalysisService.cs
```diff
@@ -24,13 +24,24 @@ internal class BackgroundAnalysisService : IBackgroundAnalysisService
 #pragma warning restore CS0649
 
         /// <inheritdoc/>
-        public async Task StartAnalysisAsync(string path, string text)
+        public async Task AnalyzeAsync(string path, string text)
         {
-            List<Task> tasks = new List<Task>(analyzers.Count());
+            var tasks = new List<Task>(analyzers.Count());
```
eddynaka, line 29: should we remove or use the same idea as the method below?

eddynaka (approved):

src/Sarif.Sarifer/BackgroundAnalysisService.cs
lgolding: I should have used that idea. Thanks.

---
In reply to: [531726321](https://github.com/microsoft/sarif-visualstudio-extension/pull/285#discussion_r531726321) [](ancestors = 531726321)

PR #287, opened 2020-11-25
Allow exceptions to escape from task status center so tests can see them

lgolding:
In the file-based API code path, the log file handling was being done in a background task in the Task Status Center. If that background task threw an exception, it was lost. The affected test would fail with an verification failure, but the stack trace did not reveal the exception that caused the failure.

In this PR, we refactor to assign the background task to a temporary variable, pass the task variable to the Task Status Center, and then await the task ourselves. Now if the task throws an exception, the viewer will rethrow it, and Apex will detect and report it.

I've tested this by intentionally introducing an exception into the error list service's log file handling and watching Apex report it.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
+++ b/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
@@ -77,7 +77,7 @@ namespace Microsoft.Sarif.Viewer.ErrorList
             Task task = ProcessLogFileAsync(filePath, toolFormat, promptOnLogConversions, cleanErrors, openInEditor);
             taskHandler.RegisterTask(task);
 
-            await task;
+            await task.ConfigureAwait(continueOnCapturedContext: false);
         }
 
         public static async Task ProcessLogFileAsync(string filePath, string toolFormat, bool promptOnLogConversions, bool cleanErrors, bool openInEditor)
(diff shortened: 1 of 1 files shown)


GabeDeBacker (approved):
:shipit:

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -69,7 +74,10 @@ public static void ProcessLogFile(string filePath, string toolFormat, bool promp
 
             ITaskHandler taskHandler = taskStatusCenterService.PreRegister(taskHandlerOptions, taskProgressData);
 
-            taskHandler.RegisterTask(ProcessLogFileAsync(filePath, toolFormat, promptOnLogConversions, cleanErrors, openInEditor));
+            Task task = ProcessLogFileAsync(filePath, toolFormat, promptOnLogConversions, cleanErrors, openInEditor);
+            taskHandler.RegisterTask(task);
+
+            await task;
```
eddynaka, line 80: Should we ConfigureAwait?

eddynaka (approved):

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
GabeDeBacker: Well, in reality you can just return the task and remove the "async" and "await" :)

But then again, I don't know what the JoinableTaskFactory.Run is going to do in that case.

---
In reply to: [530692825](https://github.com/microsoft/sarif-visualstudio-extension/pull/287#discussion_r530692825) [](ancestors = 530692825)

lgolding: I don't know why the threading analyzers didn't report this. It might be that if the `await` is the last statement in the function, it doesn't actually matter. In any case, I added `ConfigureAwait(false)`.

---
In reply to: [530694783](https://github.com/microsoft/sarif-visualstudio-extension/pull/287#discussion_r530694783) [](ancestors = 530694783,530692825)

PR #291, opened 2020-11-28
Enable cancellation token

eddynaka:


src/Sarif.Sarifer/AnalyzeProjectCommand.cs
```diff
@@ -15,14 +16,17 @@
...
 namespace Microsoft.CodeAnalysis.Sarif.Sarifer
 {
-    internal class AnalyzeProjectCommand
+    internal class AnalyzeProjectCommand : IDisposable
     {
         private DTE2 dte;
         private IComponentModel componentModel;
         private IBackgroundAnalysisService backgroundAnalysisService;
+        private readonly CancellationTokenSource cancellationTokenSource;
+        private bool disposedValue;
```
michaelcfanning, line 25: >disposedValue [](start = 21, length = 13)

suggest just using 'disposed' for this variable or 'isDisposed' if that seems clearer to you.

src/Sarif.Sarifer/AnalyzeSolutionCommand.cs
```diff
@@ -69,7 +73,27 @@ private void MenuCommandCallback(object caller, EventArgs args)
                 targetFiles.AddRange(project.GetMemberFiles());
             }
 
-            this.backgroundAnalysisService.AnalyzeAsync(solution.FullName, targetFiles).FileAndForget(FileAndForgetEventName.BackgroundAnalysisFailure);
+            this.backgroundAnalysisService.AnalyzeAsync(solution.FullName, targetFiles, cancellationTokenSource.Token)
```
michaelcfanning, line 76: >cancellationTokenSource [](start = 88, length = 23)

we should enforce a standard in this code of using this. prefix, yes?

eddynaka: yes! just updated the sarifer project!

---
In reply to: [532104079](https://github.com/microsoft/sarif-visualstudio-extension/pull/291#discussion_r532104079) [](ancestors = 532104079)

src/Sarif.Sarifer/AnalyzeProjectCommand.cs
eddynaka: good idea. just changed to `disposed` in all places that i implemented `IDisposable`

---
In reply to: [532103867](https://github.com/microsoft/sarif-visualstudio-extension/pull/291#discussion_r532103867) [](ancestors = 532103867)

michaelcfanning (approved):
:shipit:

PR #298, opened 2020-11-30
Run background analyzers when a solution is opened.

lgolding:
Based on the sample https://github.com/microsoft/VSSDK-Extensibility-Samples/tree/master/SolutionLoadEvents

UI tests in progress.

### DIFF
--- a/src/Sarif.Sarifer/SariferPackage.cs
+++ b/src/Sarif.Sarifer/SariferPackage.cs
@@ -32,7 +32,6 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer
     public sealed class SariferPackage : AsyncPackage, IDisposable
     {
         public const string PackageGuidString = "F70132AB-4095-477F-AAD2-81D3D581113B";
-        public static readonly Guid PackageGuid = new Guid(PackageGuidString);
         private bool disposed;
         private AnalyzeSolutionCommand analyzeSolutionCommand;
         private AnalyzeProjectCommand analyzeProjectCommand;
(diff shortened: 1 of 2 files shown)


src/Sarif.Sarifer/SariferPackage.cs
```diff
@@ -93,5 +110,24 @@ public void Dispose()
...
+
+            IVsSolution solutionService = await this.GetServiceAsync(typeof(SVsSolution)).ConfigureAwait(continueOnCapturedContext: true) as IVsSolution
+                ?? throw new InvalidOperationException("Cannot load solution service.");
+
+            ErrorHandler.ThrowOnFailure(solutionService.GetProperty((int)__VSPROPID.VSPROPID_IsSolutionOpen, out object value));
+
+            return value is bool isSolutionOpen && isSolutionOpen;
+        }
+
+        private void OnSolutionLoaded(object sender = null, EventArgs e = null)
```
michaelcfanning, line 126: >OnSolutionLoaded [](start = 21, length = 16)

What's going to happen here? I didn't ask for a background analysis on solution load. This could greatly slow things down after SLN load. Instead, the analyzer should inspect all opened files only, unless/until the user invokes a specific gesture to analyze the entire solution.

lgolding: Ok, I'll withdraw this.

---
In reply to: [532939181](https://github.com/microsoft/sarif-visualstudio-extension/pull/298#discussion_r532939181) [](ancestors = 532939181)

lgolding:
Not a functional requirement, and potentially slows down solution load, so abandon.

michaelcfanning:
Maybe we should be checking in together on next work to complete, so that we don't spend time on things we aren't sure we have concensus on?

PR #301, opened 2020-12-01
Enable background analyzer

eddynaka:


lgolding (approved):
:shipit:

src/Sarif.Sarifer/SpamBackgroundAnalyzer.cs
```diff
@@ -0,0 +1,88 @@
...
+using Newtonsoft.Json;
+
+namespace Microsoft.CodeAnalysis.Sarif.Sarifer
+{
+    [Export(typeof(IBackgroundAnalyzer))]
+    internal class SpamBackgroundAnalyzer : BackgroundAnalyzerBase
+    {
+        private readonly IFileSystem fileSystem;
+        private string currentSolutionDirectory;
+        private ISet<Skimmer<AnalyzeContext>> rules;
```
lgolding, line 24: >AnalyzeContext [](start = 29, length = 14)

If we're no longer using `SpamRule` we can delete it.

src/Sarif.Sarifer/SpamBackgroundAnalyzer.cs
```diff
@@ -0,0 +1,88 @@
...
+                AnalyzeCommand.AnalyzeTargetHelper(context, this.rules, disabledSkimmers);
+            }
+        }
+
+        internal static ISet<Skimmer<AnalyzeContext>> LoadSearchDefinitionsFiles(IFileSystem fileSystem, string solutionDirectory)
+        {
+            string spamDirectory = Path.Combine(solutionDirectory, ".spam");
+            if (!fileSystem.DirectoryExists(spamDirectory))
+            {
+                return new HashSet<Skimmer<AnalyzeContext>>();
```
lgolding, line 76: >HashSet [](start = 27, length = 7)

The reason for the `currentRules` logic was that when you analyze a solution or a project, `AnalyzeCore` will be called once per source file, and we didn't want to reread the rules for each file.

lgolding: That would have been worth a comment.

---
In reply to: [534509623](https://github.com/microsoft/sarif-visualstudio-extension/pull/301#discussion_r534509623) [](ancestors = 534509623)

src/Sarif.Sarifer.UnitTests/SpamBackgroundAnalyzerTests.cs
```diff
@@ -0,0 +1,106 @@
...
+
+        [Fact]
+        public void LoadPatternFiles_WhenDirectoryDoesExistWithFiles_ShouldReturnRules()
+        {
+            const string SpamDirectory = ".spam";
+            const string ProjectDirectory = @"C:\some-project-folder";
+
+            var definitions = new SearchDefinitions()
+            {
+                Definitions = new List<SearchDefinition>(new[]
```
lgolding, line 63: >new[] [](start = 57, length = 5)

Not necessary; this suffices:
```
new List<T>
{
    new T
    {
    },
    ...
};
```

lgolding:
>         public RunDataCache(Run run, int runIndex, string logFilePath)

Remove from signature if no longer used.

---
Refers to: src/Sarif.Viewer.VisualStudio/RunDataCache.cs:50 in 932501e. [](commit_id = 932501e3163134b9400d4575412ab495f7d21da6, deletion_comment = False)

src/Sarif.Sarifer/SpamBackgroundAnalyzer.cs
eddynaka: if you check, this method is called in `AnalyzeCore`, so the rules will be created once unless we change the solutionDirectory.

---
In reply to: [534509781](https://github.com/microsoft/sarif-visualstudio-extension/pull/301#discussion_r534509781) [](ancestors = 534509781,534509623)

lgolding: As you pointed out off line: The logic in `AnalyzeCore` prevents this from happening on every file:
```
if (string.IsNullOrEmpty(solutionDirectory)
                || (this.currentSolutionDirectory?.Equals(solutionDirectory, StringComparison.OrdinalIgnoreCase) != true))
```


---
In reply to: [534518965](https://github.com/microsoft/sarif-visualstudio-extension/pull/301#discussion_r534518965) [](ancestors = 534518965,534509781,534509623)

azure-pipelines.yml
```diff
@@ -5,14 +5,16 @@ pool:
...
     modifyEnvironment: true
 
-- task: PowerShell@2
-  displayName: Build and Test
+- task: BatchScript@1
+  displayName: 'Run BuildAndTest.cmd'
   inputs:
-    targetType: filePath
-    filePath: ./scripts/BuildAndTest.ps1
+    filename: 'BuildAndTest.cmd'
```
lgolding, line 20: >cmd [](start = 28, length = 3)

Just curious why you chose not to invoke the PowerShell script directly?

scripts/BuildAndTest.ps1
```diff
@@ -144,6 +144,12 @@ if (-not $NoRestore) {
 }
 
 if (-not $NoBuild) {
+    & $RepoRoot\src\sarif-pattern-matcher\BuildAndTest.cmd -NoTest
```
lgolding, line 147: >& $RepoRoot\src\sarif-pattern-matcher\BuildAndTest.cmd -NoTest [](start = 4, length = 62)

Check the exit code from each command individually.

azure-pipelines.yml
eddynaka: i thought that was the issue. i will revert and test again :)

---
In reply to: [534552440](https://github.com/microsoft/sarif-visualstudio-extension/pull/301#discussion_r534552440) [](ancestors = 534552440)

scripts/BuildAndTest.ps1
```diff
@@ -144,6 +144,12 @@ if (-not $NoRestore) {
...
+    & $RepoRoot\src\sarif-pattern-matcher\Src\sarif-sdk\BuildAndTest.cmd -NoTest
```
lgolding, line 148: >sarif-sdk [](start = 46, length = 9)

If the SDK is a submodule of Spam, and if Spam actually consumes the SDK binaries (as opposed to consuming a NuGet package), then don't you have to build the SDK first?

And if Spam does _not_ consume SDK binaries, then why do you have to build the SDK at all? For that matter, if that is the case, why are we including the SDK as a submodule at all?

eddynaka: I will have to take a look at the spam build and test. i think we can simplify it.
Let's merge for now, so we can do some work, and i will come back to this later (today or in the next few days)

---
In reply to: [534553739](https://github.com/microsoft/sarif-visualstudio-extension/pull/301#discussion_r534553739) [](ancestors = 534553739)

lgolding: Ok, but before merging, please do make that one last change of checking the exit code from each of the two builds separately. Then yes, check whether they are in the right order, and if the SDK build is needed at all.

---
In reply to: [534555694](https://github.com/microsoft/sarif-visualstudio-extension/pull/301#discussion_r534555694) [](ancestors = 534555694,534553739)

lgolding: Oh I see you already did that :-) ok.

---
In reply to: [534556847](https://github.com/microsoft/sarif-visualstudio-extension/pull/301#discussion_r534556847) [](ancestors = 534556847,534555694,534553739)

PR #302, opened 2020-12-01
Cancel before analyze

eddynaka:
Refs #282

## Changes
- Implement right-click "Analyze solution" menu item:
  - Stop any running per-solution analysis.
- Implement right-click "Analyze project" menu item:
  -  Stop any running per-project analysis.
 - When user closes solution:
   - Stop per-project analyses.
   - Stop per-solution analysis.

### DIFF
--- a/src/Sarif.Sarifer/AnalyzeProjectCommand.cs
+++ b/src/Sarif.Sarifer/AnalyzeProjectCommand.cs
@@ -21,12 +21,11 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer
         private DTE2 dte;
         private IComponentModel componentModel;
         private IBackgroundAnalysisService backgroundAnalysisService;
-        private readonly CancellationTokenSource cancellationTokenSource;
+        private CancellationTokenSource cancellationTokenSource;
         private bool disposed;
 
         public AnalyzeProjectCommand(IMenuCommandService menuCommandService)
         {
-            this.cancellationTokenSource = new CancellationTokenSource();
             var menuCommand = new MenuCommand(
                 new EventHandler(this.MenuCommandCallback),
                 new CommandID(Guids.SariferCommandSet, SariferPackageCommandIds.AnalyzeProject));
@@ -40,6 +39,10 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer
         private void MenuCommandCallback(object caller, EventArgs args)
         {
             ThreadHelper.ThrowIfNotOnUIThread();
+
+            // always cancel before start
+            this.cancellationTokenSource?.Cancel();
+            this.cancellationTokenSource = new CancellationTokenSource();
 
             if (this.dte == null)
             {
@@ -75,7 +78,8 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer
             {
                 if (disposing)
                 {
-                    this.cancellationTokenSource.Dispose();
+                    this.cancellationTokenSource?.Cancel();
+                    this.cancellationTokenSource?.Dispose();
                 }
 
                 this.disposed = true;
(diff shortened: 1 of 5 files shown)


src/Sarif.Sarifer/AnalyzeSolutionCommand.cs
```diff
@@ -40,6 +39,10 @@ private void MenuCommandCallback(object caller, EventArgs args)
         {
             ThreadHelper.ThrowIfNotOnUIThread();
 
+            // always cancel before start
+            this.cancellationTokenSource?.Cancel();
```
lgolding, line 43: >? [](start = 40, length = 1)

Do we need the null conditional operator? `cancellationTokenSource` was initialized in the ctor.

src/Sarif.Sarifer/AnalyzeProjectCommand.cs
```diff
@@ -41,6 +40,10 @@ private void MenuCommandCallback(object caller, EventArgs args)
         {
             ThreadHelper.ThrowIfNotOnUIThread();
 
+            // always cancel before start
```
lgolding, line 43: >a [](start = 15, length = 1)

"Always ... start." (Capitalized, ends with '.')

src/Sarif.Sarifer/AnalyzeSolutionCommand.cs
eddynaka: I removed from the ctor, since we always have to start a new one each time.

---
In reply to: [533004150](https://github.com/microsoft/sarif-visualstudio-extension/pull/302#discussion_r533004150) [](ancestors = 533004150)

lgolding: Oh, ok, I see it now, thanks.

---
In reply to: [533004913](https://github.com/microsoft/sarif-visualstudio-extension/pull/302#discussion_r533004913) [](ancestors = 533004913,533004150)

src/Sarif.Sarifer/AnalyzeProjectCommand.cs
```diff
@@ -41,6 +40,10 @@ private void MenuCommandCallback(object caller, EventArgs args)
...
+            this.cancellationTokenSource?.Cancel();
+            this.cancellationTokenSource = new CancellationTokenSource();
+
```
lgolding, line 46: @michaelcfanning Suppose you start analyses on two different projects (with two separate gestures). If you cancel one of them (on the day we provide a "Cancel" gesture), do you want to cancel both of them?

If not, we need a mapping from project files to cancellation token sources. We can't get by with just one source.

lgolding (changes requested):
:clock1:

lgolding:
Two issues:

1) See my comment about whether we need one cancellation token source per project.

2) Do we want to handle cancellation of per-file analyses in this PR? (I would be willing to have it in a separate PR.) For single-file analyses, _either_ (a) the TextViewCreationListener would need to track a separate cancellation token per file, or (b) we have to agree that a single gesture cancels all in-progress single-file analyses.

If we choose (a), we need to think about where that gesture lives in the UI. IMO it's a harder problem than per-project analyses (because it's kind of obvious that the "Cancel" gesture for the project or the solution can go in the same context menu as the "Start analysis" gesture).

src/Sarif.Sarifer/AnalyzeProjectCommand.cs
michaelcfanning: I don't see the value in having the ability to initiate two asynchronous analyses. You should think of this like a build gesture, you can build the SLN, or you can build a single project. Whatever you choose, you can't initiate another build operation until the first one is completed or canceled. Our feature should be the same. 


---
In reply to: [533006458](https://github.com/microsoft/sarif-visualstudio-extension/pull/302#discussion_r533006458) [](ancestors = 533006458)

eddynaka: Based on this, this PR is still valid. If we press `analyze` in any project and press it again, it would cancel the first one and maintain the second. In the other hand, still checking how we will be able to cancel everything when we do this for sln level or when we close the sln

---
In reply to: [533014515](https://github.com/microsoft/sarif-visualstudio-extension/pull/302#discussion_r533014515) [](ancestors = 533014515,533006458)

PR #310, opened 2020-12-03
Clear results before analyze + fixes

eddynaka:
## Changes:
- before solution analysis, we will clear all results
- before project analysis, we will clear all results
- if we close the solution, we will clear all results
- if we analyze a single file, it will clear just that file result
- fixed possible null reference if we analyze a single file and that file is a stand-alone file, which would generate the information bar showing that no results were found

## Minor changes:
- fixed a few message and warning issues

### DIFF
--- a/src/Sarif.Sarifer/AnalyzeProjectCommand.cs
+++ b/src/Sarif.Sarifer/AnalyzeProjectCommand.cs
@@ -38,6 +38,8 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer
         /// </summary>
         public void Cancel()
         {
+            this.backgroundAnalysisService.ClearAsync()
+                .FileAndForget(FileAndForgetEventName.BackgroundAnalysisFailure);
             this.cancellationTokenSource?.Cancel();
         }
 
--- a/src/Sarif.Sarifer/AnalyzeProjectCommand.cs
+++ b/src/Sarif.Sarifer/AnalyzeProjectCommand.cs
@@ -38,7 +38,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer
         /// </summary>
         public void Cancel()
         {
-            this.backgroundAnalysisService.ClearAsync()
+            this.backgroundAnalysisService.ClearResultsAsync()
                 .FileAndForget(FileAndForgetEventName.BackgroundAnalysisFailure);
             this.cancellationTokenSource?.Cancel();
         }
--- a/src/Sarif.Sarifer/AnalyzeProjectCommand.cs
+++ b/src/Sarif.Sarifer/AnalyzeProjectCommand.cs
@@ -38,7 +38,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer
         /// </summary>
         public void Cancel()
         {
-            this.backgroundAnalysisService.ClearResultsAsync()
+            this.backgroundAnalysisService?.ClearResultsAsync()
                 .FileAndForget(FileAndForgetEventName.BackgroundAnalysisFailure);
             this.cancellationTokenSource?.Cancel();
         }
--- a/src/Sarif.Sarifer/BackgroundAnalysisService.cs
+++ b/src/Sarif.Sarifer/BackgroundAnalysisService.cs
@@ -17,7 +17,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer
     [Export(typeof(IBackgroundAnalysisService))]
     internal class BackgroundAnalysisService : IBackgroundAnalysisService
     {
-#pragma warning disable IDE0044, CS0649 // Assigned,Filled in by MEF
+#pragma warning disable IDE0044, CS0649 // Provided by MEF
         [ImportMany]
         private IEnumerable<IBackgroundAnalyzer> analyzers;
 #pragma warning restore IDE0044, CS0649
--- a/src/Sarif.Sarifer/BackgroundAnalysisService.cs
+++ b/src/Sarif.Sarifer/BackgroundAnalysisService.cs
@@ -3,6 +3,7 @@
 
 using System.Collections.Generic;
 using System.ComponentModel.Composition;
+using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
@@ -20,32 +21,39 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer
 #pragma warning disable IDE0044, CS0649 // Provided by MEF
         [ImportMany]
         private IEnumerable<IBackgroundAnalyzer> analyzers;
+
+        [ImportMany]
+        private IEnumerable<IBackgroundAnalysisSink> sinks;
 #pragma warning restore IDE0044, CS0649
 
         /// <inheritdoc/>
         public async Task AnalyzeAsync(string path, string text, CancellationToken cancellationToken)
         {
-            var tasks = new List<Task>(this.analyzers.Count());
+            var tasks = new List<Task<Stream>>(this.analyzers.Count());
 
             foreach (IBackgroundAnalyzer analyzer in this.analyzers)
             {
                 tasks.Add(analyzer.AnalyzeAsync(path, text, cancellationToken));
             }
 
-            await Task.WhenAll(tasks).ConfigureAwait(continueOnCapturedContext: false);
+            Stream[] streams = await Task.WhenAll(tasks).ConfigureAwait(continueOnCapturedContext: false);
+            await this.WriteStreamsToSinksAsync(path, streams, cleanAll: false).ConfigureAwait(continueOnCapturedContext: false);
+            DisposeStreams(streams);
         }
 
         /// <inheritdoc/>
         public async Task AnalyzeAsync(string logId, IEnumerable<string> targetFiles, CancellationToken cancellationToken)
         {
-            var tasks = new List<Task>(this.analyzers.Count());
+            var tasks = new List<Task<Stream>>(this.analyzers.Count());
 
             foreach (IBackgroundAnalyzer analyzer in this.analyzers)
             {
                 tasks.Add(analyzer.AnalyzeAsync(logId, targetFiles, cancellationToken));
             }
 
-            await Task.WhenAll(tasks).ConfigureAwait(continueOnCapturedContext: false);
+            Stream[] streams = await Task.WhenAll(tasks).ConfigureAwait(continueOnCapturedContext: false);
+            await this.WriteStreamsToSinksAsync(logId, streams, cleanAll: true).ConfigureAwait(continueOnCapturedContext: false);
+            DisposeStreams(streams);
         }
 
         /// <inheritdoc/>
@@ -53,12 +61,50 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer
         {
             var tasks = new List<Task>(this.analyzers.Count());
 
-            foreach (IBackgroundAnalyzer analyzer in this.analyzers)
+            foreach (IBackgroundAnalysisSink sink in this.sinks)
             {
-                tasks.Add(analyzer.ClearResultsAsync());
+                tasks.Add(sink.CloseAsync());
             }
 
             await Task.WhenAll(tasks).ConfigureAwait(continueOnCapturedContext: false);
+        }
+
+        private Task WriteStreamsToSinksAsync(string logId, Stream[] streams, bool cleanAll)
+        {
+            var sinkTasks = new List<Task>(this.analyzers.Count());
+            foreach (Stream stream in streams)
+            {
+                sinkTasks.Add(this.WriteToSinksAsync(logId, stream, cleanAll));
+            }
+
+            return Task.WhenAll(sinkTasks);
+        }
+
+        private async Task WriteToSinksAsync(string logId, Stream stream, bool cleanAll)
+        {
+            foreach (IBackgroundAnalysisSink sink in this.sinks)
+            {
+                stream.Seek(0L, SeekOrigin.Begin);
+
+                if (cleanAll)
+                {
+                    await sink.CloseAsync().ConfigureAwait(false);
+                }
+                else
+                {
+                    await sink.CloseAsync(new[] { logId }).ConfigureAwait(false);
+                }
+
+                await sink.ReceiveAsync(stream, logId).ConfigureAwait(false);
+            }
+        }
+
+        private static void DisposeStreams(Stream[] streams)
+        {
+            foreach (Stream stream in streams)
+            {
+                stream.Dispose();
+            }
         }
     }
 }
--- a/src/Sarif.Sarifer/BackgroundAnalysisService.cs
+++ b/src/Sarif.Sarifer/BackgroundAnalysisService.cs
@@ -49,13 +49,13 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer
         }
 
         /// <inheritdoc/>
-        public async Task ClearAsync()
+        public async Task ClearResultsAsync()
         {
             var tasks = new List<Task>(this.analyzers.Count());
 
             foreach (IBackgroundAnalyzer analyzer in this.analyzers)
             {
-                tasks.Add(analyzer.ClearAsync());
+                tasks.Add(analyzer.ClearResultsAsync());
             }
 
             await Task.WhenAll(tasks).ConfigureAwait(continueOnCapturedContext: false);
--- a/src/Sarif.Sarifer/BackgroundAnalysisService.cs
+++ b/src/Sarif.Sarifer/BackgroundAnalysisService.cs
@@ -55,7 +55,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer
 
             foreach (IBackgroundAnalyzer analyzer in this.analyzers)
             {
-                tasks.Add(analyzer.AnalyzeAsync(logId, targetFiles, cancellationToken));
+                tasks.Add(analyzer.AnalyzeAsync(targetFiles, cancellationToken));
             }
 
             Stream[] streams = await Task.WhenAll(tasks).ConfigureAwait(continueOnCapturedContext: false);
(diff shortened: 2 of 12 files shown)


src/Sarif.Sarifer/BackgroundAnalysisService.cs
```diff
@@ -17,12 +17,10 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer
     [Export(typeof(IBackgroundAnalysisService))]
     internal class BackgroundAnalysisService : IBackgroundAnalysisService
     {
-#pragma warning disable CS0649 // Filled in by MEF
-#pragma warning disable IDE0044 // Assigned by MEF
+#pragma warning disable IDE0044, CS0649 // Assigned,Filled in by MEF
```
lgolding, line 20: >Assigned,Filled in [](start = 43, length = 18)

How about just "Provided by MEF."

lgolding, line 20: >, [](start = 31, length = 1)

I didn't know you could do that :-)

src/Sarif.Sarifer/BackgroundAnalysisService.cs
```diff
@@ -49,5 +47,18 @@ public async Task AnalyzeAsync(string logId, IEnumerable<string> targetFiles, Ca
...
         }
+
+        /// <inheritdoc/>
+        public async Task ClearResultsAsync()
+        {
+            var tasks = new List<Task>(this.analyzers.Count());
+
+            foreach (IBackgroundAnalyzer analyzer in this.analyzers)
+            {
+                tasks.Add(analyzer.ClearResultsAsync());
```
lgolding, line 58: >ClearResultsAsync [](start = 35, length = 17)

My first design instinct is that analyzers should not be responsible for clearing results. Analyzers should be responsible for one thing: analyzing. My instinct is that the background analysis service should keep track of the results and be responsible for clearing them. But I will read further and I might change my mind.

src/Sarif.Sarifer/AnalyzeProjectCommand.cs
```diff
@@ -38,6 +38,8 @@ public AnalyzeProjectCommand(IMenuCommandService menuCommandService)
         /// </summary>
         public void Cancel()
         {
+            this.backgroundAnalysisService.ClearResultsAsync()
```
lgolding, line 41: >ClearResultsAsync [](start = 43, length = 17)

We agree that only one background analysis at a time can be in progress. But if I analyze project 1, and then I analyze project 2, don't I expect the Error List to contain the results from _both_ analyses? And if I cancel one of them, don't I expect the results from the other to remain?

This is a behavioral question so @michaelcfanning needs to decide.

src/Sarif.Sarifer/BackgroundAnalyzerBase.cs
```diff
@@ -73,7 +77,7 @@ public async Task AnalyzeAsync(string path, string text, CancellationToken cance
 
                 await writer.FlushAsync().ConfigureAwait(continueOnCapturedContext: false);
 
-                await this.WriteToSinksAsync(path, stream).ConfigureAwait(false);
+                await this.WriteToSinksAsync(path, stream, false).ConfigureAwait(false);
```
lgolding, line 80: >false [](start = 59, length = 5)

Use named parameter so it's clear what the "false" means: `cleanAll: false`.

I suggest the same for all calls to `ConfigureAwait`: `continueOnCapturedContext: false`, even though it's verbose.

src/Sarif.Sarifer/AnalyzeProjectCommand.cs
eddynaka: If we follow build commands, if we build the `sarif.sarifer` project, we will see 4 warnings. Then, if we build `string8`, we will see 0 warnings. So, in that case, each build clean the error list.

---
In reply to: [535426319](https://github.com/microsoft/sarif-visualstudio-extension/pull/310#discussion_r535426319) [](ancestors = 535426319)

lgolding: Ah, perfect. Thanks.

---
In reply to: [535435792](https://github.com/microsoft/sarif-visualstudio-extension/pull/310#discussion_r535435792) [](ancestors = 535435792,535426319)

src/Sarif.Sarifer/BackgroundAnalysisService.cs
eddynaka: yeah...its strange for me as well...the issue in this case is the sink, which is available in the `BackgroundAnalyzerBase.cs`. Based on that, what we would have to do is: analyze propagate the results to the Service and the Service would be responsible to send to sink, clean, etc. And, then, we will have some issues with the dispose of the stream. Will have to check how to work on that, since we will have to propagate it.

---
In reply to: [535424123](https://github.com/microsoft/sarif-visualstudio-extension/pull/310#discussion_r535424123) [](ancestors = 535424123)

michaelcfanning: Yes, you should remove these from the analyzer interface and implementation. The goal of this code was not to enable aggregation of multiple background analyzers, so let's be careful about building this out further. Today, it should be sufficient that the analyzers write to the SARIF log. i.e., they shouldn't retain any state/etc. that would require them to get a notification of close/clear.


---
In reply to: [535424123](https://github.com/microsoft/sarif-visualstudio-extension/pull/310#discussion_r535424123) [](ancestors = 535424123)

eddynaka: Moved the sink to the service class. Let me know what do you think!
Thank you

---
In reply to: [535442762](https://github.com/microsoft/sarif-visualstudio-extension/pull/310#discussion_r535442762) [](ancestors = 535442762,535424123)

src/Sarif.Sarifer/BackgroundAnalysisService.cs
```diff
@@ -17,37 +18,93 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer
...
-#pragma warning disable CS0649 // Filled in by MEF
-#pragma warning disable IDE0044 // Assigned by MEF
+#pragma warning disable IDE0044, CS0649 // Provided by MEF
         [ImportMany]
         private IEnumerable<IBackgroundAnalyzer> analyzers;
-#pragma warning restore IDE0044
-#pragma warning restore CS0649
+
+        [ImportMany]
+        private IEnumerable<IBackgroundAnalysisSink> sinks;
```
lgolding, line 26: >sinks [](start = 53, length = 5)

You know, I really like this. It always bothered me a little that the service imported the analyzers, but the analyzer (base) imported the sinks. This is much better.

src/Sarif.Sarifer/BackgroundAnalyzerBase.cs
```diff
@@ -50,68 +40,70 @@ public abstract class BackgroundAnalyzerBase : IBackgroundAnalyzer
...
-                    sarifLogger.AnalysisStarted();
-
-                    // TODO: What do we do when path is null (text buffer with no backing file)?
-                    Uri uri = path != null ? new Uri(path, UriKind.Absolute) : null;
+                return Stream.Null;
+            }
 
-                    this.AnalyzeCore(uri, text, solutionDirectory, sarifLogger, cancellationToken);
+            var stream = new MemoryStream();
+            var writer = new StreamWriter(stream, Encoding.UTF8);
```
lgolding, line 56: >writer [](start = 16, length = 6)

It looks like we are leaking the writer. Can we do a using with the writer? Isn't there an option on the StreamWriter ctor which says "Don't dispose the stream when you dispose the writer"?

src/Sarif.Sarifer/BackgroundAnalysisService.cs
```diff
@@ -17,37 +18,93 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer
...
-            var tasks = new List<Task>(this.analyzers.Count());
+            var tasks = new List<Task<Stream>>(this.analyzers.Count());
 
             foreach (IBackgroundAnalyzer analyzer in this.analyzers)
             {
                 tasks.Add(analyzer.AnalyzeAsync(path, text, cancellationToken));
             }
 
-            await Task.WhenAll(tasks).ConfigureAwait(continueOnCapturedContext: false);
+            Stream[] streams = await Task.WhenAll(tasks).ConfigureAwait(continueOnCapturedContext: false);
```
lgolding, line 39: This needs to be done in a try/finally so we dispose the streams if there's an exception.

In fact there are two issues:

1. If an analyzer throws, analyzer base should dispose the stream.
2. If WriteStreamsToSinks throws, you need a try/finally here to ensure disposal.

I won't think this through in detail -- I leave that to you and @michaelcfanning -- but make sure you account for all of this.

eddynaka: added a try/finally, always cleaning the streams.

---
In reply to: [535510544](https://github.com/microsoft/sarif-visualstudio-extension/pull/310#discussion_r535510544) [](ancestors = 535510544)

lgolding (approved):
:shipit:

src/Sarif.Sarifer/BackgroundAnalyzerBase.cs
```diff
@@ -72,26 +70,26 @@ public async Task AnalyzeAsync(string path, string text, CancellationToken cance
                 }
 
                 await writer.FlushAsync().ConfigureAwait(continueOnCapturedContext: false);
-
-                await this.WriteToSinksAsync(path, stream).ConfigureAwait(false);
             }
+
+            return stream;
```
lgolding, line 75: >stream [](start = 19, length = 6)

The stream can still leak if the derived class's `AnalyzeCore` throws an exception.

1. after creating the stream, wrap everything else in a try catch. In the catch block, dispose the stream and set it to null. Then the final return statement will return null.

2. In the service, you will have a list of streams, _some of which might be null_. Protect against the null ones.

src/Sarif.Sarifer/BackgroundAnalyzerBase.cs
```diff
@@ -49,68 +39,93 @@ public abstract class BackgroundAnalyzerBase : IBackgroundAnalyzer
...
+                }
 
-                await this.WriteToSinksAsync(path, stream).ConfigureAwait(false);
+                return stream;
+            }
+#pragma warning disable CA1031 // Do not catch general exception types
+            catch (Exception)
+#pragma warning restore CA1031 // Do not catch general exception types
+            {
+                return Stream.Null;
```
lgolding, line 83: >Stream [](start = 23, length = 6)

That's fine, but you still have to dispose the stream :-)

eddynaka: it will be disposed by the method that called this.

---
In reply to: [535653423](https://github.com/microsoft/sarif-visualstudio-extension/pull/310#discussion_r535653423) [](ancestors = 535653423)

src/Sarif.Sarifer/BackgroundAnalysisService.cs
```diff
@@ -17,37 +18,111 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer
...
             foreach (IBackgroundAnalyzer analyzer in this.analyzers)
             {
-                tasks.Add(analyzer.AnalyzeAsync(logId, targetFiles, cancellationToken));
+                tasks.Add(analyzer.AnalyzeAsync(targetFiles, cancellationToken));
+            }
+
+            Stream[] streams = await Task.WhenAll(tasks).ConfigureAwait(continueOnCapturedContext: false);
+            try
+            {
+                await this.WriteStreamsToSinksAsync(logId, streams, cleanAll: true).ConfigureAwait(continueOnCapturedContext: false);
```
lgolding, line 64: >streams [](start = 59, length = 7)

Now some of them might be null, so perhaps

```
Stream streams = await Task.WhenAll(tasks).ConfigureAwait(continueOnCapturedContext: false).Where(s => s != Stream.Null).ToArray();
```

src/Sarif.Sarifer/BackgroundAnalyzerBase.cs
eddynaka: added a try catch, in the catch i will return a stream.null, which will be disposed in the service

---
In reply to: [535637891](https://github.com/microsoft/sarif-visualstudio-extension/pull/310#discussion_r535637891) [](ancestors = 535637891)

src/Sarif.Sarifer/BackgroundAnalyzerBase.cs
lgolding: No, the caller will never see this stream, because you returned `null`.

---
In reply to: [535655053](https://github.com/microsoft/sarif-visualstudio-extension/pull/310#discussion_r535655053) [](ancestors = 535655053,535653423)

src/Sarif.Sarifer/BackgroundAnalysisService.cs
eddynaka: in the method `WriteStreamsToSinksAsync` i'm checking for that.

---
In reply to: [535655681](https://github.com/microsoft/sarif-visualstudio-extension/pull/310#discussion_r535655681) [](ancestors = 535655681)

src/Sarif.Sarifer/BackgroundAnalyzerBase.cs
eddynaka: got it.

---
In reply to: [535656794](https://github.com/microsoft/sarif-visualstudio-extension/pull/310#discussion_r535656794) [](ancestors = 535656794,535655053,535653423)

src/Sarif.Sarifer/BackgroundAnalysisService.cs
lgolding: You also have to check in `DisposeStreams`.

So rather than checking in both places, you can filter out the `null` ones here (as in my code fragment above) and then you never have to worry about t again.

---
In reply to: [535657988](https://github.com/microsoft/sarif-visualstudio-extension/pull/310#discussion_r535657988) [](ancestors = 535657988,535655681)

eddynaka: the dispose of a Stream.null is a Noop: [https://referencesource.microsoft.com/#mscorlib/system/io/stream.cs,b7f982f75278c3c9](https://referencesource.microsoft.com/#mscorlib/system/io/stream.cs,b7f982f75278c3c9). I didn't do the where/toarray, because we will create a new array, so more objects used.

---
In reply to: [535661666](https://github.com/microsoft/sarif-visualstudio-extension/pull/310#discussion_r535661666) [](ancestors = 535661666,535657988,535655681)

lgolding: Nice! TIL. I withdraw my objection.

---
In reply to: [535665533](https://github.com/microsoft/sarif-visualstudio-extension/pull/310#discussion_r535665533) [](ancestors = 535665533,535661666,535657988,535655681)

src/Sarif.Sarifer/BackgroundAnalyzerBase.cs
```diff
@@ -49,68 +39,95 @@ public abstract class BackgroundAnalyzerBase : IBackgroundAnalyzer
...
 
-                await writer.FlushAsync().ConfigureAwait(continueOnCapturedContext: false);
+                    await writer.FlushAsync().ConfigureAwait(continueOnCapturedContext: false);
+                }
 
-                await this.WriteToSinksAsync(path, stream).ConfigureAwait(false);
+                return stream;
+            }
+#pragma warning disable CA1031 // Do not catch general exception types
+            catch (Exception)
```
michaelcfanning, line 80: >Exception [](start = 19, length = 9)

Why are we eating all exceptions here? We shouldn't do that, as CA1031 tells you. First, in cases where it is actually necessary for this kind of code, there should always be a logging mechanism capturing details. In this case, why would we catch this at all? The add-in is behaving poorly, well, let it crash. The danger of leaving this code as is is that analysis could be failing catastrophically and the user will never know. Meanwhile, analysis is completely disabled and problems could be sneaking into code.

src/Sarif.Sarifer/SpamBackgroundAnalyzer.cs
```diff
@@ -40,7 +41,7 @@ protected override void AnalyzeCore(Uri uri, string text, string solutionDirecto
                 || (this.currentSolutionDirectory?.Equals(solutionDirectory, StringComparison.OrdinalIgnoreCase) != true))
             {
                 // clear older rules
-                this.rules?.Clear();
+                this.rules.Clear();
```
michaelcfanning, line 44: >rules [](start = 21, length = 5)

Just curious, why did you make this change? What you've done is to prematurely allocate an object, which will *always* be overwritten in any case. i.e., you create it so that you can call Clear on it and then reassign a new value. Why wouldn't we just perform the null check here?

One reason you might have had in mind is that this could would cause a null deref in, say, cases where a pervious LoadSearchDefinitionsFiles call returns null. If so, I'd suggest a debug assert or an actual verification after that call (that we received a non-null result).

michaelcfanning (approved):
:shipit:

src/Sarif.Sarifer/BackgroundAnalyzerBase.cs
eddynaka: ok. removing this.

---
In reply to: [536220395](https://github.com/microsoft/sarif-visualstudio-extension/pull/310#discussion_r536220395) [](ancestors = 536220395)

src/Sarif.Sarifer/SpamBackgroundAnalyzer.cs
eddynaka: imagine the following:
opened a single file (without sln, / spam), that will bring the solutionDirectory to null, since we don't have the path, rules would still be null. With that, when we call the analyze we will have a null reference.
My idea was: to prevent that, i did create the hashset. What I can do is after this, i can check if rules is null, if thats true, i will just return.

---
In reply to: [536222467](https://github.com/microsoft/sarif-visualstudio-extension/pull/310#discussion_r536222467) [](ancestors = 536222467)

PR #312, opened 2020-12-04
Fixing analysis for single file

eddynaka:


src/Sarif.Sarifer/SpamBackgroundAnalyzer.cs
```diff
@@ -66,7 +66,10 @@ protected override void AnalyzeCore(Uri uri, string text, string solutionDirecto
             using (context)
             {
                 cancellationToken.ThrowIfCancellationRequested();
-                AnalyzeCommand.AnalyzeTargetHelper(context, this.rules, disabledSkimmers);
+
+                // Filtering file before analyzing.
+                IEnumerable<Skimmer<AnalyzeContext>> applicableSkimmers = AnalyzeCommand.DetermineApplicabilityForTargetHelper(context, this.rules, disabledSkimmers);
+                AnalyzeCommand.AnalyzeTargetHelper(context, applicableSkimmers, disabledSkimmers);
```
eddynaka, line 72: @michaelcfanning , I was trying to simulate a file where it does not finish with the expected extension, with that, we shouldn't show error.
The first thing that I noticed is that based on the unittests, we have to filter (this change does that).
The second thing: what is the difference between `DefaultNameRegex` and `NameRegex`? I saw that the filter is just applied for the `DefaultNameRegex`.
Thank you :)

src/Sarif.Sarifer/SariferPackageCommand.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            ChangeAnalyzeCommand(menuCommandService, enabled: false);
+        }
+
+        public static void EnableAnalyzeCommands(IMenuCommandService menuCommandService)
+        {
+            ChangeAnalyzeCommand(menuCommandService, enabled: true);
+        }
+
+        private static void ChangeAnalyzeCommand(IMenuCommandService menuCommandService, bool enabled)
```
lgolding, line 20: >Command [](start = 41, length = 7)

"Commands", since you're changing both of them.

src/Sarif.Sarifer/BackgroundAnalysisService.cs
```diff
@@ -20,15 +21,24 @@ internal class BackgroundAnalysisService : IBackgroundAnalysisService
     {
 #pragma warning disable IDE0044, CS0649 // Provided by MEF
         [ImportMany]
-        private IEnumerable<IBackgroundAnalyzer> analyzers;
+        private readonly IEnumerable<IBackgroundAnalyzer> analyzers;
```
lgolding, line 24: >readonly [](start = 16, length = 8)

If you do this then you don't need one of the two pragmas (IDE0044 or CS0649). One of them, I don't remember which, is about the lack of `readonly`.

src/Sarif.Sarifer/BackgroundAnalysisService.cs
```diff
@@ -46,11 +56,14 @@ public async Task AnalyzeAsync(string path, string text, CancellationToken cance
             {
                 DisposeStreams(streams);
             }
+
+            AnalysisCompleted?.Invoke(this, null);
```
lgolding, line 60: >AnalysisCompleted?.Invoke [](start = 12, length = 25)

Don't you want to do this in the `finally`? Otherwise you could end up with the menu items disabled forever.

src/Sarif.Sarifer/BackgroundAnalysisService.cs
```diff
@@ -67,6 +80,9 @@ public async Task AnalyzeAsync(string logId, IEnumerable<string> targetFiles, Ca
...
+            this.canAnalyzeFile = true;
```
lgolding, line 85: Shouldn't this be in the `finally`?

src/Sarif.Sarifer/TextBufferViewTracker.cs
```diff
@@ -61,10 +61,23 @@ public void RemoveTextView(ITextView textView)
...
+                trackingInformation.CancellationTokenSource.Dispose();
                 LastViewRemoved?.Invoke(
                     this,
-                    new LastViewRemovedEventArgs(trackingInformation.LogId));
+                    new LastViewRemovedEventArgs(trackingInformation.Path));
             }
         }
+
+        /// <inheritdoc/>
+        public void Clear()
```
lgolding, line 72: >Clear [](start = 20, length = 5)

Who calls this?

eddynaka: When we dispose the class `BackgroundAnalysisTextViewCreationListener`

---
In reply to: [539663388](https://github.com/microsoft/sarif-visualstudio-extension/pull/312#discussion_r539663388) [](ancestors = 539663388)

lgolding (approved):
:shipit:

src/Sarif.Sarifer/LastViewRemovedEventArgs.cs
```diff
@@ -17,17 +17,19 @@ public class LastViewRemovedEventArgs : EventArgs
         /// <summary>
         /// Initializes a new instance of the <see cref="LastViewRemovedEventArgs"/> class.
         /// </summary>
-        /// <param name="logId">
-        /// The unique id of the SARIF log associated with this text buffer.
+        /// <param name="path">
+        /// The path to the file whose contents are being viewed, or <c>null</c> if
+        /// <paramref name="textBuffer"/> is not associated with a file.
```
michaelcfanning, line 22: >textBuffer [](start = 28, length = 10)

what's the param we reference here?

src/Sarif.Sarifer/SariferPackageCommand.cs
```diff
@@ -0,0 +1,32 @@
...
+
+using System.ComponentModel.Design;
+
+namespace Microsoft.CodeAnalysis.Sarif.Sarifer
+{
+    internal static class SariferPackageCommand
+    {
+        public static void DisableAnalyzeCommands(IMenuCommandService menuCommandService)
+        {
+            ChangeAnalyzeCommands(menuCommandService, enabled: false);
```
michaelcfanning, line 12: >ChangeAnalyzeCommands [](start = 12, length = 21)

UpdateAnalyzeCommandEnableState? or maybe simply UpdateAnalyzeCommandState?

michaelcfanning (approved):
:shipit:

PR #313, opened 2020-12-04
Fix #114: Viewer must filter out 'absent' and 'suppressed' issues by default

### LINKED ISSUE #114: Viewer must filter out 'absent' and 'suppressed' issues by default

lgolding:
This implementation hews as closely as possible to the [description](https://github.com/microsoft/sarif-visualstudio-extension/issues/114#issuecomment-728197826) in #114; please read that before the following:

- The Suppression State column is hidden by default. It appears the first time the user opens a file containing suppression information, at which time "Suppressed" results are filtered out. If the user checks the "Suppressed" box in the filter UI, suppressed results are never again filtered out during this VS session.

- The Category column, for some reason, is visible by default as soon as the user opens any log file. The first time the user opens a log file with "Absent" results, the filter is adjusted to exclude them. If the user checks the "Absent" box in the filter UI, absent results are never again filtered out during this VS session.

Also:
- Replace the `SarifDataTableSink` class, which unnecessarily provided _wrappers_ around all sink operations, with the `SinkHolder` class, which does nothing but _hold_ a sink. The comments explain why. Mainly I did this because the existing code caused me to waste time wondering why it was necessary to wrap the sink operations, and it turned out the answer was that it _wasn't_ necessary.

- Fix a bug in a project file that prevented an `xcopy` operation from working.

**UI test authoring is in progress.**


src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -505,6 +505,16 @@ private int WriteRunToErrorList(Run run, string logFilePath)
                 }
             }
 
+            if (run.HasAbsentResults())
```
lgolding, line 508: >if [](start = 12, length = 2)

This is the relevant code, together with the implementations of the "stub" table data sources and their base class

src/Sarif.Viewer.VisualStudio/ErrorList/SarifTableDataSourceBase.cs
```diff
@@ -0,0 +1,68 @@
...
+                }
+
+                if (this.TableControlEventProcessorProviders == null)
+                {
+                    this.TableControlEventProcessorProviders = new[]
+                        { compositionService.GetService<ITableControlEventProcessorProvider>() };
+                }
+
+                ITableManager manager = this.TableManagerProvider.GetTableManager(StandardTables.ErrorsTable);
+                manager.AddSource(this, columns);
```
lgolding, line 62: >manager.AddSource(this, columns) [](start = 16, length = 32)

Here's where we register the sources. Both the "stub" sources and the "real" one (the one that actually provides Error List entries) derive from this base class. Their ctors all call this `Initialize` method with the appropriate set of columns.

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -514,6 +524,18 @@ private int WriteRunToErrorList(Run run, string logFilePath)
...
 
+        private void ShowFilteredSuppressionColumn()
+        {
+            // Creating this table source causes the Suppression State column to appear.
+            SarifSuppressedResultsTableDataSource _ = SarifSuppressedResultsTableDataSource.Instance;
+        }
+
+        private void ShowFilteredCategoryColumn()
+        {
+            // Creating this table source causes the Category column to appear.
```
lgolding, line 535: >appear [](start = 72, length = 6)

It turns out the Category column is visible by default, and there doesn't seem to be a way to change that. Still investigating.

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -514,6 +524,18 @@ private int WriteRunToErrorList(Run run, string logFilePath)
             return sarifErrors.Count;
         }
 
+        private void ShowFilteredSuppressionColumn()
```
lgolding, line 527: >Filtered [](start = 25, length = 8)

I haven't found a way to programmatically set a column filter ("Don't show results whose suppression state is 'Suppressed'".

src/Sarif.Viewer.VisualStudio/ErrorList/SinkWrapper.cs
```diff
@@ -0,0 +1,61 @@
...
+    // table data source, Visual Studio calls ITableDataSource.Subscribe on the source,
+    // providing a "sink": an object that implements ITableDataSink. In return, VS
+    // receives a disposable object associated with the sink. When the source is
+    // removed, VS calls Dispose on this object, at which point the client must
+    // stop making calls to that sink.
+    //
+    // To accomplish this, SinkWrapper raises an event when VS disposes it. The
+    // receiver must remove the associated sink from the list of sinks it maintains.
+    // See SarifTableDataSource.TableSink_Disposed.
+    internal class SinkWrapper : IDisposable
```
lgolding, line 23: >SinkWrapper [](start = 19, length = 11)

This class replaces `SarifTableDataSink`, which did more than it needed to in wrapping the calls to `IDataTableSink`.

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
lgolding: Progress: You do it by calling `SetFilter` on an `IWpfTableControl` representing the Error List. Next challenge is how to get the interface.

```
            this.ErrorListTableControl.SetFilter(
                SarifResultTableEntry.SuppressionStateColumnName,
                new ColumnHashSetFilter(
                    suppressionStateColumnDefinition,
                    VSSuppressionState.Suppressed.ToString()));
```

---
In reply to: [537694588](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r537694588) [](ancestors = 537694588)

src/Sarif.Viewer.VisualStudio/ErrorList/ColumnFilterer.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        private IWpfTableControl errorListTableControl;
+
+        private IWpfTableControl ErrorListTableControl
+        {
+            get
+            {
+                if (this.errorListTableControl == null)
+                {
+                    var componentModel = (IComponentModel)Package.GetGlobalService(typeof(SComponentModel));
```
dpugh, line 30: My expectation is you can also do this via:
    var errorList = Package.GetGlobalService(typeof(SVsErrorList)) as IErrorList;
    return errorList?.TableControl;

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -514,6 +527,27 @@ private int WriteRunToErrorList(Run run, string logFilePath)
             return sarifErrors.Count;
         }
 
+        private void ShowFilteredSuppressionStateColumn()
+        {
+            // Creating this table source adds "Suppression State" to the list of available columns.
+            SarifSuppressedResultsTableDataSource _ = SarifSuppressedResultsTableDataSource.Instance;
```
dpugh, line 533: These should no-op if the column has already been added (if they are not already doing that).

src/Sarif.Viewer.VisualStudio/ErrorList/ColumnFilterer.cs
dpugh: NM, I see the implementation of IErrorListTableControlProvider.

I'm not sure whether there's any value to the extra level of indirection.

src/Sarif.Viewer.VisualStudio/ErrorList/SarifResultTableEntry.cs
```diff
@@ -22,14 +22,35 @@ namespace Microsoft.Sarif.Viewer.ErrorList
...
+        // There is not a direct mechanism to add or remove columns based on an individual entry.
+        // You can, however, add a new ITableDataSource (ITableManager.AddSource(source[, columns])
+        // where you can specify the columns you want to appear in the table (at which point VS adds
+        // the column to the table, though it won't be shown to the user unless it is visible).
+        // Removing the data source removes the column (although we don't have a requirement to do
+        // that), unless some other source also added the same column.
+
+        // The "basic" columns are always displayed.
+        public static readonly ReadOnlyCollection<string> BasicColumns = new ReadOnlyCollection<string>(new[] {
             StandardTableKeyNames2.TextInlines,
```
dpugh, line 52: These really should be the corresponding column names (StandardTableColumnDefinitions....), It doesn't matter for these -- since the desired column name matches the key name but it is confusing.

dpugh, line 52: I don't think there is a column with this name (there is a key with this name that is used to fill in the contents of the Text column) so adding TextInlines to the columns should have no effect.

src/Sarif.Viewer.VisualStudio/ErrorList/SarifTableDataSourceBase.cs
```diff
@@ -0,0 +1,68 @@
...
+    //
+    // For an explanation of why these three sources are necessary, see the comment
+    // near the top of SarifResultTableEntry.cs
+    internal abstract class SarifTableDataSourceBase : ITableDataSource
+    {
+        [Import]
+        private ITableManagerProvider TableManagerProvider { get; set; } = null;
+
+        [ImportMany]
+        private IEnumerable<ITableControlEventProcessorProvider> TableControlEventProcessorProviders { get; set; } = null;
```
dpugh, line 31: What are you using this for?

src/Sarif.Viewer.VisualStudio/SarifViewerPackage.cs
```diff
@@ -31,6 +31,7 @@ namespace Microsoft.Sarif.Viewer
     [ProvideService(typeof(ISarifLocationTaggerService))]
     [ProvideService(typeof(ITextViewCaretListenerService<>))]
     [ProvideService(typeof(ISarifErrorListEventSelectionService))]
+    [ProvideService(typeof(IErrorListTableControlProvider))]
```
eddynaka, line 34: should we change to `IErrorListTableControlService`?

src/Sarif.Viewer.VisualStudio/SarifViewerPackage.cs
```diff
@@ -98,7 +100,7 @@ protected override async Task InitializeAsync(CancellationToken cancellationToke
         {
             await base.InitializeAsync(cancellationToken, progress);
```
eddynaka, line 101: should we call `.ConfigureAwait`?

src/Sarif.Viewer.VisualStudio/ErrorList/SarifSuppressedResultsTableDataSource.cs
```diff
@@ -0,0 +1,51 @@
...
+            get
+            {
+                if (_instance == null)
+                {
+                    _instance = new SarifSuppressedResultsTableDataSource();
+                }
+
+                return _instance;
+            }
+        }
```
eddynaka, line 43: We could change to:
```csharp
return _instance ?? (_instance = new SarifSuppressedResultsTableDataSource());
```
or
```csharp
public static SarifSuppressedResultsTableDataSource Instance => _instance ?? (_instance = new SarifSuppressedResultsTableDataSource());
```

src/Sarif.Viewer.VisualStudio/ErrorList/SarifResultTableEntry.cs
```diff
@@ -79,9 +107,7 @@ public SarifResultTableEntry(SarifErrorListItem error)
             this.columnKeyToContent[StandardTableKeyNames.ProjectName] = this.Error.ProjectName;
 
             string supressionState = this.Error.VSSuppressionState.ToString();
```
eddynaka, line 109: should we create a fixed variable containing the strings instead of creating the string each time that this method is called?

src/Sarif.Viewer.VisualStudio/ErrorList/SarifResultTableEntry.cs
```diff
@@ -22,14 +22,35 @@ namespace Microsoft.Sarif.Viewer.ErrorList
...
+        //
+        // There is not a direct mechanism to add or remove columns based on an individual entry.
+        // You can, however, add a new ITableDataSource (ITableManager.AddSource(source[, columns])
+        // where you can specify the columns you want to appear in the table (at which point VS adds
+        // the column to the table, though it won't be shown to the user unless it is visible).
+        // Removing the data source removes the column (although we don't have a requirement to do
+        // that), unless some other source also added the same column.
+
+        // The "basic" columns are always displayed.
+        public static readonly ReadOnlyCollection<string> BasicColumns = new ReadOnlyCollection<string>(new[] {
```
eddynaka, line 51: Probably we could simplify this to public const string[] ...

src/Sarif.Viewer.VisualStudio/SarifViewerPackage.cs
lgolding: Fixed.

---
In reply to: [538267062](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r538267062) [](ancestors = 538267062)

src/Sarif.Viewer.VisualStudio/ErrorList/ColumnFilterer.cs
lgolding: I originally thought about injecting the `IWpfTableControl` into this class, so I wanted a provider for it, but yes, it's overkill.

---
In reply to: [537999712](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r537999712) [](ancestors = 537999712)

src/Sarif.Viewer.VisualStudio/SarifViewerPackage.cs
lgolding: I removed it at David's suggestion. If I had kept it, I would have preferred the name "Provider" since that's what it does -- provide the control. It's a common naming pattern in VS.

---
In reply to: [538264056](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r538264056) [](ancestors = 538264056)

src/Sarif.Viewer.VisualStudio/ErrorList/SarifResultTableEntry.cs
lgolding: No, it varies from input `error` to input `error` (note that `this.Error = error`.

---
In reply to: [538282371](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r538282371) [](ancestors = 538282371)

src/Sarif.Viewer.VisualStudio/ErrorList/SarifSuppressedResultsTableDataSource.cs
lgolding: Yes, nice.

---
In reply to: [538269677](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r538269677) [](ancestors = 538269677)

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
lgolding: Yes, that is the behavior I expected and observed. Were you suggesting a code change, or just making sure I understood the implications of this line?

---
In reply to: [537999155](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r537999155) [](ancestors = 537999155)

src/Sarif.Viewer.VisualStudio/ErrorList/SarifResultTableEntry.cs
lgolding: Yes, you're right, it confused me. I'll change them.

---
In reply to: [538000615](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r538000615) [](ancestors = 538000615)

lgolding: You're right. Several of these don't have corresponding columns, and I removed them:
- `TextInlines`
- `FullText`
- `HelpLink`
- `ErrorCodeTooltip`

---
In reply to: [538001244](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r538001244) [](ancestors = 538001244)

src/Sarif.Viewer.VisualStudio/ErrorList/SarifTableDataSource.cs
```diff
@@ -4,69 +4,34 @@
...
 
 namespace Microsoft.Sarif.Viewer.ErrorList
 {
-    internal class SarifTableDataSource : ITableDataSource, IDisposable
+    internal class SarifTableDataSource : SarifTableDataSourceBase, IDisposable
     {
         private static SarifTableDataSource _instance;
         private readonly ReaderWriterLockSlimWrapper sinksLock = new ReaderWriterLockSlimWrapper(new ReaderWriterLockSlim());
-        private readonly List<SarifTableDataSink> sinks = new List<SarifTableDataSink>();
+        private readonly List<SinkHolder> sinkHolders = new List<SinkHolder>();
```
michaelcfanning, line 20: >new List<SinkHolder>(); [](start = 56, length = 23)

NIT: Just a style thing, please move all inlined field initializers to the private ctor. The reason we do this is to gather all non-default init in a common place and to prevent the debugger from jumping around unduly when stepping through.

michaelcfanning:
Curious, why is the sarif-pattern-matcher submodule updated for this change?

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -514,6 +529,31 @@ private int WriteRunToErrorList(Run run, string logFilePath)
             return sarifErrors.Count;
         }
 
+        private void ShowFilteredSuppressionStateColumn()
```
michaelcfanning, line 532: >ShowFilteredSuppressionStateColumn [](start = 21, length = 34)

Can you please add a comment that explicitly describes the behavior we're looking for? i.e., in the event that we detect one or more suppressed items, we want to make the suppressions column visible and filter out suppressed items (so that they are not visible by default). This detailed comment will help readers validate whether the implementation meets expectations.

For example, if the suppressions column is *already* visible, and the user  has explicitly enabled viewing of suppressed items, we don't want to clear their setting. Can you confirm whether your code handles this scenario correctly? On surface it doesn't appear to.

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -514,6 +529,31 @@ private int WriteRunToErrorList(Run run, string logFilePath)
...
+
+            // Creating this table source adds "Suppression State" to the list of available columns.
+            SarifSuppressedResultsTableDataSource _ = SarifSuppressedResultsTableDataSource.Instance;
+
+            this.columnFilterer.FilterOut(
+                columnName: SarifResultTableEntry.SuppressionStateColumnName,
+                filteredValue: VSSuppressionState.Suppressed.ToString());
+        }
+
+        private void ShowFilteredCategoryColumn()
```
michaelcfanning, line 544: >ShowFilteredCategoryColumn [](start = 21, length = 26)

same note as previous, add detailed comment and verify that we're not overwriting explicitly set user filters that make 'absent' results results visible.

Also, separately, we need an issue tracking our permanent mechanism for handling baseline state. It's not clear that repurposing the 'Category' column is the right thing to do for the long-term.

src/Sarif.Viewer.VisualStudio/ErrorList/SarifTableDataSourceBase.cs
lgolding: Good catch. This was existing code that I pushed down from the concrete `SarifTableDataSource` class. The stub classes certainly don't need it -- and in fact I now see that this is a private imported property that is not used by the importing class. So I will remove it.

---
In reply to: [538011095](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r538011095) [](ancestors = 538011095)

src/Sarif.Viewer.VisualStudio/ErrorList/FilteredColumnValue.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.Sarif.Viewer.ErrorList
+{
+    // Associates an error list column with a value that has been filtered from it.
+    internal class FilteredColumnValue
```
michaelcfanning, line 9: >FilteredColumnValue [](start = 19, length = 19)

Just curious, attributes of this class, value type comparison, immutability after init, data only/no behavior, suggests you could simply prefer a value type. Is there a reason not to?

lgolding: No reason not to, good catch. Changed to `struct`.

---
In reply to: [538749153](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r538749153) [](ancestors = 538749153)

src/Sarif.Viewer.VisualStudio/ErrorList/SarifAbsentResultsTableDataSource.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        private static SarifAbsentResultsTableDataSource _instance;
+
+        private SarifAbsentResultsTableDataSource()
+        {
+            if (!SarifViewerPackage.IsUnitTesting)
+            {
+#pragma warning disable VSTHRD108 // Assert thread affinity unconditionally
+                ThreadHelper.ThrowIfNotOnUIThread();
+#pragma warning restore VSTHRD108 // Assert thread affinity unconditionally
```
michaelcfanning, line 26: >// Assert thread affinity unconditionally [](start = 34, length = 41)

More style, I personally don't find this duplication of rule description provides for better readability when scope is limited to a single line like this. I can appreciate that this is the default code emit behavior, but I usually remove these. My care factor is very small on this point, though.

src/Sarif.Viewer.VisualStudio/ErrorList/SarifAbsentResultsTableDataSource.cs
```diff
@@ -0,0 +1,51 @@
...
+            }
+        }
+
+        public static SarifAbsentResultsTableDataSource Instance
+        {
+            get
+            {
+                if (_instance == null)
+                {
+                    _instance = new SarifAbsentResultsTableDataSource();
```
michaelcfanning, line 38: >_instance [](start = 20, length = 9)

You aren't using the assignment coalescing operator I assume because we haven't upgraded to 8.0 yet? If so: [https://github.com/microsoft/sarif-visualstudio-extension/issues/314](https://github.com/microsoft/sarif-visualstudio-extension/issues/314). Low priority update.

src/Sarif.Viewer.VisualStudio/ErrorList/SarifAbsentResultsTableDataSource.cs
```diff
@@ -0,0 +1,51 @@
...
+
+                return _instance;
+            }
+        }
+
+        public override string Identifier => $"{Guids.GuidVSPackageString}-{nameof(SarifAbsentResultsTableDataSource)}";
+
+        public override string DisplayName => Resources.ErrorListAbsentResultsDataSourceDisplayName;
+
+        public override IDisposable Subscribe(ITableDataSink sink) => new StubDisposable();
```
michaelcfanning, line 49: >Subscribe [](start = 36, length = 9)

Can you verify that calls to this method are sufficiently 'not chatty' to warrant creating a new object on each call, rather than, say, initializing and returning a class member?

michaelcfanning:
>             this.columnKeyToContent[StandardTableKeyNames2.TextInlines] = new Lazy<object>(() =>

You didn't make a change here, but there's something I don't understand about this code: if we produce inlines in a lazy fashion, how does the IDE understand whether to produce the expand/collapse chrome on the error list? are these values retrieved when messages are displayed (only) in the virtualized error list?

---
Refers to: src/Sarif.Viewer.VisualStudio/ErrorList/SarifResultTableEntry.cs:114 in 21bf8b0. [](commit_id = 21bf8b06fb90447171635f418d90051d403debc9, deletion_comment = False)

src/Sarif.Viewer.VisualStudio/ErrorList/SarifResultTableEntry.cs
```diff
@@ -22,14 +22,35 @@ namespace Microsoft.Sarif.Viewer.ErrorList
...
         private bool isDisposed;
 
         private readonly Dictionary<string, object> columnKeyToContent = new Dictionary<string, object>(StringComparer.InvariantCulture);
 
-        public static readonly ReadOnlyCollection<string> SupportedColumns = new ReadOnlyCollection<string>(new[] {
+        // The set of columns we display depends on the contents of the displayed log file.
+        // Per https://github.com/microsoft/sarif-visualstudio-extension/issues/114, we do not
+        // display "suppressed" or "absent" issues by default.
+        //
+        // If we open a log file that contains a suppressed issue, we display the "Suppression
```
michaelcfanning, line 35: >contains [](start = 38, length = 8)

Great comment here, as noted, co-locating some of this w/the actual implementation rendering this behavior is a good idea.

src/Sarif.Viewer.VisualStudio/ErrorList/SarifSuppressedResultsTableDataSource.cs
michaelcfanning: NIT: C# 8.0 one-liner:

return _instance ??= new SarifSuppressedResultsTableDataSource();

---
In reply to: [538737288](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r538737288) [](ancestors = 538737288,538269677)

lgolding:
>             this.columnKeyToContent[StandardTableKeyNames2.TextInlines] = new Lazy<object>(() =>

For each table entry (and this class does implement the `ITableEntry` interface), VS calls `TryGetValue` (implemented below) when it needs information from the entry. For instance, if it asks for the key "fulltext", and the entry has data for that key, that's how VS knows to show the expander. As you say, it presumably only asks for information about the items it needs to display.






---
In reply to: [740923590](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#issuecomment-740923590) [](ancestors = 740923590)

---
Refers to: src/Sarif.Viewer.VisualStudio/ErrorList/SarifResultTableEntry.cs:114 in 21bf8b0. [](commit_id = 21bf8b06fb90447171635f418d90051d403debc9, deletion_comment = False)

src/Sarif.Viewer.VisualStudio/ErrorList/SarifTableDataSource.cs
```diff
@@ -84,42 +49,33 @@ public static SarifTableDataSource Instance
...
+        public override string Identifier => Guids.GuidVSPackageString;
 
-        public string DisplayName
-        {
-            get { return Resources.ErrorListTableDataSourceDisplayName; }
-        }
+        public override string DisplayName => Resources.ErrorListTableDataSourceDisplayName;
 
-        public IDisposable Subscribe(ITableDataSink sink)
+        public override IDisposable Subscribe(ITableDataSink sink)
```
michaelcfanning, line 57: >Subscribe [](start = 36, length = 9)

I can infer some information on Subscribe chattiness, etc., requested in my previous comment from this implementation. :)

src/Sarif.Viewer.VisualStudio/ErrorList/SarifSuppressedResultsTableDataSource.cs
lgolding: Indeed, my favorite C# 8 feature.

---
In reply to: [538759807](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r538759807) [](ancestors = 538759807,538737288,538269677)

src/Sarif.Viewer.VisualStudio/ErrorList/SarifTableDataSource.cs
```diff
@@ -250,35 +206,35 @@ internal bool HasErrors(string sourceFileName)
...
-                return logFileToTableEntries.Values.Any((errorList) => errorList.Any((error) => error.Error.FileName.Equals(sourceFileName, StringComparison.Ordinal)));
+                return this.logFileToTableEntries.Values.Any((errorList) => errorList.Any((error) => error.Error.FileName.Equals(sourceFileName, StringComparison.Ordinal)));
             }
         }
 
-        private void CallSinks(Action<SarifTableDataSink> action)
+        private void CallSinks(Action<ITableDataSink> action)
         {
-            IReadOnlyList<SarifTableDataSink> sinkManagers;
+            IReadOnlyList<SinkHolder> sinkHolers;
```
michaelcfanning, line 215: >sinkHolers [](start = 38, length = 10)

type in variable name

src/Sarif.Viewer.VisualStudio/ErrorList/SarifTableDataSourceBase.cs
```diff
@@ -0,0 +1,68 @@
...
+        public abstract string DisplayName { get; }
+
+        protected void Initialize(IReadOnlyCollection<string> columns)
+        {
+            ThreadHelper.ThrowIfNotOnUIThread();
+
+            var compositionService = ServiceProvider.GlobalProvider.GetService(typeof(SComponentModel)) as IComponentModel;
+
+            // The composition service will only be null in unit tests.
+            if (compositionService != null)
```
michaelcfanning, line 46: >(compositionService != null) [](start = 15, length = 28)

style nit, i would just return early here when compositionService is null, since init is entirely unnecessary for this scenario

src/Sarif.Viewer.VisualStudio/ErrorList/StubDisposable.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.Sarif.Viewer.ErrorList
+{
+    // This class exists solely to provide an object for the "stub" table data sources to return
+    // when the TableDataManager calls Subscribe on them.
+    internal class StubDisposable : IDisposable
```
michaelcfanning, line 10: >StubDisposable [](start = 19, length = 14)

looks like you don't require more than a single static instance of this class?

lgolding: I suppose, since `Dispose` doesn't actually do anything -- but I really don't want a Singleton `IDisposable` which in fact will be disposed multiple times by VS (once per stub data source). And since there are only ever two such stub data sources, I'd rather not change this.

---
In reply to: [538766265](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r538766265) [](ancestors = 538766265)

src/Sarif.Viewer.VisualStudio/ErrorList/SarifAbsentResultsTableDataSource.cs
lgolding: Yes, that's the reason.

---
In reply to: [538753105](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r538753105) [](ancestors = 538753105)

src/Sarif.Viewer.VisualStudio/ErrorList/SarifAbsentResultsTableDataSource.cs
lgolding: Yes: VS calls this once on each new table data source that is registered, and there are only 3 of them.

---
In reply to: [538753999](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r538753999) [](ancestors = 538753999)

lgolding: Edit: Sorry, once per _sink_, on each new table data source. I admit I don't know how many sinks there are.

---
In reply to: [538775760](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r538775760) [](ancestors = 538775760,538753999)

src/Sarif.Viewer.VisualStudio/ErrorList/SarifTableDataSource.cs
lgolding: Not sure what you mean. What about the implementation do you notice, and what do you infer from it?

---
In reply to: [538760758](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r538760758) [](ancestors = 538760758)

src/Sarif.Viewer.VisualStudio/ErrorList/SarifAbsentResultsTableDataSource.cs
lgolding: Changed everywhere.

---
In reply to: [538750192](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r538750192) [](ancestors = 538750192)

lgolding:
I updated the viewer's submodule pointer to use the latest `sarif-pattern-matcher`, and I did that only to get a version of `sarif-pattern-matcher` whose submodule pointer pointed to the latest `sarif-sdk`, and I did that because I wanted the newest `Sarif\Core\Result` helper methods.

And I wanted them because of the recently added `Result.IsSuppressed` and `Result.IsAbsent` methods that I wanted to use in my UI tests...

... except...

... that it turns out I misremembered, and it was actually **Run** that has the new extension methods, and they are _HasSuppressedResults_, _HasAbsentResults_. The `Result` methods that I wanted to do, don't actually exist.

But I decided there was no harm in keeping the viewer up to date anyway.

---
In reply to: [740892336](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#issuecomment-740892336) [](ancestors = 740892336)

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
lgolding: The only way to fix this would be to define a column just like Category (say, BaseLine State), make if invisible by default, and provide a programmatic interface to turn it on. This is consistent with our desire not to overload the Category column.

@mikefan@microsoft.com, would you like me to do that as part of this PR, or file an issue per your comment above?

---
In reply to: [537693891](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r537693891) [](ancestors = 537693891)

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -514,6 +529,31 @@ private int WriteRunToErrorList(Run run, string logFilePath)
             return sarifErrors.Count;
         }
 
+        private void ShowFilteredSuppressionStateColumn()
+        {
+            ThreadHelper.ThrowIfNotOnUIThread();
+
+            // Creating this table source adds "Suppression State" to the list of available columns.
+            SarifSuppressedResultsTableDataSource _ = SarifSuppressedResultsTableDataSource.Instance;
```
GabeDeBacker, line 537: >     SarifSuppressedResultsTableDataSource _ = SarifSuppressedResultsTableDataSource.Instance; [](start = 7, length = 94)

Why not create a static method to create the table source (if needed) rather than using discard?

src/Sarif.Viewer.VisualStudio/ErrorList/ColumnFilterer.cs
```diff
@@ -0,0 +1,63 @@
...
+using Microsoft.VisualStudio.Shell.Interop;
+using Microsoft.VisualStudio.Shell.TableControl;
+
+namespace Microsoft.Sarif.Viewer.ErrorList
+{
+    // This class allows the caller to filter specified values from specified columns.
+    // It keeps track of which column/value pairs have been filtered, so that it only
+    // does so once per column/value pair. The idea is that if the user ever checks the
+    // box in the filter UI to show a certain value in a certain column, we should never
+    // hide that column again.
```
lgolding, line 16: >again [](start = 24, length = 5)

@mikefan@microsoft.com Note that this comment contains the information you asked for above the definitions of `ErrorListService.ShowFilteredSuppressionsStateColumn` and `ErrorListService.ShowFilteredCategoryColumn`. I will repeat it in those locations.

src/Sarif.Viewer.VisualStudio/ErrorList/SarifAbsentResultsTableDataSource.cs
```diff
@@ -0,0 +1,36 @@
...
+        private static SarifAbsentResultsTableDataSource _instance;
+
+        private SarifAbsentResultsTableDataSource()
+        {
+            ThreadHelper.ThrowIfNotOnUIThread();
+
+            this.Initialize(SarifResultTableEntry.AbsentResultColumns);
+        }
+
+        public static SarifAbsentResultsTableDataSource Instance =>
```
GabeDeBacker, line 27: >static [](start = 15, length = 6)

So sad, anyway to make this not static?

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
lgolding: The user's filter setting to display absent or suppressed results persists through the lifetime of the VS session. I've verified that repeatedly, and am in the process of enhancing the UI tests to demonstrate it. I filed [https://github.com/microsoft/sarif-visualstudio-extension/issues/315](https://github.com/microsoft/sarif-visualstudio-extension/issues/315), "Decide where to display Baseline State".

---
In reply to: [538747072](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r538747072) [](ancestors = 538747072)

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
lgolding: I added the comment. The code does handle this, in the `ColumnFilterer` class (which you probably hadn't read at the time you made the comment).

---
In reply to: [538746026](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r538746026) [](ancestors = 538746026)

src/Sarif.Viewer.VisualStudio/ErrorList/SarifResultTableEntry.cs
lgolding: Done.

---
In reply to: [538756999](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r538756999) [](ancestors = 538756999)

src/Sarif.Viewer.VisualStudio/ErrorList/SarifAbsentResultsTableDataSource.cs
GabeDeBacker: Using MEF services local to the extension (not promoted) is the way to remove statics.

I could dig up lots of Watson hangs in VS due to the way the CLR initializes statics. (Let's just say it involves a global lock and it's super easy to deadlock between VS's UI thread and initialization of a static object)

---
In reply to: [538922382](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r538922382) [](ancestors = 538922382)

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
lgolding: On second thought, you said elsewhere to file an issue for this, so I did and we'll move on. [https://github.com/microsoft/sarif-visualstudio-extension/issues/315](https://github.com/microsoft/sarif-visualstudio-extension/issues/315), "Decide where to display baseline state."

---
In reply to: [538916975](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r538916975) [](ancestors = 538916975,537693891)

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
lgolding: I see your point. The `Instance` property creates an instance of `SarifSuppressesResultsTableDataSource`, and that class's ctor _registers_ it with the table manager. I could instead just define `SarifSuppressesResultsTableDataSource.Register`, which would create the instance (again, registering the source), but would not return it. Ok, will do.

---
In reply to: [538919013](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r538919013) [](ancestors = 538919013)

src/Sarif.Viewer.VisualStudio/ErrorList/SarifAbsentResultsTableDataSource.cs
lgolding: Ok, good advice. I will look at this tomorrow.

---
In reply to: [538923639](https://github.com/microsoft/sarif-visualstudio-extension/pull/313#discussion_r538923639) [](ancestors = 538923639,538922382)

GabeDeBacker (approved):
:shipit:

michaelcfanning (approved):
:shipit:

PR #317, opened 2020-12-14
Removing duplicated analysis

eddynaka:
With this change when we do:
1. solution / project analysis
2. open a file that contains one error

it will:
remove the solution / project analysis and add the file open analysis. With that, we will prevent duplication when opening single file


michaelcfanning:
>                     ToImmutableList();

Should this be a set?

---
Refers to: src/Sarif.Viewer.VisualStudio/ErrorList/SarifTableDataSource.cs:124 in 8236e1f. [](commit_id = 8236e1fac89d87a3de2478a003e77264d93d0749, deletion_comment = False)

michaelcfanning (approved):
:shipit:

eddynaka:
>                     ToImmutableList();

yes and no...

the reason we have an IImultableList is that the method `sink.RemoveEntries` requires an `IReadOnlyList`, and then, it calls `this.tableDataSink.RemoveEntries` which also requires `IReadOnlyList`. The tableDataSink is from `Microsoft.VisualStudio.Shell.TableManager` which does not provide any overload to accept hashsets.

---
In reply to: [749186309](https://github.com/microsoft/sarif-visualstudio-extension/pull/317#issuecomment-749186309) [](ancestors = 749186309)

---
Refers to: src/Sarif.Viewer.VisualStudio/ErrorList/SarifTableDataSource.cs:124 in 8236e1f. [](commit_id = 8236e1fac89d87a3de2478a003e77264d93d0749, deletion_comment = False)

PR #327, opened 2020-12-23
Fixing annoying infobar and spam search

### LINKED ISSUE #303: Info bar feedback too chatty for background analysis
When a client sends a SARIF log to the viewer _via_ the viewer's interop API, and the log contains no results, the viewer displays an info bar with the message "no results". This is fine for the scenario where the client has performed an operation (such as a build) in the background, and the client wants the user to know that the operation is now complete (even if nothing has shown up in the Error List).

This UI doesn't work as well in the background analysis scenario implemented in the Sarifer extension. If the user has a solution open and opens a source file, Sarifer kicks off a background analysis of that file. In the common case where the file produces no results, the "no results" info bar appears. Because the viewer keeps track of which info bars are visible, there is never more than one such info bar -- but because most files _won't_ produce results, that info bar is almost always visible (in my experience), and if I close it, it just pops up again the next time I open a file.

@michaelcfanning @eddynaka

eddynaka:
Fixes #303

## Changes:
- Sarifer will only call InfoBar if it analyzed something. IF we don't have rules, we won't show the InfoBar
- Sarifer will search for all directories for .json files


michaelcfanning (approved):
:shipit:

PR #334, opened 2021-01-02
Fixing wide project/solution analysis

### LINKED ISSUE #333: Run Analysis isn't looking for all files from a project/solution

eddynaka:
Fixes #333

Before this change, the "Run Static Analysis" wasn't finding the txt files that we had inside a project, inside a folder.

With this change, it will look at all files and folder referenced by the project itself.

### DIFF
--- a/src/Sarif.Sarifer/AnalyzeProjectCommand.cs
+++ b/src/Sarif.Sarifer/AnalyzeProjectCommand.cs
@@ -79,10 +79,49 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer
                 {
                     // Disable the menu click when we are analysing.
                     SariferPackageCommand.DisableAnalyzeCommands(this.menuCommandService);
-                    List<string> targetFiles = project.GetMemberFiles();
+
+                    List<string> targetFiles = GetFiles(project);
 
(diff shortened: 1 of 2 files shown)


src/Sarif.Sarifer/AnalyzeProjectCommand.cs
```diff
@@ -79,14 +79,53 @@ private void MenuCommandCallback(object caller, EventArgs args)
...
+
+                    List<string> targetFiles = GetFiles(project);
 
                     this.backgroundAnalysisService.AnalyzeAsync(project.FullName, targetFiles, this.cancellationTokenSource.Token)
                         .FileAndForget(FileAndForgetEventName.BackgroundAnalysisFailure);
                 }
             }
         }
 
+        private static List<string> GetFiles(Project project)
```
michaelcfanning, line 91: >static List [](start = 16, length = 11)

are these helpers identical between the two modified files? if so, use a single version only? you could leave the helpers in a single file, mark them as internal, and then have the other file call into them. ?

michaelcfanning (approved):
:shipit:

src/Sarif.Sarifer/AnalyzeProjectCommand.cs
eddynaka: i can create a follow up merging those! thank you

PR #350, opened 2021-01-21
Enable NetAnalyzer

eddynaka:
## Changes
- Enable NETAnalyzers
- Fix a few error / warning / messages
- Remove FxCop (the new NETAnayzer deprecates FxCop)

## Why
- From our security rules, we should use Roslyn analyzers in IDE/build

### DIFF
--- a/src/Sarif.Viewer.VisualStudio.sln
+++ b/src/Sarif.Viewer.VisualStudio.sln
@@ -10,6 +10,7 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{2D702C24-103A-4561-8D86-62DBB2831900}"
 	ProjectSection(SolutionItems) = preProject
 		.editorconfig = .editorconfig
+		.ruleset = .ruleset
 		build.props = build.props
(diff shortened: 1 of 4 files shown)


src/Sarif.Viewer.VisualStudio.sln
```diff
@@ -10,6 +10,7 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{2D702C24-103A-4561-8D86-62DBB2831900}"
 	ProjectSection(SolutionItems) = preProject
 		.editorconfig = .editorconfig
+		.ruleset = .ruleset
```
yongyan-gh, line 13: should we specify right path for .ruleset and stylecop.json? they are not in root nor src folder

eddynaka: those files does not exist yet. so, we should add when we finish the fixes.

---
In reply to: [562226502](https://github.com/microsoft/sarif-visualstudio-extension/pull/350#discussion_r562226502) [](ancestors = 562226502)

yongyan-gh (approved):
:shipit:

PR #367, opened 2021-02-05
Try to resolve SARIF result file path using solution directory

yongyan-gh:
If current solution is opened. 
If unique file is found, remap path to this file.
If more than 1 files found, will skip to next manual remapping.


yongyan-gh:
Was trying to traverse Visual Studio solution/project object model (Solution/ProjectItem) to search file, but the performance is bad. E.g. it took 4-5 seconds to traverse all files in Sarif-Sdk solution. So changed to use file system searching.

eddynaka (approved):

Issue #374, opened 2021-02-25
SARIF viewer show partial message if the message includes long hyper link


yongyan-gh:
If the message contains hyper link with very long uri, viewer shows part of full messages.

PR #375, opened 2021-02-25
Fix partial message issue when there is long hyper link url in message.

yongyan-gh:
Fix for issue #374 

The issue caused by the inline texts (rich text format with hyperlink) are calculated by Message property of SarifErrorListItem, not the raw message. 

The fix is to use raw message to generate inline texts.


eddynaka:
Can you add a print screen before and after?

yongyan-gh:
Yes.

* This is the example sarif log opened by current version viewer. The result with hyper link only shows partial message (in 6th row)
image

* with this change, the result now shows full message texts with links.
image

eddynaka (approved):

Issue #336, opened 2021-01-02
SARIF files don't reload error list automatically when touched.


michaelcfanning:
Currently you need to close a SARIF log and reload to see refreshed results. There may be a file watcher mechanism in VS we can use to make this happen automatically. Project files reload automatically, for example.

Issue #318, opened 2020-12-16
SARIF viewer doesn't update automatically when log files are edited outside IDE


michaelcfanning:
Repro:

1) Generate a SARIF log file in an external process.
2) Open the SARIF log file in the SARIF viewer (results go to error list)
3) Generate a new SARIF log (with different results) in the external process

Result: log file must be closed and opened again before results refresh. It'd be nicer if the IDE determined the file had changed and reloaded.

@eddynaka
@yongyan-gh

@GabeDeBacker may have some useful information on file watching/refresh within the IDE

PR #378, opened 2021-03-04
Refresh Error List automatically if Sarif Log is changed

yongyan-gh:
Fix for #336 #318

Added FileSystemWatcher for each opened Sarif log file, and update Error list items when change is detected.

When Sarif log file is oped in VS text editor and updated/saved in VS, it wont fire file updated event.
Instead it saves file content to a new temp file and rename current file to another temp file, then rename 1st temp file with latest content to current file name and delete 2nd temp file.

So we need to catch the rename event for changes made in VS.

Will add corresponding Apex UI automation test cases.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/FileWatcher/SarifLogsMonitor.cs
+++ b/src/Sarif.Viewer.VisualStudio/FileWatcher/SarifLogsMonitor.cs
@@ -19,7 +19,8 @@ namespace Microsoft.Sarif.Viewer.FileWatcher
     {
         private readonly IFileSystem fileSystem;
 
-        private IDictionary<string, IFileWatcher> FileWatcherMap { get; } = new ConcurrentDictionary<string, IFileWatcher>();
+        private IDictionary<string, IFileWatcher> FileWatcherMap { get; } =
+            new ConcurrentDictionary<string, IFileWatcher>(StringComparer.OrdinalIgnoreCase);
 
         internal SarifLogsMonitor(IFileSystem fs)
         {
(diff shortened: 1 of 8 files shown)


src/Sarif.Viewer.VisualStudio/FileWatcher/SarifLogsMonitor.cs
```diff
@@ -0,0 +1,100 @@
...
+                FileWatcherMap.Add(logPath, watcher);
+                watcher.Start();
+            }
+        }
+
+        internal void Clear()
+        {
+            if (FileWatcherMap.Any())
+            {
+                FileWatcherMap.Values.ToList().ForEach(w => w.Stop());
```
eddynaka, line 53: >FileWatcherMap.Values.ToList().ForEach(w => w.Stop()); [](start = 16, length = 54)

You could replace to something like this:
foreach (IFileWatcher item in FileWatcherMap.Values)
                {
                    (item as IDisposable)?.Dispose();
                }

src/Sarif.Viewer.VisualStudio/FileWatcher/SarifLogsMonitor.cs
```diff
@@ -0,0 +1,100 @@
...
+        private void Watcher_SarifLogFileRenamed(object sender, System.IO.RenamedEventArgs e)
+        {
+            /*
+             * When updating a file in VS, it saves file content to a new temp file and rename current file to another temp file,
+             * then rename 1st temp file with latest content to current file name and delete 2nd temp file.
+             * Here we need to catch the event the 1st temp file is renamed to current file name
+             * and ignore event current file is renamed to 2nd temp file.
+             */
+            string filePath = e.FullPath;
+            if (FileWatcherMap.ContainsKey(filePath.ToLower()))
```
eddynaka, line 67: >filePath.ToLower() [](start = 43, length = 18)

in the FileWatcherMap, you could create using new Concurrent<>(StringComparer.xxxx), then remove the tolower()

src/Sarif.Viewer.VisualStudio/FileWatcher/SarifLogsMonitor.cs
yongyan-gh: Fixed

src/Sarif.Viewer.VisualStudio/FileWatcher/SarifLogsMonitor.cs
yongyan-gh: Fixed

eddynaka (approved):
:shipit:

Issue #380, opened 2021-03-12
Create file watcher to check .sarif folder to update errorList


eddynaka:
Feature:
- monitor a folder in the solution level called .sarif
- this folder will have n .sarif files and every time we see a change, we should update the error list
- if the file is deleted, we should remove the errors
- we should create a flag (just like the others) to check if we want to monitor or not

cc: @yongyan-gh @michaelcfanning

PR #381, opened 2021-03-17
Add FileWatcher to load sarif logs from .sarif folder automatically

yongyan-gh:
for #380 

- Automatically load all sarif log files from .sarif folder under solution directory to error lists when solution is loaded.
- Close all sarif logs when solution is closed.
- Monitor the folder and watch sarif log files created/deleted, load/unload results accordingly.
- Updates error list when any sarif log file in the folder is updated.
- Enabled/disable the behavior using option.


yongyan-gh:
Abandon this PR since the functionality should be added to sarif viewer not sarifer.
Created another PR #382 instead.

Issue #386, opened 2021-03-29
Sarifer - Analyze Solution throws null reference exception


yongyan-gh:
Repo: 
* Open a solution, e.g. SarifPatternMatcher.sln
* Right click solution and click "Run Static Analyzer" menu.
* Error pops up
![image](https://user-images.githubusercontent.com/76094515/112906821-b1ad6f80-90a1-11eb-9d77-7c340b0c3848.png)

PR #387, opened 2021-03-29
Fix null reference exception issue when analyze solution

yongyan-gh:
Fix for #386

Sarifer traverses all files of the solution tree using VS data model. Some special items e.g. items under solution items have null properties and caused the issue. Need to handle these items to avoid the null reference exception.

### DIFF
--- a/src/Sarif.Sarifer/Commands/SariferPackageCommand.cs
+++ b/src/Sarif.Sarifer/Commands/SariferPackageCommand.cs
@@ -60,7 +60,8 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer.Commands
             }
             else
             {
-                string localPath = projectItem.Properties.Item("LocalPath").Value.ToString();
+                // handle some solution items has null property. e.g. Solution Items, avoid null reference exception
+                string localPath = projectItem.Properties?.Item("LocalPath").Value.ToString();
                 if (!string.IsNullOrWhiteSpace(localPath))
                 {
                     targetFiles.Add(localPath);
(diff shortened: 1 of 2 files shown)


src/Sarif.Sarifer/Commands/SariferPackageCommand.cs
```diff
@@ -60,7 +60,8 @@ private static void GetFilesRecursive(List<string> targetFiles, ProjectItem proj
             }
             else
             {
-                string localPath = projectItem.Properties.Item("LocalPath").Value.ToString();
+                // handle some solution items has null property. e.g. Solution Items, avoid null reference exception
+                string localPath = projectItem.Properties?.Item("LocalPath").Value.ToString();
```
eddynaka, line 64: what is the properties type?

yongyan-gh: I think these are COM objects

projectItem.GetType().ToString()
"Microsoft.VisualStudio.ProjectSystem.VS.Implementation.Package.Automation.OAProjectItem"
projectItem.Properties.GetType().ToString()
"Microsoft.VisualStudio.ProjectSystem.VS.Implementation.Package.Automation.OAProperties"

---
In reply to: [603657849](https://github.com/microsoft/sarif-visualstudio-extension/pull/387#discussion_r603657849) [](ancestors = 603657849)

src/Sarif.Sarifer/Commands/SariferPackageCommand.cs
```diff
@@ -58,9 +58,24 @@ private static void GetFilesRecursive(List<string> targetFiles, ProjectItem proj
...
                 }
             }
+            else if (projectItem.Kind == Constants.vsProjectItemKindSolutionItems && projectItem.SubProject != null)
+            {
+                // this is for case solution has a folder at solution level and looking to its sub project
+                targetFiles.AddRange(GetFiles(projectItem.SubProject));
+            }
+            else if (projectItem.Kind == Constants.vsProjectItemKindVirtualFolder)
+            {
+                // this is for case a project has a virutal folder which contains a bunch of physical files
```
eddynaka, line 68: >virutal [](start = 52, length = 7)

virtual

eddynaka (approved):

PR #388, opened 2021-04-12
Fix issue cannot load SARIF log size greater than 5 MBs

### LINKED ISSUE #385: VS editor won't load JSON file >5 MB, so SARIF results don't load
The JSON editor doesn't like files > 5MB and won't load them. As a result, we won't show results in those files.

SARIF files can be very large due to things like embedded file content. If the file is too large, we should not try to load the file (or catch the exception so we can continue processing) but *should* try to process the SARIF by loading all results in the error list.

yongyan-gh:
Fix #385

Setup listener to base type "text" so that TextViewCreated event can be caught even Json editor throws error.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
+++ b/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
@@ -359,7 +359,7 @@ namespace Microsoft.Sarif.Viewer.ErrorList
 
         public static bool IsSarifLogOpened(string logFile)
         {
-            return SarifTableDataSource.Instance.HasErrors(logFile) ||
+            return SarifTableDataSource.Instance.HasErrorsFromLog(logFile) ||
                 CodeAnalysisResultManager.Instance.RunIndexToRunDataCache.
                     Any(runDataCacheKvp => runDataCacheKvp.Value.LogFilePath?.Equals(logFile, StringComparison.OrdinalIgnoreCase) == true);
         }
--- a/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
+++ b/src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
@@ -357,6 +357,13 @@ namespace Microsoft.Sarif.Viewer.ErrorList
             CleanAllErrors();
         }
 
+        public static bool IsSarifLogOpened(string logFile)
+        {
+            return SarifTableDataSource.Instance.HasErrors(logFile) ||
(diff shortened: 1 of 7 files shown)


src/Sarif.Viewer.VisualStudio/SarifTextViewCreationListener.cs
```diff
@@ -51,10 +52,21 @@ public void TextViewCreated(ITextView textView)
 
             if (this.TryGetFileNameFromTextView(textView, out string filename))
             {
+                if (!filename.EndsWith(".sarif", StringComparison.OrdinalIgnoreCase) &&
+                    !filename.EndsWith(".json", StringComparison.OrdinalIgnoreCase))
+                {
+                    // since Json (base type of sarif log) editor throws error when file size is greater than 5 MBs
+                    // need to listner to content type "text". Only process log if file extension is .sarif or .json
```
eddynaka, line 59: >listner [](start = 31, length = 7)

listener

eddynaka, line 59: >.json [](start = 111, length = 5)

add a period at the end of the sentence

src/Sarif.Viewer.VisualStudio/SarifTextViewCreationListener.cs
```diff
@@ -71,6 +83,14 @@ private void TextView_Closed(object sender, EventArgs e)
 
                 if (this.TryGetFileNameFromTextView(textView, out string filename))
                 {
+                    if (!filename.EndsWith(".sarif", StringComparison.OrdinalIgnoreCase) &&
+                        !filename.EndsWith(".json", StringComparison.OrdinalIgnoreCase))
+                    {
+                        // since Json (base type of sarif log) editor throws error when file size is greater than 5 MBs
+                        // need to listner to content type "text". Only process log if file extension is .sarif or .json
```
eddynaka, line 90: >listner [](start = 35, length = 7)

same comments above (listener and period at the end)

src/Sarif.Viewer.VisualStudio/ErrorList/ErrorListService.cs
```diff
@@ -357,6 +357,13 @@ public static void CloseAllSarifLogs()
             CleanAllErrors();
         }
 
+        public static bool IsSarifLogOpened(string logFile)
```
eddynaka, line 360: >IsSarifLogOpened [](start = 27, length = 16)

create a test adding dummy file and verifying.

eddynaka (approved):
:shipit:

PR #390, opened 2021-04-15
Fix issue when opening a .json file the checking logic invoked

yongyan-gh:
Fixes 2 issues:
- When opens .json file, the sarif file verification logic is invoked. Change to only process when file extension is ".sarif". 
- Handles a null reference in resolving source file.


eddynaka (approved):
:shipit:

PR #398, opened 2021-05-18
Fix an issue cannot load sarif tool execution notification

yongyan-gh:
Sarif viewer tries to resolve the tool execution notification location local file path when loading a sarif log contains tool notification, if it cannot resolve the path then it stops loading the log.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
+++ b/src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
@@ -710,7 +710,7 @@ namespace Microsoft.Sarif.Viewer
                 {
                     path = uri.LocalPath;
                 }
-                catch
+                catch (InvalidOperationException)
                 {
                     // if cannot resolve local path return original uri string
                     // it will try to resolve the path when user navigates to the error list item
--- a/src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
+++ b/src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
@@ -13,6 +13,7 @@ using System.Windows;
 using System.Windows.Forms;
 
 using Microsoft.CodeAnalysis.Sarif;
+using Microsoft.Sarif.Viewer.Sarif;
 using Microsoft.VisualStudio;
 using Microsoft.VisualStudio.Shell;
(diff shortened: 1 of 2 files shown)


src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
```diff
@@ -696,15 +697,25 @@ internal static string GetFileLocationPath(ArtifactLocation artifactLocation, in
...
                     Uri baseUri = dataCache.OriginalUriBasePaths[uriBaseId];
                     uri = new Uri(baseUri, uri);
                 }
 
-                path = uri.LocalPath;
+                try
+                {
+                    path = uri.LocalPath;
+                }
+                catch
```
eddynaka, line 713: >catch [](http://example.com/codeflow?start=16&length=5)

which exception are  we expecting?
can we prevent this checking somehow?

yongyan-gh: If the URI is not a local path e.g. "/example.cs", it throws InvalidOperation when access [LocalPath](https://docs.microsoft.com/en-us/dotnet/api/system.uri.localpath?view=netframework-4.7.2&f1url=%3FappId%3DDev16IDEF1%26l%3DEN-US%26k%3Dk(System.Uri.LocalPath);k(TargetFrameworkMoniker-.NETFramework,Version%253Dv4.7.2);k(DevLang-csharp)%26rd%3Dtrue)

I don't see a method similar to TryGet... or property to check if uri is a local path from Uri.

yongyan-gh: How about adding an extension method bool TryGetLocalPath(out string localPath) to handle this logic?

eddynaka: for now, lets not create the extension, since we just have this use-case.
Catch the specific exception and we will be good.

yongyan-gh: updated

---
In reply to: [634683700](https://github.com/microsoft/sarif-visualstudio-extension/pull/398#discussion_r634683700) [](ancestors = 634683700)

eddynaka (approved):

PR #399, opened 2021-05-19
Fix issue tooltip displays partial text for error list items in editor.

yongyan-gh:
If error list item message includes inline hyperlinks and its too long to be truncated, the tooltip when you move mouse over the selected text in editor shows partial message text.

The fix is to show the completed text. If raw text includes hyperlinks, convert to plain text first.


yongyan-gh:
Screenshot before the fix: tool tip text was truncated.

image


Screenshot after the change:

image

eddynaka (approved):
:shipit:

PR #411, opened 2021-07-09
Add output pane for Sarif viewer debug version.

yongyan-gh:
So we can print debug information into VS output window by using Trace.WriteLine().
Screenshot:

image

Pls let me know if any concern


eddynaka:
> So we can print debug information into VS output window by using Trace.WriteLine().
> Screenshot:
> 
> image
> 
> Pls let me know if any concern

I think we have something similar to the Sarifer, can you take a look?
If that's true, we can create only one class and pass the extension name to prevent code duplication.

PR #415, opened 2021-07-28
Convert Github file link to raw content link

yongyan-gh:
Convert Github file link to raw content link

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
+++ b/src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
@@ -762,5 +762,32 @@ namespace Microsoft.Sarif.Viewer
 
             return false;
         }
+
+        internal static string ConvertToGithubRawPath(string url)
+        {
+            // convert github file access page link to file raw content link
+            // from http://github.com/<username>/<repo>/<blob|tree>/<branch>/path/to/file
+            // to   http://raw.githubusercontent.com/<username>/<repo>/<branch>/path/to/file
+
+            if (string.IsNullOrWhiteSpace(url) ||
+                !(url.StartsWith("http://github.com") || url.StartsWith("https://github.com")))
+            {
+                return url;
+            }
+
+            Regex regex = new Regex("^(?<protocol>https?://)(?<site>github.com)/(?<user>.*?)/(?<repo>.*?)(?<folder>/tree/|/blob/)(?<path>.*?)$");
+
+            if (!regex.IsMatch(url))
+            {
+                return url;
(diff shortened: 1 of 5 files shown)


src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
```diff
@@ -762,5 +762,32 @@ internal static bool OpenExternalUrl(string uriString, int maxLength = 150)
 
             return false;
         }
+
+        internal static string ConvertToGithubRawPath(string url)
```
michaelcfanning, line 766: >ConvertToGithubRawPath [](http://example.com/codeflow?start=31&length=22)

Why are we linking to raw git hub content? how does that help customers?

why aren't we jumping to the github hosted content with the appropriate line selected?

michaelcfanning (changes requested):
:clock1:

src/Sarif.Viewer.VisualStudio/SdkUiUtilities.cs
yongyan-gh: this is the scenario that a result points to a github file url.
e.g.:
```
          "locations": [
            {
              "physicalLocation": {
                "artifactLocation": {
                  "uri": "https://github.com/microsoft/sarif-visualstudio-extension/blob/main/src/Sarif.Viewer.VisualStudio/Data/ruleLookup.json"
                },
```
if send http request to this url directly (https://github.com/microsoft/sarif-visualstudio-extension/blob/main/src/Sarif.Viewer.VisualStudio/Data/ruleLookup.json), the response are html texts, its actually a html page includes a file content.
image

if only want to get the file content, the url need to be converted to raw github url e.g.:
https://raw.githubusercontent.com/microsoft/sarif-visualstudio-extension/main/src/Sarif.Viewer.VisualStudio/Data/ruleLookup.json
this url returns raw file content in plain text format.

In the extension when user navigates to a result which location is a github file link, it downloads the file content and saves it to local temp folder and opens the local file in VS editor. This change is to try to download the raw file content and not the html file content.

yongyan-gh: This is an example .sarif file can demonstrate the issue.
If you navigate to the result, it opens a file named applicationsetting.json but content is html text.

michaelcfanning: Thanks for this explanation. I understand now.

michaelcfanning (approved):
:shipit:

PR #416, opened 2021-08-04
Try to resolve file location using version control provenance

yongyan-gh:
When it cannot resolve the file path using other information in the result e.g. base Uri id/solution folder etc., try to use VersionControlProvenance data to find the source file related to the results.
- If can find source file in VersionControlProvenance MappedTo location, use this location.
- If cannot, try to download the source file from repository according to VersionControlProvenance data.
- Save to temp folder and return the temp file path.

Can support Github repository now.
For ADO the right link can be generated but cannot download the file since needs of credentials.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio/VersionControlParser/GithubVersionControlParser.cs
+++ b/src/Sarif.Viewer.VisualStudio/VersionControlParser/GithubVersionControlParser.cs
@@ -1,0 +1,75 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Text.RegularExpressions;
+
+using Microsoft.CodeAnalysis.Sarif;
+using Microsoft.Sarif.Viewer.Sarif;
+
+namespace Microsoft.Sarif.Viewer
+{
+    internal class GithubVersionControlParser : IVersionControlParser
+    {
+        private readonly VersionControlDetails details;
+
+        internal GithubVersionControlParser(VersionControlDetails versionControl)
+        {
+            this.details = versionControl;
+        }
+
+        public Uri GetSourceFileUri(string relativeFilePath)
+        {
+            // github link format:
+            // https://github.com/<username>/<reponame>/<blob|tree>/<commitmentid|branch>/path/to/file
+
+            string sourceRelativePath = "blob/";
+            sourceRelativePath += this.details.Branch ?? this.details.RevisionId;
+            sourceRelativePath += "/";
+            sourceRelativePath += relativeFilePath;
+
+            if (Uri.TryCreate(this.details.RepositoryUri, sourceRelativePath, out Uri sourceUri) &&
+                sourceUri.IsHttpScheme())
+            {
+                return new Uri(ConvertToRawPath(sourceUri.ToString()));
+            }
+
+            return null;
+        }
+
+        public string ConvertToRawPath(string url)
+        {
+            // convert github file access page link to file raw content link
+            // from https://github.com/<username>/<repo>/<blob|tree>/<branch>/path/to/file
+            // to   https://raw.githubusercontent.com/<username>/<repo>/<branch>/path/to/file
+
+            if (string.IsNullOrWhiteSpace(url) ||
+                !(url.StartsWith("http://github.com") || url.StartsWith("https://github.com")))
+            {
+                return url;
+            }
+
+            var regex = new Regex("^(?<protocol>https?://)(?<site>github.com)/(?<user>.*?)/(?<repo>.*?)(?<folder>/tree/|/blob/)(?<path>.*?)$");
+
+            if (!regex.IsMatch(url))
+            {
+                return url;
+            }
+
+            return regex.Replace(url,
+                          m => m.Groups["protocol"].Value +
+                               "raw.githubusercontent.com" + "/" +
+                               m.Groups["user"].Value + "/" +
+                               m.Groups["repo"].Value + "/" +
+                               m.Groups["path"]);
+        }
+
+        public string GetLocalRelativePath(Uri uri, string relativeFilePath)
+        {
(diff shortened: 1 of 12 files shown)


src/Sarif.Viewer.VisualStudio/VersionControlParser/VersionControlParserFactory.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            // no corresponding parser found
+            parser = null;
+            return false;
+        }
+
+        internal static string ConvertToRawFileLink(string url)
+        {
+            if (!string.IsNullOrWhiteSpace(url) &&
+                (url.StartsWith($"http://{GithubHost}") || url.StartsWith($"https://{GithubHost}")))
```
eddynaka, line 45: >(url.StartsWith($"http://{GithubHost}") || url.StartsWith($"https://{GithubHost}")) [](http://example.com/codeflow?start=16&length=83)

this is strange

src/Sarif.Viewer.VisualStudio/VersionControlParser/GithubVersionControlParser.cs
```diff
@@ -0,0 +1,75 @@
...
+            if (Uri.TryCreate(this.details.RepositoryUri, sourceRelativePath, out Uri sourceUri) &&
+                sourceUri.IsHttpScheme())
+            {
+                return new Uri(ConvertToRawPath(sourceUri.ToString()));
+            }
+
+            return null;
+        }
+
+        public string ConvertToRawPath(string url)
```
eddynaka, line 40: >public string ConvertToRawPath(string url) [](http://example.com/codeflow?start=8&length=42)

make this static

src/Sarif.Viewer.VisualStudio/VersionControlParser/GithubVersionControlParser.cs
```diff
@@ -0,0 +1,75 @@
...
+            // from https://github.com/<username>/<repo>/<blob|tree>/<branch>/path/to/file
+            // to   https://raw.githubusercontent.com/<username>/<repo>/<branch>/path/to/file
+
+            if (string.IsNullOrWhiteSpace(url) ||
+                !(url.StartsWith("http://github.com") || url.StartsWith("https://github.com")))
+            {
+                return url;
+            }
+
+            var regex = new Regex("^(?<protocol>https?://)(?<site>github.com)/(?<user>.*?)/(?<repo>.*?)(?<folder>/tree/|/blob/)(?<path>.*?)$");
```
eddynaka, line 52: >Regex [](http://example.com/codeflow?start=28&length=5)

create a static instance regexoptions.compile, regexoptions.ignorecase

yongyan-gh: fixed

src/Sarif.Viewer.VisualStudio/VersionControlParser/VersionControlParserFactory.cs
yongyan-gh: changed to use static regex

src/Sarif.Viewer.VisualStudio/VersionControlParser/VersionControlParserFactory.cs
```diff
@@ -0,0 +1,67 @@
...
+            {
+                return new GithubVersionControlParser(null).ConvertToRawPath(url);
```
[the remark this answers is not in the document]
yongyan-gh: fixed

src/Sarif.Viewer.VisualStudio/VersionControlParser/GithubVersionControlParser.cs
yongyan-gh: I'd like make this method as a method of interface IVersionControlParser.
But C# only supports static interface method in version 8.0 (our version is 7.3). 

Another way is creating a static class member method instead of using interface method.

src/Sarif.Viewer.VisualStudio/VersionControlParser/VersionControlParserFactory.cs
```diff
@@ -0,0 +1,69 @@
...
+
+namespace Microsoft.Sarif.Viewer
+{
+    internal class VersionControlParserFactory
+    {
+        // known version control host
+        internal static string GithubHost = "github.com";
+        internal static string AdoHost = "dev.azure.com";
+
+        internal static Regex githubRegx = new Regex(@"^https?://github\.com", RegexOptions.IgnoreCase | RegexOptions.Compiled);
```
eddynaka, line 18: >githubRegx [](http://example.com/codeflow?start=30&length=10)

since you are not going to share besides this class:
1. make it private
2. for static, we normally use s_githubRegex (add the 'e'). Also apply the same thing to the s_adoRegex

src/Sarif.Viewer.VisualStudio/VersionControlParser/GithubVersionControlParser.cs
eddynaka: got it

yongyan-gh:
pls hold on i found an issue that in vs 2017 it cannot found the assembly System.Net.Http 4.2.0.0. I think it may caused by using HttpClient in my change.

eddynaka (approved):
:shipit:

eddynaka:
> pls hold on i found an issue that in vs 2017 it cannot found the assembly System.Net.Http 4.2.0.0. I think it may caused by using HttpClient in my change.

Were you able to fix this?

yongyan-gh:
> > pls hold on i found an issue that in vs 2017 it cannot found the assembly System.Net.Http 4.2.0.0. I think it may caused by using HttpClient in my change.
> 
> Were you able to fix this?

Yes I fixed the issue by replacing the reference with PackageReference (Nuget)

PR #422, opened 2021-08-30
Catch specific exceptions

yongyan-gh:


src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
```diff
@@ -744,9 +744,13 @@ internal bool TryResolveFilePathFromSolution(string solutionPath, string pathFro
                     return true;
                 }
             }
-            catch
+            catch (Exception ex) when (ex is ArgumentException ||
+                                       ex is IOException ||
+                                       ex is PathTooLongException ||
+                                       ex is UnauthorizedAccessException)
             {
                 // do not threw exception so that it keeps trying to resolve path the next
```
michaelcfanning, line 752: >threw [](http://example.com/codeflow?start=26&length=5)

'throw'

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
```diff
@@ -744,9 +744,13 @@ internal bool TryResolveFilePathFromSolution(string solutionPath, string pathFro
...
+                Trace.WriteLine($"{nameof(TryResolveFilePathFromSolution)} threw exception: {ex.Message}");
```
michaelcfanning, line 753: >Message [](http://example.com/codeflow?start=96&length=7)

replace 'ex.Message' with simply 'ex' (which is a synonym for ex.ToString(). If you do this, the generated string will contain not only the exception message but also the stack, which is usually essential data for debugging.

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
```diff
@@ -744,9 +744,13 @@ internal bool TryResolveFilePathFromSolution(string solutionPath, string pathFro
                     return true;
                 }
             }
-            catch
+            catch (Exception ex) when (ex is ArgumentException ||
+                                       ex is IOException ||
+                                       ex is PathTooLongException ||
```
michaelcfanning, line 749: >PathTooLongException [](http://example.com/codeflow?start=45&length=20)

PathTooLongException derives from IOException so you don't need to call it out on its own.

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
```diff
@@ -744,9 +744,13 @@ internal bool TryResolveFilePathFromSolution(string solutionPath, string pathFro
                     return true;
                 }
             }
-            catch
+            catch (Exception ex) when (ex is ArgumentException ||
```
michaelcfanning, line 747: >ArgumentException [](http://example.com/codeflow?start=45&length=17)

Generally, argument exception is not something you would swallow, as this indicates a condition a caller can check. There might not be a good helper to do so, though, so I don't know if my note is actionable.

michaelcfanning:
Approving with suggestions (i.e., please consider acting on them). :)

michaelcfanning (approved):
:shipit:

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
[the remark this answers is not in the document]
yongyan-gh: like 'or' instead of '||' here. but compiler complains 'or' is only available in C# 9.0

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
yongyan-gh: fixed the typo

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
yongyan-gh: removed.

src/Sarif.Viewer.VisualStudio/CodeAnalysisResultManager.cs
yongyan-gh: was thinking not print too many details to the user. we can add an option or --verbose switch to control the log behavior.

PR #431, opened 2021-09-17
Handle empty temporary solution path

yongyan-gh:
When VS opens a single file, not a folder or solution, it creates temporary solution "Solution1" and its FullName is empty in this case.

The change is to handle this case and not break following Path.GetDirectoryName().


src/Sarif.Viewer.VisualStudio.UnitTests/CodeAnalysisResultManagerTests.cs
```diff
@@ -679,6 +683,84 @@ public void CodeAnalysisResultManager_VerifyFileWithArtifactHash_HashDoesNotMatc
             this.numPrompts.Should().Be(1);
         }
 
+        [Fact]
+        public void CodeAnalysisResultManager_GetSolutionPath_SolutionFolderOpened()
+        {
+            string folder = @"C:\github\repo\myproject\";
```
eddynaka, line 689: >string folder [](http://example.com/codeflow?start=12&length=13)

mark all strings const

eddynaka (approved):
:shipit:

PR #436, opened 2021-11-01
Add functionality: suppress results in SARIF log file

yongyan-gh:
Allow VS extension users to suppress results in SARIF log file through:
- Gesture of error list
- Light bulb suppress action

Currently only allows suppression of Accepted status, does not support additional fields of suppression like justification/timestamp/expirydays etc.


src/Sarif.Viewer.VisualStudio.Core/Fixes/FixSuggestedActionsSource.cs
```diff
@@ -274,6 +277,29 @@ private IEnumerable<SuggestedActionSet> CreateActionSetFromErrors(IEnumerable<Sa
...
+            // If there are no actions, return null rather than an empty list. Otherwise VS will display
+            // a light bulb with no suggestions in its dropdown. This way, VS refrains from displaying
+            // the light bulb.
+            return suggestedActions.Any() ?
+                new List<SuggestedActionSet>
+                {
+                    new SuggestedActionSet(SuggestionCategory, suggestedActions),
+                }
+                :
+                null;
```
eddynaka, line 300: nit:

```csharp
return suggestedActions.Any()
                ? new List<SuggestedActionSet>{ new SuggestedActionSet(SuggestionCategory, suggestedActions) }
                : null;
```

This way, the conditions will be aligned.

src/Sarif.Viewer.VisualStudio.Core/CodeAnalysisResultManager.cs
```diff
@@ -683,6 +685,69 @@ internal SarifLog GetPartitionedLog(IEnumerable<SarifErrorListItem> listItems)
             return partitions[guid];
         }
 
+        internal void AddSuppressionToSarifLog(SuppressionModel suppressionModel)
+        {
+            if (suppressionModel == null || suppressionModel.SelectedErrorListItems?.Any() != true)
+            {
+                return;
+            }
```
eddynaka, line 693: you can apply the `?` to the first condition. Something like:
```
if  (suppressionModel?.SelectedErrorListItems?.Any() != true)
{
return;
}
```

src/Sarif.Viewer.VisualStudio.Core/CodeAnalysisResultManager.cs
```diff
@@ -683,6 +685,69 @@ internal SarifLog GetPartitionedLog(IEnumerable<SarifErrorListItem> listItems)
...
+
+            if (suppressionAdded)
+            {
+                // add empty suppression for results don't have suppression
+                // this is to satisfy sarif spec: either all results have non-null suppressions or have no suppressions
+                foreach (SarifErrorListItem errorListItem in dataCache.SarifErrors)
+                {
+                    if (errorListItem.SarifResult.Suppressions == null)
+                    {
+                        errorListItem.SarifResult.Suppressions = new Suppression[] { };
```
eddynaka, line 734: >new Suppression[] { }; [](http://example.com/codeflow?start=65&length=22)

check if you can do Array.Empty<Suppression>()

src/Sarif.Viewer.VisualStudio.Core/CodeAnalysisResultManager.cs
```diff
@@ -683,6 +685,69 @@ internal SarifLog GetPartitionedLog(IEnumerable<SarifErrorListItem> listItems)
...
+                !this.RunIndexToRunDataCache.TryGetValue(runIndex, out RunDataCache dataCache) ||
+                dataCache.SarifLog == null)
+            {
+                return;
+            }
+
+            if (suppressionAdded)
+            {
+                // add empty suppression for results don't have suppression
+                // this is to satisfy sarif spec: either all results have non-null suppressions or have no suppressions
```
eddynaka, line 729: i personally do not like this.
reason: we are adding  a lot of suppression tags for things that might not have.

imagine a file with 100 results and only one should be suppressed. then, we added 99 empty suppressions and 1 real suppression.

eddynaka: Also, add the link to the spec in the comments

src/Sarif.Viewer.VisualStudio.Core/Fixes/SuppressSuggestedAction.cs
```diff
@@ -0,0 +1,69 @@
...
+namespace Microsoft.Sarif.Viewer.Fixes
+{
+    internal class SuppressSuggestedAction : ISuggestedAction
+    {
+        private readonly SarifErrorListItem sarifErrorListItem;
+
+        public SuppressSuggestedAction(SarifErrorListItem errorListItem)
+        {
+            this.sarifErrorListItem = errorListItem;
+            this.DisplayText = $"Suppress issue {this.sarifErrorListItem.Rule.Id} in Sarif file";
```
eddynaka, line 23: >Sarif [](http://example.com/codeflow?start=85&length=5)

`SARIF file.`

src/Sarif.Viewer.VisualStudio.Core/Models/SuppressionModel.cs
```diff
@@ -0,0 +1,67 @@
...
+    internal class SuppressionModel : NotifyPropertyChangedObject
+    {
+        private SuppressionKind _kind = SuppressionKind.External;
+        private SuppressionStatus _status = SuppressionStatus.Accepted;
+        private IEnumerable<SarifErrorListItem> _sarifErrorListItems;
+
+        public SuppressionModel(IEnumerable<SarifErrorListItem> sarifErrorListItems)
+        {
+            this._sarifErrorListItems = sarifErrorListItems;
+        }
```
eddynaka, line 25: you could create another constructor that accepts only one sarifErrorListItem, with that, internally you could handle the way you want and update all the other places.

src/Sarif.Viewer.VisualStudio.Core/Models/SuppressionCollection.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.ObjectModel;
+
+namespace Microsoft.Sarif.Viewer.Models
+{
+    internal class SuppressionCollection : ObservableCollection<SuppressionModel>
+    {
+        public SuppressionCollection()
```
eddynaka, line 10: >SuppressionCollection [](http://example.com/codeflow?start=15&length=21)

do u need this?

eddynaka: I would also rename to SuppressionModelCollection

eddynaka: or remove this entire class and use like the others
var suppressions = new ObservableCollection<SuppressionModel>();

src/Sarif.Viewer.VisualStudio.Core/CodeAnalysisResultManager.cs
yongyan-gh: According to [spec](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317661), either all results have suppression or none of the results have suppression.

>The suppressions values for all result objects in theRun SHALL be either all null or all non-null.
NOTE: The rationale is that an engineering system will generally evaluate all results for suppression, or none of them. Requiring that the suppressions values be either all null or all non-null enables a consumer to determine whether suppression information is available for the run by examining a single result object.

Also the sdk Run.HasSuppressedResults() is based on same logic. vs extension use this API to check if hide suppressed results in error list by default

https://github.com/microsoft/sarif-sdk/blob/45e21b5671100f48807fb74e085342f196ce7513/src/Sarif/Core/Run.cs#L338-L341

Please let me know what you think

eddynaka: let's follow the spec and in all results.

src/Sarif.Viewer.VisualStudio.Core/Models/SuppressionCollection.cs
yongyan-gh: removed. it was original created for data binding in SARIF explorer suppression tab. since the design changed not needed any more

src/Sarif.Viewer.VisualStudio.Core/CodeAnalysisResultManager.cs
```diff
@@ -683,6 +685,72 @@ internal SarifLog GetPartitionedLog(IEnumerable<SarifErrorListItem> listItems)
             return partitions[guid];
         }
 
+        internal void AddSuppressionToSarifLog(SuppressionModel suppressionModel)
+        {
+            if (suppressionModel?.SelectedErrorListItems?.Any() != true)
```
shaopeng-gh, line 690: >.Any() != true [](http://example.com/codeflow?start=57&length=14)

question, do VS recommend !Any() or  Any() != true

src/Sarif.Viewer.VisualStudio.UnitTests/SuppressionTests.cs
```diff
@@ -0,0 +1,230 @@
...
+            {
+                await VerifySuppressionTests(testCase);
+            }
+        }
+
+        public async Task<SarifLog> VerifySuppressionTests(dynamic testCase)
+        {
+            DateTime now = DateTime.UtcNow;
+            string[] resultToBeSuppressed = testCase.ResultToBeSuppressed;
+            string sarifLogFilePath = @"E:\Users\Yong\Sarif\Logs\ResultsToSuppress.sarif";
```
shaopeng-gh, line 133: >Users\Yong [](http://example.com/codeflow?start=43&length=10)

remove your name in local path before checkin?

src/Sarif.Viewer.VisualStudio.Core/SarifCommandPackage.vsct
```diff
@@ -132,6 +132,14 @@
           <ButtonText>Clear all SARIF results</ButtonText>
         </Strings>
       </Button>
+      <Button guid="guidSarifErrorListCmdSet" id="cmdidSuppressResultCommand" priority="0x0100" type="Button">
+        <Parent guid="guidSarifErrorListCmdSet" id="ErrorListContextMenuGroup" />
+        <CommandFlag>DefaultInvisible</CommandFlag>
+        <CommandFlag>DynamicVisibility</CommandFlag>
+          <Strings>
+            <ButtonText>Suppress result in log file</ButtonText>
```
shaopeng-gh, line 140: >Suppress result in log file [](http://example.com/codeflow?start=24&length=27)

question, I see two kind of string
"Suppress result in log file"
$"Suppress issue {this.sarifErrorListItem.Rule.Id} in SARIF file"

should we also say in same format like
"Suppress issue in SARIF file"

src/Sarif.Viewer.VisualStudio.Core/CodeAnalysisResultManager.cs
yongyan-gh: this is because the '?' operator, the left side of the equation can be null or a boolean. cannot use '!' operator since it cannot be used for null value. but both null and boolean can be compared with another boolean.

src/Sarif.Viewer.VisualStudio.Core/SarifCommandPackage.vsct
yongyan-gh: Other menu items for error list use same term 'result', to be consistent with context, changed it to "Suppress result in SARIF file"

shaopeng-gh (approved):
:shipit:

PR #441, opened 2021-11-19
Fix randomly test failures

yongyan-gh:
2 test cases fail randomly fails in build pipeline. Never fail in local  debug run. Both tests create and write a file under %AppData% folder. When they fail, the file is not created without exception/error. not a timing issue.

The fix is to use mock FileSystem to avoid dealing the real file system in unit tests.


shaopeng-gh (approved):
:shipit:

PR #432, opened 2021-09-21
Add Support for both VS 2022 (preview) and VS 2017/2019

yongyan-gh:
In order to support latest Visual Studio 2022 we need to upgrade the VS SDK to 17.0, which doesn't compatible with pre-2022 VS. To support both VS 2022 and previous version of VS, refactor the projects:

- Create a shared project Sarif.Viewer.VisualStudio.Core.
- Move all C# code files into shared project.
- Update existing VSIX project Sarif.Viewer.VisualStudio to only build up to VS 2019 and update VS SDK reference to 16.0
- Add a new VSIX project Sarif.Viewer.VisualStudio.2022 which references to VS SDK 17.0
- Fix breaking API changes by using conditional compilation symbol: DEV16 / DEV17.
- Both 2 VSIX projects reference to shared project Sarif.Viewer.VisualStudio.Core.
- Both 2 VSIX projects build .vsix files, 1 for VS 2022 and another for previous version.

PR #443, opened 2021-11-24
Add VS 2022 support for Sarifer

yongyan-gh:
Similar to changes for VS 2022 support for Microsoft.Sarif.Viewer #432 

- Create a shared project Sarif.Sarifer.Core.
- Move all C# code files from original Sarifer project into shared project.
- Update existing VSIX project Sarif.Sarifer to reference VS SDK 16.0 and only build vsix supporting up to VS 2019.
- Add a new VSIX project Sarif.Sarifer.2022 which references to VS SDK 17.0 and build vsix supporting VS 2022


eddynaka (approved):
:shipit:

Issue #304, opened 2020-12-01
Add-in aggressively creates fix models on populating error list


michaelcfanning:
This significantly compromises performance when loading many results, all of which contain fixes. We should defer the fix model creation unless/until a file is loaded into the IDE.

PR #444, opened 2021-11-29
Defer loading fix models

yongyan-gh:
Instead loading fix models in SarifErrorListItem constructor, populate the fix models whenever the VS editor loads suggested actions (fixes).

related to #304


src/Sarif.Viewer.VisualStudio.UnitTests/ErrorList/SarifErrorListItemTests.cs
```diff
@@ -255,7 +255,7 @@ public void SarifErrorListItem_WhenFixHasRelativePath_UsesThatPath()
             };
 
             SarifErrorListItem item = MakeErrorListItem(result);
-
+            item.PopulateFixModelsIfNot();
```
eddynaka, line 258: >PopulateFixModelsIfNot [](http://example.com/codeflow?start=17&length=22)

you could check if you have fixes before and after.
i would expect that if we don't call the method, the fix wouldn't be loaded

src/Sarif.Viewer.VisualStudio.Core/Sarif/ArtifactChange.Extensions.cs
```diff
@@ -41,5 +41,23 @@ public static ArtifactChangeModel ToArtifactChangeModel(this ArtifactChange file
 
             return model;
         }
+
+        public static ArtifactChangeModel ToArtifactChangeModel(this ArtifactChange fileChange, IDictionary<string, Uri> originalUriBaseIds, FileRegionsCache fileRegionsCache)
```
eddynaka, line 45: >ToArtifactChangeModel [](http://example.com/codeflow?start=42&length=21)

pls, add some tests to cover this change

src/Sarif.Viewer.VisualStudio.Core/Sarif/Fix.Extensions.cs
```diff
@@ -30,6 +30,23 @@ public static FixModel ToFixModel(this Fix fix, IDictionary<string, ArtifactLoca
             return model;
         }
 
+        public static FixModel ToFixModel(this Fix fix, IDictionary<string, Uri> originalUriBaseIds, FileRegionsCache fileRegionsCache)
+        {
+            fix = fix ?? throw new ArgumentNullException(nameof(fix));
+
```
eddynaka, line 36: pls test:
1. null fix
2. fix.artifactchanges = null

src/Sarif.Viewer.VisualStudio.Core/Models/SarifErrorListItem.cs
```diff
@@ -150,14 +150,7 @@ public SarifErrorListItem(Run run, int runIndex, Result result, string logFilePa
...
 
-            if (result.Fixes != null)
-            {
-                foreach (Fix fix in result.Fixes)
-                {
-                    var fixModel = fix.ToFixModel(run.OriginalUriBaseIds, FileRegionsCache.Instance);
-                    this.Fixes.Add(fixModel);
-                }
-            }
+            // fixes are not populated in constructor.
```
eddynaka, line 153: >// fixes are not populated in constructor [](http://example.com/codeflow?start=12&length=41)

this comment should not exist.
this comment should be a codeflow comment

src/Sarif.Viewer.VisualStudio.Core/Sarif/ArtifactChange.Extensions.cs
```diff
@@ -41,5 +41,28 @@ public static ArtifactChangeModel ToArtifactChangeModel(this ArtifactChange file
...
+                // convert IDictionary<string, Uri> to IDictionary<string, ArtifactLocation> which
+                // is needed by Sarif.SDK extension method ArtifactLocation.TryReconstructAbsoluteUri()
+                uriToArtifactLocationMap = new Dictionary<string, ArtifactLocation>(originalUriBaseIds.Count);
+                foreach (KeyValuePair<string, Uri> entry in originalUriBaseIds)
+                {
+                    uriToArtifactLocationMap.Add(entry.Key, new ArtifactLocation { Uri = entry.Value });
+                }
+            }
+
+            return fileChange.ToArtifactChangeModel(uriToArtifactLocationMap, fileRegionsCache);
```
eddynaka, line 65: >fileChange.ToArtifactChangeModel(uriToArtifactLocationMap, fileRegionsCache) [](http://example.com/codeflow?start=19&length=76)

there is a possibility where originalUriBaseIds == null, with that, uriToArtifactLocationMap would be null.

Can you check if fileChange.ToArtifactChangeModel won't throw an exception, pls?

yongyan-gh: yes fileChange.ToArtifactChangeModel() allows null Dictionary<string, ArtifactLocation> as parameter and code handles that and will not throw exception.

so the change here allows null Dictionary<string, Uri> as parameter of the overload method, only add values if originalUriBaseIds is not null.

eddynaka (approved):
:shipit:

PR #446, opened 2021-12-08
Defer loading additional properties

yongyan-gh:
The properties of SARIF results: Locations/RelatedLocations/Stacks/CodeFlows(AnalysisSteps)/PropertyBags are not needed when loading SARIF results into Error List. They can be populated when the data are needed, like before SARIF explorer display result details.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio.Core/Models/SarifErrorListItem.cs
+++ b/src/Sarif.Viewer.VisualStudio.Core/Models/SarifErrorListItem.cs
@@ -109,57 +109,6 @@ namespace Microsoft.Sarif.Viewer
                 this.LineNumber = this.Region.StartLine;
                 this.ColumnNumber = this.Region.StartColumn;
             }
-
-            if (result.Locations?.Any() == true)
-            {
-                // Adding in reverse order will make them display in the correct order in the UI.
-                for (int i = result.Locations.Count - 1; i >= 0; --i)
-                {
-                    this.Locations.Add(result.Locations[i].ToLocationModel(run, resultId: this.ResultId, runIndex: this.RunIndex));
-                }
-            }
-
-            if (result.RelatedLocations?.Any() == true)
-            {
-                for (int i = result.RelatedLocations.Count - 1; i >= 0; --i)
-                {
-                    this.RelatedLocations.Add(result.RelatedLocations[i].ToLocationModel(run, resultId: this.ResultId, runIndex: this.RunIndex));
-                }
-            }
-
-            if (result.CodeFlows != null)
-            {
-                foreach (CodeFlow codeFlow in result.CodeFlows)
-                {
-                    var analysisStep = codeFlow.ToAnalysisStep(run, resultId: this.ResultId, runIndex: this.RunIndex);
-                    if (analysisStep != null)
-                    {
-                        this.AnalysisSteps.Add(analysisStep);
-                    }
-                }
-
-                this.AnalysisSteps.Verbosity = 100;
-                this.AnalysisSteps.IntelligentExpand();
-            }
-
-            if (result.Stacks != null)
-            {
-                foreach (Stack stack in result.Stacks)
-                {
-                    this.Stacks.Add(stack.ToStackCollection(resultId: this.ResultId, runIndex: this.RunIndex));
-                }
-            }
-
-            if (result.PropertyNames?.Any() == true)
-            {
-                foreach (string propertyName in result.PropertyNames)
-                {
-                    this.Properties.Add(
-                        new KeyValuePair<string, string>(
-                            propertyName,
-                            result.GetSerializedPropertyValue(propertyName)));
-                }
-            }
         }
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


src/Sarif.Viewer.VisualStudio.Core/Models/SarifErrorListItem.cs
```diff
@@ -391,7 +340,11 @@ public string SelectedTab
         public ObservableCollection<KeyValuePair<string, string>> Properties { get; }
 
         [Browsable(false)]
-        public bool HasDetails => this.Locations.Any() || this.RelatedLocations.Any() || this.AnalysisSteps.Any() || this.Stacks.Any() || this.Fixes.Any();
+        public bool HasDetails => this.SarifResult.Locations?.Any() == true ||
+                                  this.SarifResult.RelatedLocations?.Any() == true ||
+                                  this.SarifResult.CodeFlows?.Any() == true ||
+                                  this.SarifResult.Stacks?.Any() == true ||
+                                  this.SarifResult.Fixes?.Any() == true;
```
eddynaka, line 347: Order by length

src/Sarif.Viewer.VisualStudio.Core/Sarif/CodeFlow.Extensions.cs
```diff
@@ -39,7 +39,7 @@ public static LocationCollection ToThreadFlowLocationCollection(this CodeFlow co
 
         public static AnalysisStep ToAnalysisStep(this CodeFlow codeFlow, Run run, int resultId, int runIndex)
         {
-            ThreadHelper.ThrowIfNotOnUIThread();
+            // ThreadHelper.ThrowIfNotOnUIThread();
```
eddynaka, line 42: If not useful, remove it. Also remove the blank line.

src/Sarif.Viewer.VisualStudio.Core/Sarif/StackFrame.Extensions.cs
```diff
@@ -12,7 +12,7 @@ public static StackFrameModel ToStackFrameModel(this StackFrame stackFrame, int
         {
             var model = new StackFrameModel(resultId, runIndex);
 
-            model.FullyQualifiedLogicalName = stackFrame.Location.LogicalLocation.FullyQualifiedName;
+            model.FullyQualifiedLogicalName = stackFrame.Location?.LogicalLocation?.FullyQualifiedName;
```
eddynaka, line 15: Do we have a test covering this nullable parts? If not, pls, add.

eddynaka (changes requested):

src/Sarif.Viewer.VisualStudio.Core/Sarif/StackFrame.Extensions.cs
yongyan-gh: found the issue by new added SarifErrorListItemTests, also added unit tests for StackFrameExtensions.ToStackFrameModel itself.

eddynaka (approved):
:shipit:

src/Sarif.Viewer.VisualStudio.Core/Sarif/CodeFlow.Extensions.cs
michaelcfanning: Why was this removed exactly?

src/Sarif.Viewer.VisualStudio.UnitTests/Sarif/StackFrameExtensionsTests.cs
```diff
@@ -0,0 +1,99 @@
...
+            var stackFrameModel = stackFrame.ToStackFrameModel(resultId, runIndex);
+            stackFrameModel.Should().NotBeNull();
+            stackFrameModel.ResultId.Should().Be(resultId);
+            stackFrameModel.RunIndex.Should().Be(runIndex);
+            stackFrameModel.Module.Should().BeNull();
+            stackFrameModel.Message.Should().BeEquivalentTo("stack frame location 1");
+            stackFrameModel.FullyQualifiedLogicalName.Should().BeEquivalentTo(@"\root\FQDN\path-to\file.ext");
+            stackFrameModel.FilePath.Should().BeEquivalentTo("path/to/file.cpp");
+            stackFrameModel.Region.Should().NotBeNull();
+            stackFrameModel.Region.StartLine.Should().Be(10);
```
michaelcfanning, line 90: >StartLine [](http://example.com/codeflow?start=35&length=9)

This test code is not very maintainable, as it creates multiple copies of literals. You could have cleaned this up to some degree by consulting values from the stackFrame object itself, which you retained. 

stackFrameModel.Region.Startlein.Should().Be(stackFrame.Location.PhysicalLocation.Region.StartLine).

In general, when you find yourself repeating hard-coded strings and values, like your relative path @"\root\", etc., that's a clue to try to reduce the code to a single copy, i.e., authoring the code where the literal value only appears once.

This is a basic quality standard that we enforce through-out our codebase (and which is very common in many projects).

yongyan-gh: thanks for the feedback, refactor the test case and move the verification into a common function which does not depends on hardcoded values. please review pr: [https://github.com/microsoft/sarif-visualstudio-extension/pull/448](https://github.com/microsoft/sarif-visualstudio-extension/pull/448)

src/Sarif.Viewer.VisualStudio.Core/Sarif/CodeFlow.Extensions.cs
yongyan-gh: this function does not invoke anything related to UI elements, so its not required.

Issue #130, opened 2020-04-15
SARIF add-in opens binary files without prompting on double-click


michaelcfanning:
Repro:

* Create a SARIF file with a result that points to a binary file on disk, such as a nupkg or .dll. 
* Open the log file in VS, double-click the result.

The viewer finds the file and forces it to load in VS. It may be better if we pass the file extension through a MIME type helper to determine whether it's textual. If not, we could prompt the user for the extension and ask for confirmation to open what is potentially a binary file. We should cache the answer per extension/per session.

@lgolding

PR #451, opened 2021-12-20
Prompt on navigating to a SARIF result artifact is binary file

yongyan-gh:
# Description
When the Viewer navigate to a result, it opens the result's target file in VS editor. If the target file is a binary file, the viewer also tries to open it without confirming with user.  We should prompt the users to confirm if they really want to open the binary file in VS editor. Related issue #130.

# Fixes:
- When user navigates to a SARIF result which artifact is a binary file, popup a dialog asking if to open binary file. 
  Screenshot:
image

- Option to remember user's preference

- Check if a file is a binary file by validating if any NUL character in first N characters of the file. Similar to binary file check in git: similar check by git 
  https://github.com/git/git/blob/9c9b961d7eb15fb583a2a812088713a68a85f1c0/xdiff-interface.c#L187-L193


src/Sarif.Viewer.VisualStudio.Core/CodeAnalysisResultManager.cs
```diff
@@ -45,12 +45,14 @@ internal sealed class CodeAnalysisResultManager : IVsSolutionEvents
...
         internal const int S_OK = 0;
         private const string AllowedDownloadHostsFileName = "AllowedDownloadHosts.json";
+        private const string AllowedFileExtensionsFileName = "AllowedFileExtensions.json";
         private const string TemporaryFileDirectoryName = "SarifViewer";
         private readonly string temporaryFilePath;
 
         // Cookie for registration and unregistration
         private uint m_solutionEventsCookie;
         private readonly List<string> _allowedDownloadHosts;
+        private readonly List<string> _allowedFileExtensions;
```
eddynaka, line 55: >List [](http://example.com/codeflow?start=25&length=4)

this could be a hashset

src/Sarif.Viewer.VisualStudio.Core/CodeAnalysisResultManager.cs
```diff
@@ -77,6 +79,7 @@ internal sealed class CodeAnalysisResultManager : IVsSolutionEvents
             this._promptForEmbeddedFileDelegate = promptForEmbeddedFileDelegate ?? this.PromptForEmbeddedFile;
 
             this._allowedDownloadHosts = SdkUIUtilities.GetStoredObject<List<string>>(AllowedDownloadHostsFileName) ?? new List<string>();
+            this._allowedFileExtensions = SdkUIUtilities.GetStoredObject<List<string>>(AllowedFileExtensionsFileName) ?? new List<string>();
```
eddynaka, line 82: >List [](http://example.com/codeflow?start=125&length=4)

if you change to hashset, you could ignorecase to prevent .pdb and .PDB as two different extensions

src/Sarif.Viewer.VisualStudio.Core/SdkUiUtilities.cs
```diff
@@ -141,6 +142,18 @@ internal static IVsWindowFrame OpenDocument(IServiceProvider provider, string fi
...
+            // prompt user if open binary file in editor
+            if (promptForBinFile)
+            {
+                using (FileStream stream = File.OpenRead(file))
+                {
+                    if (IsBinaryFile(stream) && !AllowOpenBinaryFile(file))
+                    {
+                        return null;
+                    }
+                }
```
eddynaka, line 154: you have a flag that you check if you want to prompt or not. How do u define this?

Also, here is what u are doing:
1. opening/reading the file
2. checking if that is a binary file
3. checking the extension and prompting the user.

shouldn't you do the checking + prompting first? If the user says `no`, so you dont need to open/check the binary, for example.

src/Sarif.Viewer.VisualStudio.Core/CodeAnalysisResultManager.cs
yongyan-gh: changed to hashset .also can have better searching performance.

src/Sarif.Viewer.VisualStudio.Core/SdkUiUtilities.cs
yongyan-gh: The logic is in method AllowOpenBinaryFile. In latest change I updated the flow now its:
1. Checking if file extension is already allowed.
    a. If allowed, return true.
    b. If not allowed, go to step 2.
2. Checking if file is binary file.
    a. If yes then go to step 3.
    b. If no return true.
3. Prompting user if open the binary in VS.

Here the reason why checking if the file is binary file before prompting user, is that we cannot determine whether a file is binary or not according to file extension. E.g. file with extension txt but content is binary, or a unknow type of file extension. Please let me know what do you think

src/Sarif.Viewer.VisualStudio.Core/Resources.resx
```diff
@@ -386,4 +386,14 @@
   <data name="ImportClangTidyFilter" xml:space="preserve">
     <value>Clang log files (*.yaml)|*.yaml</value>
   </data>
+  <data name="ConfirmOpenBinaryFileDialog_AlwaysAllowLabel" xml:space="preserve">
+    <value>Always allow open file with extension '{0}'</value>
+  </data>
+  <data name="ConfirmOpenBinaryFileDialog_Message" xml:space="preserve">
+    <value>The target file with extension '{0}' is a binary file.
+Do you want to open this file in the Visual Studio?</value>
```
EasyRhin0, line 394: >the [](http://example.com/codeflow?start=33&length=3)

Remove

EasyRhinoMSFT: Testing

EasyRhinoMSFT: test

src/Sarif.Viewer.VisualStudio.Core/Resources.resx
```diff
@@ -386,4 +386,14 @@
   <data name="ImportClangTidyFilter" xml:space="preserve">
     <value>Clang log files (*.yaml)|*.yaml</value>
   </data>
+  <data name="ConfirmOpenBinaryFileDialog_AlwaysAllowLabel" xml:space="preserve">
+    <value>Always allow open file with extension '{0}'</value>
+  </data>
+  <data name="ConfirmOpenBinaryFileDialog_Message" xml:space="preserve">
+    <value>The target file with extension '{0}' is a binary file.
+Do you want to open this file in Visual Studio?</value>
```
EasyRhinoMSFT, line 394: test

EasyRhinoMSFT (dismissed):
:clock1:

EasyRhinoMSFT:
Trying to delete...

---
In reply to: [857252923](https://github.com/microsoft/sarif-visualstudio-extension/pull/451#issuecomment-857252923) [](http://example.com/codeflow?ancestors=857252923,857252923)

src/Sarif.Viewer.VisualStudio.Core/CodeAnalysisResultManager.cs
eddynaka: You need to update the constructor and add the ignorecase option

src/Sarif.Viewer.VisualStudio.Core/CodeAnalysisResultManager.cs
```diff
@@ -1084,5 +1087,19 @@ internal static string GetSolutionPath(DTE2 dte, IVsFolderWorkspaceService works
 
             return solutionPath;
         }
+
+        internal void AddAllowedFileExtension(string fileExtension)
+        {
+            if (!this._allowedFileExtensions.Contains(fileExtension, StringComparer.OrdinalIgnoreCase))
```
eddynaka, line 1093: >StringComparer. [](http://example.com/codeflow?start=69&length=15)

once u change the constructor of the hashset, this won't be needed.

src/Sarif.Viewer.VisualStudio.Core/SdkUiUtilities.cs
```diff
@@ -762,5 +769,79 @@ internal static bool OpenExternalUrl(string uriString, int maxLength = 150)
...
+                    out alwaysAllow);
+
+            return result == MessageDialogCommand.Yes;
+        }
+
+        internal static bool AllowOpenBinaryFile(string filePath, IFileSystem filesystem = null, PromptOpenBinaryFileDelegate promptDelegate = null)
+        {
+            string fileExtension = Path.GetExtension(filePath);
+            HashSet<string> allowedFileExtensions = CodeAnalysisResultManager.Instance.GetAllowedFileExtensions();
+            bool allowed = allowedFileExtensions.Contains(fileExtension, StringComparer.OrdinalIgnoreCase);
```
eddynaka, line 819: >StringComparer. [](http://example.com/codeflow?start=73&length=15)

we will be able to remove once u change the constructor

src/Sarif.Viewer.VisualStudio.Core/CodeAnalysisResultManager.cs
yongyan-gh: When deserialize the hashset using `SdkUIUtilities.GetStoredObject()`, the comparer will be reset to `EqualityComparer<string>.Default`. I created a override method to deserialize and populate data to existing hashset instance with custom comparer. Please let me know what do you think.

src/Sarif.Viewer.VisualStudio.Core/CodeAnalysisResultManager.cs
```diff
@@ -1084,5 +1089,19 @@ internal static string GetSolutionPath(DTE2 dte, IVsFolderWorkspaceService works
 
             return solutionPath;
         }
+
+        internal void AddAllowedFileExtension(string fileExtension)
```
eddynaka, line 1093: >AddAllowedFileExtension [](http://example.com/codeflow?start=22&length=23)

can we test this and the following method?

yongyan-gh:
close the PR for now maybe re-visit later

src/Sarif.Viewer.VisualStudio.UnitTests/TestData/BinaryTestFiles/NonBinaries/csharpfile.cs
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace AnalysisTestProject1
+{
+    internal class Class1
+    {
+    }
+}
```
marmegh, line 9: What is the function of this empty class?

michaelcfanning:
I reopened this. This issue happens to be important for the binary analysis scenario, so I'd love to see us get this into a minimally functional condition. Can I propose that someone perhaps demo the current function of the proposed change? Via video or a conference call?

yongyan-gh:
> I reopened this. This issue happens to be important for the binary analysis scenario, so I'd love to see us get this into a minimally functional condition. Can I propose that someone perhaps demo the current function of the proposed change? Via video or a conference call?

Please see the demo gif below:

OpenBinaryFile

Before the change, certain binary files cannot be opened and will show the dialog below:

image

src/Sarif.Viewer.VisualStudio.Core/Resources.resx
```diff
@@ -386,4 +386,14 @@
   <data name="ImportClangTidyFilter" xml:space="preserve">
     <value>Clang log files (*.yaml)|*.yaml</value>
   </data>
+  <data name="ConfirmOpenBinaryFileDialog_AlwaysAllowLabel" xml:space="preserve">
+    <value>Always allow open file with extension '{0}'</value>
```
EasyRhinoMSFT, line 390: >allow open file [](http://example.com/codeflow?start=18&length=15)

open files

src/Sarif.Viewer.VisualStudio.Core/SdkUiUtilities.cs
```diff
@@ -106,6 +107,33 @@ internal static T GetStoredObject<T>(string storageFileName)
...
         }
 
+        /// <summary>
+        /// Reads the contents of an isolated storage file and populate to an specified object instance.
+        /// </summary>
+        /// <typeparam name="T">The type of the deserialized object.</typeparam>
+        /// <param name="storageFileName">The isolated storage file.</param>
+        /// <param name="instance">Instance to be populated.</param>
+        /// <returns>Object deserialized from isolated storage file if exists, otherwize the same instance.</returns>
+        internal static T GetStoredObject<T>(string storageFileName, T instance)
```
EasyRhinoMSFT, line 117: >GetStoredObject [](http://example.com/codeflow?start=26&length=15)

This is almost identical to the method above. Can we combine them?

src/Sarif.Viewer.VisualStudio.UnitTests/TestData/BinaryTestFiles/NonBinaries/csharpfile.cs
yongyan-gh: this is a test .cs file (non-binary file) along with the other binaries test files, they are used for unit tests for binary scenarios. The content should not matter, I just put test C# code to simulate a C# code file.

src/Sarif.Viewer.VisualStudio.Core/CodeAnalysisResultManager.cs
yongyan-gh: Yes, pls see the test: `GetStoredObject_PopuplateHashSet_WithCustomComparer`

src/Sarif.Viewer.VisualStudio.Core/Resources.resx
yongyan-gh: fixed

src/Sarif.Viewer.VisualStudio.Core/SdkUiUtilities.cs
yongyan-gh: Combined into 1 method.
The need for the `T instance` parameter is because `JsonConvert.DeserializeObject<T>(reader.ReadToEnd());` in original method  will return instance of T using default comparer.
In my case I need to deserialize a `HashSet<string>` with specified comparer `StringComparer.OrdinalIgnoreCase`, so I need to passed the instance already initialized and call `JsonConvert.PopulateObject(reader.ReadToEnd(), instance);` to deserialize.

src/Sarif.Viewer.VisualStudio.Core/SdkUiUtilities.cs
```diff
@@ -106,6 +107,33 @@ internal static T GetStoredObject<T>(string storageFileName)
             return null;
         }
 
+        /// <summary>
+        /// Reads the contents of an isolated storage file and populate to an specified object instance.
+        /// </summary>
+        /// <typeparam name="T">The type of the deserialized object.</typeparam>
+        /// <param name="storageFileName">The isolated storage file.</param>
+        /// <param name="instance">Instance to be populated.</param>
+        /// <returns>Object deserialized from isolated storage file if exists, otherwize the same instance.</returns>
```
EasyRhinoMSFT: >z [](http://example.com/codeflow?start=86&length=1)

s
:blush:

EasyRhinoMSFT (approved):
:shipit:

Issue #449, opened 2021-12-13
Sarif Viewer VS extension does not support codeflows


xi-liu-ds:
Hi,

We found that Sarif Viewer VS extension does not support `codeFlow` property in Sarif file. We are wondering do you have a timeline to support it? Thanks!

eddynaka:
@yongyan-gh, can you create a SARIF with CodeFlow properties, test in vscode, and see if that appears correctly in both tools?

xi-liu-ds:
I see. I have double checked. It seems that if drop the `physicalLocation` in codeflows, Sarif Viewer extension aren't able to load it? For example:

```
{
    "$schema": "http://json.schemastore.org/sarif-2.1.0-rtm.5",
    "version": "2.1.0",
    "runs": [
        {
            "tool": {
                "driver": {
                    "name": "Infer",
                    "informationUri": "https://github.com/facebook/infer",
                    "version": "1.1.0",
                    "rules": [
                        {
                            "id": "NULLPTR_DEREFERENCE",
                            "shortDescription": {
                                "text": "Null Dereference"
                            }
                        }
                    ]
                }
            },
            "results": [
                {
                    "message": {
                        "text": "Pulse found a potential null pointer dereference on line 41."
                    },
                    "level": "error",
                    "ruleId": "NULLPTR_DEREFERENCE",
                    "codeFlows": [
                        {
                            "threadFlows": [
                                {
                                    "locations": [
                                        {
                                            "nestingLevel": 1,
                                            "location": {
                                                "message": {
                                                    "text": "in call to `Object Program.ReturnNullReference()`"
                                                }
                                            }
                                        }
                                    ]
                                }
                            ]
                        }
                    ],
                    "locations": [
                        {
                            "physicalLocation": {
                                "artifactLocation": {
                                    "uri": "file:C:\\Users\\matjin\\source\\repos\\ConsoleApp4\\ConsoleApp4\\Program.cs",
                                    "uriBaseId": "/root/C:\\Users\\matjin\\source\\repos\\ConsoleApp4\\ConsoleApp4\\Program.cs"
                                },
                                "region": {
                                    "startLine": 41,
                                    "startColumn": 13
                                }
                            }
                        }
                    ]
                }
            ]
        }
    ]
}
```

yongyan-gh:
[...]

PR #452, opened 2022-01-11
Fix CodeFlows/ThreadFlows rendering issue

yongyan-gh:
# Description

The viewer fails to render code/thread flows in SARIF explorer in certain scenarios even it is a valid SARIF file, e.g. the first element of thread flow locations has nesting level does not start at 0. 
This is due to the code logic builds thread flow location list into parent/child hierarchy based on nesting level, it assumes first element should be root node and nesting level always starts with 0.

# Fixes

After discussing with team, the customers don't have a requirement to see the code/thread flows in parent/child hierarchy.
Also, to be consistent with VS code viewer's UX, decided to change current way rendering code/thread flows in tree view, to the way rendering them in flat list. And adding indents before the list item label to present the nesting level.

screenshot:
image

related to #449

### DIFF
--- a/src/Sarif.Viewer.VisualStudio.Core/CodeFlowToTreeConverter.cs
+++ b/src/Sarif.Viewer.VisualStudio.Core/CodeFlowToTreeConverter.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 
 using Microsoft.CodeAnalysis.Sarif;
 using Microsoft.Sarif.Viewer.Models;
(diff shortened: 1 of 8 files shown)


src/Sarif.Viewer.VisualStudio.Core/Converters/AnalysisStepNodeToTextConverter.cs
```diff
@@ -11,6 +11,8 @@ namespace Microsoft.Sarif.Viewer.Converters
 {
     internal class AnalysisStepNodeToTextConverter : IValueConverter
     {
+        internal static readonly char IndentChar = ' ';
```
marmegh, line 14: >internal static readonly char IndentChar = ' '; [](http://example.com/codeflow?start=8&length=47)

Have there been any accessibility considerations for the single space indent? Looking at the screenshot, I only registered the indents because I read the description and code.

yongyan-gh: updated the solution to set indent by setting left margin of the text. It can be more accurate to pixel level and doesn't affect text content itself. 

Updated the screenshot in description. Please take a look and let me know what do you think

marmegh: Awesome. So much easier for me to read. @EasyRhinoMSFT, can you take a look at this?

EasyRhinoMSFT: Looks good!

EasyRhin0: Looks good! Using a margin is the right way to go.

src/Sarif.Viewer.VisualStudio.Core/CodeFlowToTreeConverter.cs
```diff
@@ -68,5 +69,44 @@ internal static List<AnalysisStepNode> Convert(CodeFlow codeFlow, Run run, int r
...
+                // min nesting level can be used to offset nesting level starts from value greater than 0.e
+                int minNestingLevel = threadFlow.Locations.Min(l => l.NestingLevel);
+
+                foreach (ThreadFlowLocation location in threadFlow.Locations)
+                {
+                    ArtifactLocation artifactLocation = location.Location?.PhysicalLocation?.ArtifactLocation;
+
+                    if (artifactLocation != null)
+                    {
+                        if (artifactLocation.Uri == null && artifactLocation.Index > -1)
```
EasyRhin0, line 92: NIT: these could be collapsed, but I see the readability value of doing it this way.

src/Sarif.Viewer.VisualStudio.UnitTests/Models/AnalysisStepNodeTests.cs
```diff
@@ -11,6 +14,8 @@ namespace Microsoft.Sarif.Viewer.Models
 {
     public class AnalysisStepNodeTests
     {
+        private static readonly Random random = new Random();
```
shaopeng-gh, line 17: >random [](http://example.com/codeflow?start=39&length=6)

I remember Michael mention add s_ to static vars

EasyRhin0 (approved):
:shipit:

src/Sarif.Viewer.VisualStudio.UnitTests/Models/AnalysisStepNodeTests.cs
yongyan-gh: updated

src/Sarif.Viewer.VisualStudio.Core/CodeFlowToTreeConverter.cs
yongyan-gh: combined into one if condition

shaopeng-gh (approved):
:shipit:

PR #472, opened 2022-04-06
Fix bug not able to navigate to source file under solution

yongyan-gh:
# Description

When users open a Sarif log file with the solution opened in VS, they cannot navigate to the source file except the first result they double clicked. The inline hyper-links only work for the first result they clicked.

Below is the root cause analysis:
1. When first time navigate to an error list item (Sarif result), the extension tries to resolve the result's file location.
2. The matched file is found in the solution folder and the extension navigates to the file. It also caches the solution path in a remapping table.
3. If navigate to another result the extension checks if any existing path in the remapping table and try to use the cached path to find the matched file. 
The result's file path is in URI format "src/sarif-sdk/src/Sarif/Autogenerated/AddressEqualityComparer.cs"
The cached solution path is in file system format "C:\\repo\\github.com\\binskim"
And the combined path is "C:\\repo\\github.com\\binskim\\src/sarif-sdk/src/Sarif/Autogenerated/AddressEqualityComparer.cs", which causes issue in the following code.
4. Hyperlinks doesn't work due to the same issue.

# Fixes:
In the function `TryResolveFilePathFromRemappings` called in the step 3 above, change the file path to file system format before combines with cached path.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio.UnitTests/CodeAnalysisResultManagerTests.cs
+++ b/src/Sarif.Viewer.VisualStudio.UnitTests/CodeAnalysisResultManagerTests.cs
@@ -277,6 +277,39 @@ namespace Microsoft.Sarif.Viewer.VisualStudio.UnitTests
         }
 
         [Fact]
+        public void CodeAnalysisResultManager_GetRebaselinedFileName_ConstructPathFromSolutionPathRemapping()
+        {
+            const int RunId = 1;
+            const string ResultArtifactPath1 = @"Sarif/Notes.cs";
+            const string ResultArtifactPath2 = @"Sarif/OffsetInfo.cs";
+            const string SolutionPath = @"D:\Users\John\source\sarif-sdk\src";
+            const string ExpectedResolvedPath1 = SolutionPath + @"\Sarif\Notes.cs";
+            const string ExpectedResolvedPath2 = SolutionPath + @"\Sarif\OffsetInfo.cs";
+            this.existingFiles.Add(ExpectedResolvedPath1);
+            this.existingFiles.Add(ExpectedResolvedPath2);
+
+            var target = new CodeAnalysisResultManager(
+                fileSystem: this.mockFileSystem.Object,
+                this.FakePromptForResolvedPath);
+            var dataCache = new RunDataCache();
+            target.RunIndexToRunDataCache.Add(RunId, dataCache);
+
+            // Mimic first time user navigate to a result, resolve file path using solution path.
+            // The resolved path should be cached for next resolving
+            target.SaveResolvedPathToUriBaseMapping(null, ResultArtifactPath1, ResultArtifactPath1, ExpectedResolvedPath1, dataCache);
+            Tuple<string, string>[] remappedPathPrefixes = target.GetRemappedPathPrefixes();
+            remappedPathPrefixes.Length.Should().Be(1);
+            remappedPathPrefixes[0].Item1.Should().Be("");
+            remappedPathPrefixes[0].Item2.Should().Be(SolutionPath);
+
+            string actualResolvedPath = target.GetRebaselinedFileName(
+                uriBaseId: null,
+                pathFromLogFile: ResultArtifactPath2,
+                dataCache: dataCache);
+            actualResolvedPath.Should().Be(ExpectedResolvedPath2);
+        }
+
+        [Fact]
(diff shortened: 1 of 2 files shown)


src/Sarif.Viewer.VisualStudio.UnitTests/CodeAnalysisResultManagerTests.cs
```diff
@@ -276,6 +276,39 @@ public void CodeAnalysisResultManager_GetRebaselinedFileName_RelativeUri_Without
...
 
+        [Fact]
+        public void CodeAnalysisResultManager_GetRebaselinedFileName_ConstructPathFromSolutionPathRemapping()
+        {
+            const int RunId = 1;
+            const string ResultArtifactPath1 = @"Sarif/Notes.cs";
+            const string ResultArtifactPath2 = @"Sarif/OffsetInfo.cs";
+            const string SolutionPath = @"D:\Users\John\source\sarif-sdk\src";
+            const string ExpectedResolvedPath1 = SolutionPath + @"\Sarif\Notes.cs";
+            const string ExpectedResolvedPath2 = SolutionPath + @"\Sarif\OffsetInfo.cs";
```
marmegh, line 287: Path.Combine would be preferable to concatenation.

marmegh:
Is there a release history that we can update?

EasyRhinoMSFT (approved):
:shipit:

src/Sarif.Viewer.VisualStudio.UnitTests/CodeAnalysisResultManagerTests.cs
yongyan-gh: fixed

yongyan-gh:
no we don't have an release history yet, we are using the automatically generated release note i think

---
In reply to: [1090711289](https://github.com/microsoft/sarif-visualstudio-extension/pull/472#issuecomment-1090711289) [](http://example.com/codeflow?ancestors=1090711289,1090711289)

eddynaka:
We actually have:
[https://github.com/microsoft/sarif-visualstudio-extension/blob/main/src/ReleaseHistory.md](https://github.com/microsoft/sarif-visualstudio-extension/blob/main/src/ReleaseHistory.md)

but it is stale.
we should start updating it.

---
In reply to: [1090805670](https://github.com/microsoft/sarif-visualstudio-extension/pull/472#issuecomment-1090805670) [](http://example.com/codeflow?ancestors=1090805670,1090711289)

yongyan-gh:
oh sorry i didn't see it. Will update it using the notes we have in release.

---
In reply to: [1090807100](https://github.com/microsoft/sarif-visualstudio-extension/pull/472#issuecomment-1090807100) [](http://example.com/codeflow?ancestors=1090807100,1090711289)

src/Sarif.Viewer.VisualStudio.UnitTests/CodeAnalysisResultManagerTests.cs
```diff
@@ -276,6 +276,39 @@ public void CodeAnalysisResultManager_GetRebaselinedFileName_RelativeUri_Without
             remappedPathPrefixes[0].Item2.Should().Be(@"D:\Users\John\source\sarif-sdk");
         }
 
+        [Fact]
+        public void CodeAnalysisResultManager_GetRebaselinedFileName_ConstructPathFromSolutionPathRemapping()
```
eddynaka, line 280: >CodeAnalysisResultManager_GetRebaselinedFileName_ConstructPathFromSolutionPathRemapping [](http://example.com/codeflow?start=20&length=87)

are you testing two functionalities at the same time? If yes, can you split the tests?

eddynaka: Can path be null? if yes, are we covering that? If no, we don't need the path?....in the normalize function that you created

michaelcfanning (approved):

src/Sarif.Viewer.VisualStudio.UnitTests/CodeAnalysisResultManagerTests.cs
yongyan-gh: its a like a sequence steps, the issue happens when construct file path using the cached solution path. So the first step need to setup a paths mapping. the second step to resolve the file path.

The path can be empty string, ideally it should not be null, it cannot enter this code. But we should have test to cover it will add test case

PR #477, opened 2022-04-21
Enable Dynamic Mode

marmegh:
This change includes, preventing null reference exception and adding dynamic mode validation. Please sync with me directly offline for testing directions.

### DIFF
--- a/src/Sarif.Sarifer.Core/BackgroundAnalyzerBase.cs
+++ b/src/Sarif.Sarifer.Core/BackgroundAnalyzerBase.cs
@@ -201,7 +201,7 @@ namespace Microsoft.CodeAnalysis.Sarif.Sarifer
                 dataToInsert: OptionallyEmittedData.ComprehensiveRegionProperties | OptionallyEmittedData.TextFiles | OptionallyEmittedData.VersionControlDetails,
                 dataToRemove: OptionallyEmittedData.None,
                 tool: this.MakeTool(),
-                levels: new List<FailureLevel> { FailureLevel.Error, FailureLevel.Warning, FailureLevel.Note, FailureLevel.None },
+                levels: new List<FailureLevel> { FailureLevel.Error }, // Options: FailureLevel.Warning, FailureLevel.Note, FailureLevel.None.
                 kinds: this.GetResultKinds(this.ExtensionOption),
                 closeWriterOnDispose: false);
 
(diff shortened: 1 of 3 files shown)


src/Sarif.Sarifer.Core/SpamBackgroundAnalyzer.cs
```diff
@@ -88,6 +88,7 @@ protected override bool AnalyzeCore(Uri uri, string text, string solutionDirecto
                 TargetUri = uri,
                 FileContents = text,
                 Logger = sarifLogger,
+                DynamicValidation = true, // Enable dynamic validations.
```
yongyan-gh, line 91: >DynamicValidation = true [](http://example.com/codeflow?start=16&length=24)

wondering how long will the dynamic validations take to get the result. Maybe we can let user to decide if to enable the dynamic validation through an option?

src/Sarif.Sarifer.Core/BackgroundAnalyzerBase.cs
```diff
@@ -201,7 +201,7 @@ private static async Task<string> GetSolutionDirectoryAsync()
                 dataToInsert: OptionallyEmittedData.ComprehensiveRegionProperties | OptionallyEmittedData.TextFiles | OptionallyEmittedData.VersionControlDetails,
                 dataToRemove: OptionallyEmittedData.None,
                 tool: this.MakeTool(),
-                levels: new List<FailureLevel> { FailureLevel.Error, FailureLevel.Warning, FailureLevel.Note, FailureLevel.None },
+                levels: new List<FailureLevel> { FailureLevel.Error }, // Options: FailureLevel.Warning, FailureLevel.Note, FailureLevel.None.
```
yongyan-gh, line 204: >// Options: FailureLevel.Warning, FailureLevel.Note, FailureLevel.None. [](http://example.com/codeflow?start=71&length=71)

why remove these levels?
if we need support Pass result, when kind = Pass, all levels will be Note. it will filter out these pass results

src/Sarif.Sarifer.Core/SpamBackgroundAnalyzer.cs
marmegh: Valid point. Any latency changes are negligible at the moment. The greatest impact is going to be number of open tabs.
Letting the user decide isn't the best idea given the intent of this change. Ideally, we would want that to be an environment variable of some sort that requires administrator privileges to modify. This effort now falls outside of what is likely achievable to get to MVP for this week.

yongyan-gh (approved):
:shipit:

EasyRhinoMSFT (approved):
:shipit:

PR #485, opened 2022-05-16
Ignoring component governance submodules

eddynaka:
This change will ignore any issues in the sarif-pattern-matcher submodules.
This will prevent component governance from generating double issues: one in the submodule and one in this repository.

No tests were added since no product code was changed.
No release change was updated since it does not affect customers.


marmegh (approved):
:shipit:

EasyRhinoMSFT (approved):

PR #494, opened 2022-07-05
GHAS service + unit tests

EasyRhinoMSFT:
This change introduces the GitHubAdvancedSecurity result source service. The platform, services, and supporting code uses the Functional Programming pattern. The following is a breakdown of the feature components and their function.

**ResultSources.ACL:** anti-corruption layer
* Secure secret storage repository. Uses Windows Credential Manager to store secrets such as GitHub access tokens

**ResultSources.Domain**
* Defines abstractions, entities, models, and base services/interfaces.
* Provides an HTTP helper to simplify requests

**ResultSources.Factory**
* Provides a factory that uses DI to create source service instances and check for compatibility (IResultSourceService.IsActive) with the current solution, wrapped by a host class
* This approach maintains loose coupling with the Viewer project

** Shell**
* Provides services to interact with VS UI components such as the status bar and InfoBars
* Provides Git service
* FileWatcher moved here from Viewer.FileWatcher

**GitHubAdvancedSecurityService**
* The first result source, connects to GHAS
* Uses the device code auth flow to avoid a secret in code.
* Updates use polling and are triggered by pushes and branch changes


src/Sarif.Viewer.VisualStudio.Core/SarifViewerPackage.cs
```diff
@@ -203,12 +215,60 @@ private void SolutionEvents_OnBeforeCloseSolution(object sender, EventArgs e)
...
             // stop watcher when the solution is closed.
             this.sarifFolderMonitor?.StopWatch();
+
+            var fileSystem = new FileSystem();
+
+            try
+            {
+                fileSystem.FileDelete(Path.Combine(GetDotSarifDirectoryPath(), "scan-results.sarif"));
+            }
+            catch (Exception) { }
```
shaopeng-gh, line 225: will it cause any issue if fail, 
suggest to add explanation.
e.g.
// Doesn't matter because this delete is best-effort.

src/Sarif.Viewer.VisualStudio.Core/SarifViewerPackage.cs
```diff
@@ -203,12 +215,60 @@ private void SolutionEvents_OnBeforeCloseSolution(object sender, EventArgs e)
         {
             // stop watcher when the solution is closed.
             this.sarifFolderMonitor?.StopWatch();
+
+            var fileSystem = new FileSystem();
+
+            try
+            {
+                fileSystem.FileDelete(Path.Combine(GetDotSarifDirectoryPath(), "scan-results.sarif"));
```
shaopeng-gh, line 223: >"scan-results.sarif" [](http://example.com/codeflow?start=79&length=20)

is this file name a default value that user can change, or is it fixed and always the same?

src/Sarif.Viewer.VisualStudio.Shell.Core/GitExe.cs
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        private async ValueTask<string> ExecuteGitCommandAsync(string arguments)
+        {
+            if (string.IsNullOrWhiteSpace(this.vsInstallDir))
+            {
+                this.vsInstallDir = await GetVsInstallDirectoryAsync();
+            }
+
+            // Get the trusted min Git executable path.
```
shaopeng-gh, line 66: >trusted min Git executable path [](http://example.com/codeflow?start=23&length=31)

sorry did not get what the min means here.

src/Sarif.Viewer.VisualStudio.Shell.Core/GitExe.cs
```diff
@@ -0,0 +1,108 @@
...
+            string minGitPath = Path.Combine(this.vsInstallDir, @"CommonExtensions\Microsoft\TeamFoundation\Team Explorer\Git\cmd\git.exe");
```
shaopeng-gh, line 67: >Path.Combine(this.vsInstallDir, @"CommonExtensions\Microsoft\TeamFoundation\Team Explorer\Git\cmd\git.exe") [](http://example.com/codeflow?start=32&length=107)

if user install git separately using official git website and does not install it with the VS installer will this have problem

shaopeng-gh: fyi I see there is a `git --exec-path` command can find where current working git location

src/Sarif.Viewer.VisualStudio.Core/SarifViewerPackage.cs
EasyRhinoMSFT: This is not configurable. We may need more sophisticated cleanup code in the future but for now this is fine.

src/Sarif.Viewer.VisualStudio.Shell.Core/GitExe.cs
EasyRhinoMSFT: Yes you're right. I have several installs on my machine but I haven't found a reliable way to get the path. I'll see if I can run it by just using "git.exe".

EasyRhinoMSFT: that works. I'll do some more digging to make sure it's reliable

src/Sarif.Viewer.VisualStudio.Core/SarifViewerPackage.cs
```diff
@@ -203,12 +215,60 @@ private void SolutionEvents_OnBeforeCloseSolution(object sender, EventArgs e)
...
+            await this.resultSourceService.RequestAnalysisScanResultsAsync();
+        }
+
+        private void ResultSourceService_ResultsUpdated(object sender, ResultsUpdatedEventArgs e)
+        {
+            string path = Path.Combine(GetDotSarifDirectoryPath(), e.LogFileName);
+            e.SarifLog.Save(path);
+        }
+
+        private static string GetSolutionDirectoryPath()
```
yongyan-gh, line 263: >GetSolutionDirectoryPath [](http://example.com/codeflow?start=30&length=24)

nit there are couple functions defined in different projects, we should put it in a common place

src/Sarif.Viewer.VisualStudio.Core/SarifViewerPackage.cs
```diff
@@ -203,12 +215,60 @@ private void SolutionEvents_OnBeforeCloseSolution(object sender, EventArgs e)
...
 
         private void SolutionEvents_OnAfterBackgroundSolutionLoadComplete(object sender, EventArgs e)
         {
             // start to watch when the solution is loaded.
             this.sarifFolderMonitor?.StartWatch();
+
+            this.JoinableTaskFactory.Run(async () => await RequestAnalysisResultsAsync());
+        }
+
+        private async Task RequestAnalysisResultsAsync()
```
yongyan-gh, line 237: >RequestAnalysisResultsAsync [](http://example.com/codeflow?start=27&length=27)

nit: would create a wrapper class to deal with ResultSourceService call/event handling, make the vs package code short, what do you think

src/Sarif.Viewer.VisualStudio.ResultSources.GitHubAdvancedSecurity/GitHubInfoBarHelper.cs
```diff
@@ -0,0 +1,16 @@
...
+using Microsoft.VisualStudio.Imaging;
+using Microsoft.VisualStudio.Imaging.Interop;
+
+namespace Microsoft.Sarif.Viewer.ResultSources.Domain
+{
+    internal static class GitHubInfoBarHelper
+    {
+        internal static ImageMoniker GetInfoBarImageMoniker()
+        {
+            return KnownMonikers.StatusWarning;
```
yongyan-gh, line 13: >StatusWarning [](http://example.com/codeflow?start=33&length=13)

why the same function in Sarif.Viewer.VisualStudio.ResultSources.GitHubAdvancedSecurity.2022 returns a different value (KnownMonikers.GitHub)?

src/Sarif.Viewer.VisualStudio.Core/SarifViewerPackage.cs
EasyRhinoMSFT: The other one I'm aware of is in SarifFolderMonitor and I didn't want to take a dependency on that. I'll file a bug to find a better central home for this.

EasyRhinoMSFT: [https://dev.azure.com/mseng/1ES/_workitems/edit/1967398](https://dev.azure.com/mseng/1ES/_workitems/edit/1967398)

src/Sarif.Viewer.VisualStudio.Shell.Core/InfoBarService.cs
```diff
@@ -0,0 +1,95 @@
...
+        private readonly IServiceProvider serviceProvider;
+        private uint cookie;
+
+        public InfoBarService(IServiceProvider serviceProvider)
+        {
+            this.serviceProvider = serviceProvider;
+        }
+
+        /// <inheritdoc cref="IInfoBarService.ShowInfoBar(InfoBarModel)"/>
+        public IVsInfoBarUIElement ShowInfoBar(InfoBarModel infoBarModel)
```
yongyan-gh, line 28: >IVsInfoBarUIElement [](http://example.com/codeflow?start=15&length=19)

why not return Result<...> as other methods do?

src/Sarif.Viewer.VisualStudio.Shell.Core/StatusBarService.cs
```diff
@@ -0,0 +1,103 @@
...
+            await ThreadHelper.JoinableTaskFactory.SwitchToMainThreadAsync();
+
+            this.StatusBar.IsFrozen(out int frozen);
+
+            if (frozen != 0)
+            {
+                this.StatusBar.FreezeOutput(0);
+            }
+
+            this.StatusBar.SetText(text);
```
yongyan-gh, line 61: >; [](http://example.com/codeflow?start=40&length=1)

after setting text, should we freeze status bar`this.StatusBar.FreezeOutput(1);`?

I saw the doc below unset/set the freeze
[https://github.com/MicrosoftDocs/visualstudio-docs/blob/main/docs/extensibility/extending-the-status-bar.md](https://github.com/MicrosoftDocs/visualstudio-docs/blob/main/docs/extensibility/extending-the-status-bar.md)

src/Sarif.Viewer.VisualStudio.Shell.Core/InfoBarService.cs
EasyRhinoMSFT: Yes should be Maybe<> in this case.

EasyRhinoMSFT: Ah I remember why I could,'t do it this way. IVsInfoBarUIElement can't be passed between assemblies as a generic type.
[https://stackoverflow.com/questions/35270785/c-sharp-cannot-be-used-across-assembly-boundaries-because-it-has-a-generic-type](https://stackoverflow.com/questions/35270785/c-sharp-cannot-be-used-across-assembly-boundaries-because-it-has-a-generic-type)

yongyan-gh (approved):
:shipit:

PR #501, opened 2022-07-11
Enable Run Static Analyzer Gesture For Folder View

yongyan-gh:
#Issue Description

When open a solution in VS, user can right click the solution/project/file in the solution explorer, in the context menu, click "Run Static Analyzer" to start the Sarifer analysis.
When open a folder view in VS (main menu -> "File" -> "Open" -> "Folder..."), there is no way to start Sarifer static analyzer. 

This change enables a context menu item "Run Static Analyzer", if user right click on the folders and files in folder view of solution explorer, and it will perform the Sarifer analysis for the selected folder and files.


src/Sarif.Sarifer.Core/BackgroundAnalyzerBase.cs
```diff
@@ -57,7 +52,7 @@ public async Task<Stream> AnalyzeAsync(string path, string text, CancellationTok
         {
             text = text ?? throw new ArgumentNullException(nameof(text));
 
-            string solutionDirectory = await GetSolutionDirectoryAsync().ConfigureAwait(continueOnCapturedContext: false);
+            string solutionDirectory = await VsUtilities.GetSolutionDirectoryAsync().ConfigureAwait(continueOnCapturedContext: false);
```
EasyRhinoMSFT, line 55: >VsUtilities [](http://example.com/codeflow?start=45&length=11)

I think we need to put this into its own leaf-node assembly (maybe the new Shell one) so all projects can access it.

src/Sarif.Sarifer.Core/Commands/AnalyzeSolutionFolderNodeExtender.cs
```diff
@@ -0,0 +1,210 @@
...
+
+                        targetFiles.AddRange(
+                            GetFilesInDirectory(
+                                directory
+                                .GetDirectories("*.*", SearchOption.TopDirectoryOnly)
+                                .Where(d => !d.Attributes.HasFlag(FileAttributes.Hidden))
+                                .Select(d => d.FullName)));
+                    }
+
+                    // ignore the directory if error occured
```
EasyRhinoMSFT, line 200: >occured [](http://example.com/codeflow?start=53&length=7)

occurred

src/Sarif.Sarifer.Core/BackgroundAnalyzerBase.cs
yongyan-gh: Started moving it to Shell project, but found there are some dependcies on Sarif.SDK and test flag etc need to be take care, it introduces many changes so I prefer to do that in another PR. created work item [https://github.com/microsoft/sarif-visualstudio-extension/issues/519](https://github.com/microsoft/sarif-visualstudio-extension/issues/519) for tracking.

EasyRhinoMSFT (approved):
:shipit:

PR #508, opened 2022-07-20
Pull 'users/cmeyer/key events interop' into 'feature/key events'

EasyRhinoMSFT:
Add a new interop endpoint with a corresponding service that loads the SarifLog into the tool window and opens the location file, if any.


src/Sarif.Viewer.VisualStudio.Core/Services/DataService.cs
```diff
@@ -0,0 +1,56 @@
...
+            Assumes.True(sarifLog.Runs?.Count == 1);
+
+            var componentModel = (IComponentModel)AsyncPackage.GetGlobalService(typeof(SComponentModel));
+            if (componentModel != null)
+            {
+                var dte = Package.GetGlobalService(typeof(DTE)) as DTE2;
+                var projectNameCache = new ProjectNameCache(dte?.Solution);
+
+                Result result = sarifLog.Runs.First().Results.First();
+                var sarifErrorListItem = new SarifErrorListItem(sarifLog.Runs.First(), 0, result, string.Empty, projectNameCache);
```
yongyan-gh, line 44: >sarifErrorListItem [](http://example.com/codeflow?start=20&length=18)

wondering if directly create SarifErrorListItem should work or not. It needs CodeAnalysisResultManager to pre-process results and create cache etc. See ErrorListService.WriteRunToErrorList() for example.

Another option maybe is re-use ErrorListService.ProcessSarifLogAsync() if possible.

EasyRhinoMSFT: Yeah I think we'll need to do some refactoring to accommodate this scenario

yongyan-gh (approved):
:shipit:

PR #510, opened 2022-07-27
Update CONTRIBUTING.md

EasyRhinoMSFT:


eddynaka:
Hi @EasyRhinoMSFT ,

The change looks good, but I have a few points:
- should we include the link to sarif-pattern-matcher readme, since besides the VS options, you will also need to install C++/spectre/etc?
- we should also recommend installing VS2019 and, if you need, VS2022. The reason for that is: VS2019 contains .NET Framework 4.5.2 and 4.6.1. Since those versions got deprecated (out of support), you cannot install from the .net website anymore.
- you could follow the other repositories and add a CODEOWNERS file that would automatically add people to review (low priority and not that relevant).

eddynaka:
Actually, updating, you can install from the website, but we should give them the links:
- [https://dotnet.microsoft.com/en-us/download/dotnet-framework/net452](https://dotnet.microsoft.com/en-us/download/dotnet-framework/net452)
- [https://dotnet.microsoft.com/en-us/download/dotnet-framework/net461](https://dotnet.microsoft.com/en-us/download/dotnet-framework/net461)
etc

---
In reply to: [1197515417](https://github.com/microsoft/sarif-visualstudio-extension/pull/510#issuecomment-1197515417) [](http://example.com/codeflow?ancestors=1197515417)

CONTRIBUTING.md
```diff
@@ -1,13 +1,153 @@
...
-provided by the bot. You will only need to do this once across all repos using our CLA.
+This project has adopted the [Microsoft Open Source Code of Conduct](https://opensource.microsoft.com/codeofconduct/). For more information see the [Code of Conduct FAQ](https://opensource.microsoft.com/codeofconduct/faq/) or contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additional questions or comments.
 
-This project has adopted the [Microsoft Open Source Code of Conduct](https://opensource.microsoft.com/codeofconduct/).
-For more information see the [Code of Conduct FAQ](https://opensource.microsoft.com/codeofconduct/faq/) or
-contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additional questions or comments.
+## Overview
+This document guides you through the steps necessary to begin development work on the SARIF Viewer extension for Visual Studio. It also provides useful debugging tips and key breakpoint locations.
+
+## First-time setup
```
eddynaka, line 11: >First-time setup [](http://example.com/codeflow?start=3&length=16)

you should also point to sarif-sdk / sarif-pattern-matcher contributing, since it requires those to compile

CONTRIBUTING.md
```diff
@@ -1,13 +1,153 @@
...
+
+### 1. Install Visual Studio
+You can use VS 2019 or 2022 for SARIF Viewer dev work, but you will need both installed to fully test your changes.
+[Visual Studio 2019](https://visualstudio.microsoft.com/vs/older-downloads/)
+[Visual Studio 2022](https://visualstudio.microsoft.com/vs/enterprise/) 
+
+Within the Visual Studio Installer:
```
yongyan-gh, line 18: also need to install "MSVC v143 VS 2022 C++ x64/x86 Spectre-mitigated libs (Latest)" component, its required to build spam submodule.

eddynaka (approved):

yongyan-gh (approved):

PR #515, opened 2022-08-04
Fix null reference exception when viewer package is not loaded.

yongyan-gh:
# Description

If open any file in VS editor before Viewer package is loaded, the `KeyEventAdornmentManager` checks Viewer's option and gets null reference exception.

To fix this issue:
- If Option instance is null, means the viewer is not loaded yet or failed to load, do not render the key event text adornments.
- In the constructor method of `KeyEventAdornmentManager`, try to load the Viewer extension if not. This will happen when an editor window is created, e.g. open a file in VS.


EasyRhinoMSFT (approved):
:shipit:

PR #518, opened 2022-08-08
Key event interop intergration fix

yongyan-gh:
- Utilize existing `ErrorListService.ProcessSarifLogAsync` method to load the Sarif log. It calls `WriteRunToErrorList` and build caches required by extension.
- Another option is build key event's own logic around `WriteRunToErrorList`

### DIFF
--- a/src/Sarif.Viewer.VisualStudio.Core/Services/DataService.cs
+++ b/src/Sarif.Viewer.VisualStudio.Core/Services/DataService.cs
@@ -57,29 +57,30 @@ namespace Microsoft.Sarif.Viewer.Services
             Assumes.NotNull(sarifLog);
             Assumes.True(sarifLog.Runs?.Count == 1);
 
-            string logPath = $"{Guid.NewGuid()}.sarif";
-
-            await ErrorListService.ProcessSarifLogAsync(sarifLog, logPath, cleanErrors: false, openInEditor: false, monitorSarifFile: false);
-
-            SarifErrorListItem sarifItem = null;
-            IList<SarifErrorListItem> sarifErrorListItems = CodeAnalysisResultManager.Instance.CurrentRunDataCache.SarifErrors;
-
-            if (sarifErrorListItems?.Any() == true)
+            if (this.componentModel != null)
             {
-                sarifErrorListItems.ToList().ForEach(item => item?.PopulateAdditionalPropertiesIfNot());
-                sarifItem = sarifErrorListItems.First();
-            }
+                int runIndex = CodeAnalysisResultManager.Instance.GetNextRunIndex();
+                var dataCache = new RunDataCache(runIndex, "EnhancedResultData", sarifLog);
+                CodeAnalysisResultManager.Instance.RunIndexToRunDataCache.Add(runIndex, dataCache);
 
-            if (sarifItem != null)
-            {
-                sarifItem.PopulateAdditionalPropertiesIfNot();
+                var dte = Package.GetGlobalService(typeof(DTE)) as DTE2;
+                var projectNameCache = new ProjectNameCache(dte?.Solution);
+                var items = new List<SarifErrorListItem>();
+                Run run = sarifLog.Runs.First();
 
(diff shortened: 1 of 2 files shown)


src/Sarif.Viewer.VisualStudio.Core/Services/DataService.cs
```diff
@@ -48,34 +50,29 @@ public void SendEnhancedResultData(SarifLog sarifLog)
...
-                {
-                    var sarifErrorListItem = new SarifErrorListItem(run, 0, r, string.Empty, projectNameCache);
-                    sarifErrorListItem.PopulateAdditionalPropertiesIfNot();
-                    items.Add(sarifErrorListItem);
-                }
+            SarifErrorListItem sarifItem = CodeAnalysisResultManager.Instance.CurrentRunDataCache.SarifErrors?[0];
+
+            if (sarifItem != null)
+            {
+                sarifItem.PopulateAdditionalPropertiesIfNot();
```
EasyRhinoMSFT, line 68: >PopulateAdditionalPropertiesIfNot [](http://example.com/codeflow?start=26&length=33)

Does this get called for the other items in the error list?

src/Sarif.Viewer.VisualStudio.Core/Services/DataService.cs
```diff
@@ -48,34 +50,29 @@ public void SendEnhancedResultData(SarifLog sarifLog)
...
             Assumes.True(sarifLog.Runs?.Count == 1);
 
-            var componentModel = (IComponentModel)AsyncPackage.GetGlobalService(typeof(SComponentModel));
-            if (componentModel != null)
-            {
-                var dte = Package.GetGlobalService(typeof(DTE)) as DTE2;
-                var projectNameCache = new ProjectNameCache(dte?.Solution);
-                var items = new List<SarifErrorListItem>();
-                Run run = sarifLog.Runs.First();
+            string logPath = $"{Guid.NewGuid()}.sarif";
```
EasyRhinoMSFT, line 60: >logPath [](http://example.com/codeflow?start=19&length=7)

If they send us the same file again, will we have duplicate error list items?

yongyan-gh: how do we identify each key event Sarif log?
if we know the identifier, we can close the items from previous call, and re-open all items from new call if they have same identifier.

src/Sarif.Viewer.VisualStudio.Core/Services/DataService.cs
yongyan-gh: Current implementation these properties get populated when user navigates to (double click on) an error list item to postpone the loading time of these properties from loading Sarif content to navigating to specific item. 

If we expect to navigate to the item without manual operation, then yes we need to populate the properties for all item. Or we can add this call to navigate method so don't need to call it explicitly.

src/Sarif.Viewer.VisualStudio.Core/Services/DataService.cs
yongyan-gh: hi @kylereedmsft 

if the same key event log send to viewer twice, should we show 2 error list items? 

if the 2nd key event log send to viewer, with different key event data compared to previous one, should we refresh previous error list item?

with an identifier we can control what error list items we should update (remove then add again). if not, we need to remove all existing key event error list items for each key event log sent, and create new error list items from the log. 

Had a discussion with Chris he will send some questions regarding the scenario.

kylereedmsft: When we send a log via the interop path, we do not want anything added to the error list.
Visual Studio already has the items in the error list because of background code analysis so if your code adds them too, there will be duplicate entries.

EasyRhinoMSFT: Understood, we will make the change.

yongyan-gh: sorry for the confusing.. updated the fix pls review

yongyan-gh:
close this PR since something wrong with it got error when open it with codeflow, the build doesn't show.

yongyan-gh:
new PR #519 created

PR #520, opened 2022-08-10
Key event interop integration fix

yongyan-gh:
Fix the issue tags/adornments don't show up in the editor.

- Because the interop API call show should not generate any entry in the error list, it does not go to normal `ProcessLogFileAsync` function. 
- TextMarker tags creation depends on the cached data in CodeAnalysisResultManager. Need to add the key event Sarif results into it.


yongyan-gh:
looks like the issue is in my branch. created a new branch and new PR #521

PR #522, opened 2022-08-11
Fix adornments displaying issues

yongyan-gh:
- Key event data is not removed after the VS editor is closed. When the same file re-opened the adornments appear without VS sending the key event data.

- Switch between 2 errors in the same file the adornments are not refreshed.
-- The adornments refresh only when text view layout changes. If switch to another error within same viewport, the layoutChanged event will not be fired.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio.Core/Tags/KeyEventAdornmentManager.cs
+++ b/src/Sarif.Viewer.VisualStudio.Core/Tags/KeyEventAdornmentManager.cs
@@ -2,19 +2,16 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
-using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
 using System.Windows;
 using System.Windows.Controls;
-using System.Windows.Input;
 using System.Windows.Media;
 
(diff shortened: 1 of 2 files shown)


src/Sarif.Viewer.VisualStudio.Core/Tags/KeyEventAdornmentManager.cs
```diff
@@ -69,7 +66,9 @@ internal class KeyEventAdornmentManager
             this.layer = view.GetAdornmentLayer(nameof(KeyEventAdornmentManager));
 
             this.view = view;
-            this.view.LayoutChanged += this.OnLayoutChanged;
+            this.view.LayoutChanged += this.View_LayoutChanged;
+            this.view.GotAggregateFocus += this.View_GotAggregateFocus;
```
yongyan-gh, line 70: Tried listening to `this.sarifEventService.NavigatedItemChanged`, this happens immediately after the `NavigatedItem` is set, at this time the text marker tags are not added to the text view yet. The text adornments depend on the text marker tags. 

`GotAggregateFocus` fires after the navigating to the error location, can be used to refresh the adornments after switching to another error.

EasyRhinoMSFT (approved):
:shipit:

PR #523, opened 2022-08-14
Fix issue the adornments don't show the first time

yongyan-gh:
# Description

Caused by an timing issue, the KeyEventAdornmentManager depends on the ITextMarkerTags in the text view to generate the text adornment. But after navigate to a Key Event error, KeyEventAdornmentManager starts to refresh before the ITextMarkerTags got created.

The fix is listen to `ITagAggregator<ITextMarkerTag>.BatchedTagsChanged` event which occurs when tags are added to or removed.


EasyRhinoMSFT (approved):
:shipit:

PR #524, opened 2022-08-16
Update DataService interfaces

yongyan-gh:
# Description

According to discussion, upodate the DataService interop APIs with 2 changes:
- `SendEnhancedResultData` returns a unique int cookie value for each Sarif log sent.
- Add a new `CloseEnhancedResultData` accept the cookie value as a parameter to cleanup tags/adornments and close Sarif explorer window.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio.Core/Services/DataService.cs
+++ b/src/Sarif.Viewer.VisualStudio.Core/Services/DataService.cs
@@ -26,10 +26,18 @@ namespace Microsoft.Sarif.Viewer.Services
     {
         public const string EnhancedResultDataLogName = "EnhancedResultData";
 
-        private readonly IComponentModel componentModel = (IComponentModel)AsyncPackage.GetGlobalService(typeof(SComponentModel));
+        private readonly IComponentModel componentModel;
+
+        private readonly ISarifErrorListEventSelectionService sarifErrorListEventSelectionService;
+
+        public DataService()
+        {
+            this.componentModel = (IComponentModel)AsyncPackage.GetGlobalService(typeof(SComponentModel));
+            this.sarifErrorListEventSelectionService = this.componentModel.GetService<ISarifErrorListEventSelectionService>();
+        }
 
         /// <inheritdoc/>
-        public void SendEnhancedResultData(Stream stream)
+        public int SendEnhancedResultData(Stream stream)
         {
             Assumes.NotNull(stream);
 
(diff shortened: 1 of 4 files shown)


src/Sarif.Viewer.VisualStudio.Core/Services/DataService.cs
```diff
@@ -80,14 +103,27 @@ private async Task SendEnhancedResultDataAsync(SarifLog sarifLog)
...
 
                 items[0].Locations?.FirstOrDefault()?.NavigateTo(usePreviewPane: false, moveFocusToCaretLocation: true);
             }
 
             SarifExplorerWindow.Find()?.Show();
+
+            return runIndex;
+        }
+
+        private async Task CloseEnhancedResultDataAsync(int cookie)
```
yongyan-gh, line 117: actually we don't need the cookie to clean up since we only keep current key event data.
if we will support multiple key event data, the cookie can be used to distinguish them.

yongyan-gh:
cc @kylereedmsft, the interop API changes

src/Sarif.Viewer.VisualStudio.Core/Services/DataService.cs
```diff
@@ -42,28 +50,43 @@ public void SendEnhancedResultData(Stream stream)
...
+
+            int cookie = -1;
+
+            ThreadHelper.JoinableTaskFactory.Run(async () => cookie = await this.SendEnhancedResultDataAsync(sarifLog));
+
+            return cookie;
         }
 
-        private async Task SendEnhancedResultDataAsync(SarifLog sarifLog)
+        public void CloseEnhancedResultData(int cookie)
```
EasyRhinoMSFT, line 71: >CloseEnhancedResultData [](http://example.com/codeflow?start=20&length=23)

Need an interop method for this

src/Sarif.Viewer.VisualStudio.Core/Services/DataService.cs
```diff
@@ -42,28 +50,43 @@ public void SendEnhancedResultData(Stream stream)
...
-                this.SendEnhancedResultData(sarifLog);
+                return this.SendEnhancedResultData(sarifLog);
             }
+
+            return -1;
         }
 
         /// <inheritdoc/>
-        public void SendEnhancedResultData(SarifLog sarifLog)
+        public int SendEnhancedResultData(SarifLog sarifLog)
```
EasyRhinoMSFT, line 60: >int [](http://example.com/codeflow?start=15&length=3)

Doesn't look like we can return this to the interop caller with the current interop code. It's not critical now, since Kyle is calling this directly, but we should keep it in mind for future investigation.

yongyan-gh: is it because the API has to return Task<bool>?
can we have an out parameter to return it? E.g.
```
        public Task<bool> SendEnhancedResultDataAsync(SarifLog sarifLog, out int cookie)
        {
            return this.CallServiceApiAsync(ViewerDataServiceInterfaceName, (service) =>
            {
                cookie = service.SendEnhancedResultData(sarifLog);
                return true;
            });
        }
```

EasyRhinoMSFT: I'm not sure. Maybe it can return Task<int>?

yongyan-gh: updated return Task<int> pls take a look.

src/Sarif.Viewer.VisualStudio.Interop/SarifViewerInterop.cs
```diff
@@ -281,25 +282,47 @@ public IVsPackage LoadSariferExtension()
...
+        {
+            int cookie = -1;
+
+            bool result = await this.CallServiceApiAsync(ViewerDataServiceInterfaceName, (service) =>
+            {
+                cookie = service.SendEnhancedResultData(stream);
+                return true;
+            });
+
+            return result ? cookie : -1;
```
EasyRhinoMSFT, line 313: >result ? cookie : -1; [](http://example.com/codeflow?start=19&length=21)

Nice and simple

EasyRhinoMSFT (approved):
:shipit:

PR #528, opened 2022-08-18
Update Related Locations tool window UI

EasyRhinoMSFT:
This change adapts the Related Locations portion of the Locations view to handle hierarchical locations based on "nestingLevel" in the location's property bag. This integer value is used by the new NestingLevelToMarginConverter to create a Thickness instance with its Left property set to nestingLevel * [parameter]. Currently the parameter value we use is 20px. This margin accomplishes the visual indentation in the Message column.

This change also moves the Related Locations view to a ListView control and simplifies the location text element. The text in the Location column trims with an ellipsis while the text in the Message column wraps.

image

Note: I'll be working on avoiding the horizontal scrollbar in a separate change.


yongyan-gh:
do you know why there is a 3rd column in the list view?

image

EasyRhinoMSFT:
There isn't, it's just that the two columns are sized smaller than the control. I'm looking into making that better (separate change).

---
In reply to: [1220038088](https://github.com/microsoft/sarif-visualstudio-extension/pull/528#issuecomment-1220038088) [](http://example.com/codeflow?ancestors=1220038088,1220038088)

yongyan-gh:
got it!

---
In reply to: [1220049982](https://github.com/microsoft/sarif-visualstudio-extension/pull/528#issuecomment-1220049982) [](http://example.com/codeflow?ancestors=1220049982,1220038088)

src/Sarif.Viewer.VisualStudio.Core/Themes/LocationsStyles.xaml
```diff
@@ -61,6 +69,7 @@
 
     <Style x:Key="LocationTextStyle" BasedOn="{StaticResource LocationTextBaseStyle}" TargetType="TextBlock">
         <Setter Property="Margin" Value="5,0,0,0"/>
+        <Setter Property="TextTrimming" Value="CharacterEllipsis"/>
```
yongyan-gh, line 72: ><Setter Property="TextTrimming" Value="CharacterEllipsis"/> [](http://example.com/codeflow?start=8&length=59)

:thumbsup:

yongyan-gh (approved):
:shipit:

PR #530, opened 2022-08-23
Dismiss key event adornments after Sarif Explorer window closed

yongyan-gh:
# Description

As requested, listen to ToolWindow's OnClose event, clean up selected item, the tags/adornments are removed.

Also updated the "enable key event text adornments in VS editor" option's default value to true as requested.


EasyRhinoMSFT (approved):
:shipit:

PR #533, opened 2022-09-01
Use TreeView for Related Locations

EasyRhinoMSFT:
This change moves Related Locations from a ListView to a TreeView. A Result can specify an int property called NestingLevel that indicates its depth in the tree. NestingLevel values are expected to be in zero-based and sequentially numbered in depth-first order.
For example, this tree would be described with these NestingLevel values, with result objects in this order:
image

If the NestingLevel property is not specified, it has an implied value of zero and is therefore a root node.
Tab traversal is improved, and navigation to locations is performed by double-click, Enter, or spacebar.
To enable key binding, a new control class that inherits TreeView is necessary.

image

This shows a selected item (gray) and a mouse-over item:
image


src/Sarif.Viewer.VisualStudio.Core/Models/SarifErrorListItem.cs
```diff
@@ -645,6 +642,42 @@ internal void PopulateAdditionalPropertiesIfNot()
...
+        internal void BuildRelatedLocationsTree()
+        {
+            var rootNode = new LocationModel();
+            LocationModel lastNode = rootNode;
+            int lastLevel = -1;
+
+            foreach (Location location in this.SarifResult.RelatedLocations)
+            {
+                var locationModel = location.ToLocationModel(this.SarifResult.Run, resultId: this.ResultId, runIndex: this.RunIndex);
+                int levelChange = locationModel.NestingLevel - lastLevel;
```
yongyan-gh, line 654: >NestingLevel [](http://example.com/codeflow?start=48&length=12)

do we assume there always a first location in the array nesting level starts with 0? tested if first location nesting level is not 0, the tree doesn't render.

EasyRhinoMSFT: Like I said in the description, I'm assuming the locations are numbered correctly. I'll ask Hwi-sung if this is safe to do.

EasyRhinoMSFT: Confirmed that Kyle is making the same assumption

yongyan-gh (approved):
:shipit:

PR #534, opened 2022-09-02
Key Event bug fixes

yongyan-gh:
## Issue 1
After key event enhanced data sent to viewer, the tags/adornments show in editor and sarif explorer show details. Now if open a Sarif log file, VS open the file in a new editor window, but the Sarif explorer is not cleared still showing key event data of previous selected error/warning, 

## Issue 2
The key event adornment texts are not aligned when the lines contain tab indents and space indents mixed.
Before:
image

After:
image


EasyRhinoMSFT (approved):
:shipit:

PR #535, opened 2022-09-07
Fix couple Key Event issues

yongyan-gh:
# Description

- Enable tooltip for all text adornments. [Bug 1611751](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1611751): 

- Show index in AnalysisSteps list. [Bug 1611754](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1611754)
image

PR #536, opened 2022-09-07
Revert DataService.CloseEnhancedResultData interface change

EasyRhinoMSFT:
Undo removal of the cookie parameter

### DIFF
--- a/src/Sarif.Viewer.VisualStudio.Core/OpenLogFileCommands.cs
+++ b/src/Sarif.Viewer.VisualStudio.Core/OpenLogFileCommands.cs
@@ -259,7 +259,7 @@ namespace Microsoft.Sarif.Viewer
             try
             {
                 await ErrorListService.ProcessLogFileAsync(logFile, toolFormat, promptOnLogConversions: true, cleanErrors: true, openInEditor: true).ConfigureAwait(continueOnCapturedContext: false);
-                new DataService().CloseEnhancedResultData();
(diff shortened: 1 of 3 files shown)


yongyan-gh (approved):

src/Sarif.Viewer.VisualStudio.Core/OpenLogFileCommands.cs
```diff
@@ -259,7 +259,7 @@ private async System.Threading.Tasks.Task MenuItemCallbackAsync(object sender, E
             try
             {
                 await ErrorListService.ProcessLogFileAsync(logFile, toolFormat, promptOnLogConversions: true, cleanErrors: true, openInEditor: true).ConfigureAwait(continueOnCapturedContext: false);
-                new DataService().CloseEnhancedResultData();
+                new DataService().CloseEnhancedResultData(0);
```
michaelcfanning, line 262: >0 [](http://example.com/codeflow?start=58&length=1)

suggest using a named parameter here, 'cookie: 0'

michaelcfanning (approved):
:shipit:

src/Sarif.Viewer.VisualStudio.Core/OpenLogFileCommands.cs
yongyan-gh: applied this change in #535

PR #537, opened 2022-09-08
Fix an issue TextMarkerTags for related locations/analysis steps not updated at right time

yongyan-gh:
# Description

This issue does not impact the VC++ CA key event integration, only occurs if you open a Sarif file contains key event data using Sarif Viewer.

'ISarifErrorListEventSelectionService' has 2 events:
- `SelectedItemChanged` when user single clicks on an item in error list.
- `NavigatedItemChanged` when user double clicks on an item in error list.

The additional properties of `SarifErrorListItem` only got populated after `NavigatedItemChanged` since navigating may produce popup windows etc needs user interactions.

SarifLocationTextMarkerTagger only register `SelectedItemChanged' and set `tagsDirty` to true. When next `NavigatedItemChanged` happens it doesn't refresh tags since `tagsDirty` is not updated.


EasyRhinoMSFT (approved):
:shipit:

PR #540, opened 2022-09-12
Migration to 1ES Templates

payton2022:
Here I have migrated the sarif-visualstudio-extension pipeline so that it has 1ES compliance. [Here](https://dev.azure.com/mseng/1ES/_build/results?buildId=18270846&view=results) is an example of the pipeline before I made these changes, and [here](https://dev.azure.com/mseng/1ES/_build/results?buildId=18266298&view=results) is an example of the pipeline once I added the 1ES templates.


azure-pipelines.yml
```diff
@@ -4,22 +4,63 @@ pr:
...
+            templateContext:
+              inputs:
+              - input: checkout
+                repository: self
+                submodules: recursive
+              outputs:
+              - output: pipelineArtifact
+                targetPath: $(System.DefaultWorkingDirectory)
+            
+              breakGlass:
```
sirkiza, line 44: why do we have breakglass here? just to my understanding.
please add meaningful justification why it's here.

azure-pipelines.yml
```diff
@@ -4,22 +4,63 @@ pr:
...
+              - script: "tree $(System.DefaultWorkingDirectory)"
+                displayName: "tree"
+
+              - task: PowerShell@2
+                displayName: Build and Test
+                inputs:
+                  targetType: filePath
+                  filePath: $(System.DefaultWorkingDirectory)/scripts/BuildAndTest.ps1
+
+              - task: ComponentGovernanceComponentDetection@0
```
sirkiza, line 64: it seems that we run `ComponentGovernanceComponentDetection` anyway as we have `output`

payton2022: removed

azure-pipelines.yml
payton2022: This was because of the batch script command, it seemed to only work when using break glass. Otherwise it threw errors

azure-pipelines.yml
```diff
@@ -4,22 +4,63 @@ pr:
...
+      name: 1ESPipelineTemplates/1ESPipelineTemplates
+      ref: refs/tags/release
+extends:
+    template: v1/1ES.Unofficial.PipelineTemplate.yml@1esPipelines
+    parameters:
+      pool:
+        name: 1ES-Shared-Hosted-Pool_Windows-Server-2022
+        os: windows
+      containers:
+        default_windows_container:
```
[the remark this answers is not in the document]
payton2022: removed container

guardian/SDL/.gdnbaselines
```diff
@@ -0,0 +1,26 @@
+{
+  "version": "latest",
+  "baselines": {
+    "default": {
+      "name": "default",
+      "createdDate": "2022-09-09 14:14:48Z",
+      "lastUpdatedDate": "2022-09-09 14:14:48Z"
+    }
+  },
+  "results": {
```
EasyRhinoMSFT, line 10: >results [](http://example.com/codeflow?start=3&length=7)

I am not familiar with Guardian, can you provide a quick explanation of this file? Also, this element is called "results" but it's not an array?

eddynaka: Guardian is the team that owns the tasks based on some tools such as BinSkim, CredScan, PREFast, etc.

They work as a wrapped to expose the tools in AzureDevOps pipelines.

It is a KeyValuePair.
The key is a hash of the file, which allows you to deduplicate.

In this case, Payton created the suppression due to a false-positive in the tests that I also shared in the email.

eddynaka: If the test can be updated and still work for you, then, yes, I agree that that should be done :)

EasyRhinoMSFT: Changed and merged into main.

guardian/SDL/.gdnbaselines
```diff
@@ -0,0 +1,26 @@
+{
+  "version": "latest",
```
eddynaka, line 2: >version": "latest [](http://example.com/codeflow?start=3&length=17)

@payton2022 , please, pull the last commits and remove this file. We won't need it anymore since  Chris removed the "secrets" from the code.

eddynaka (approved):

PR #541, opened 2022-09-12
Refactor TreeView templates

EasyRhinoMSFT:
There was some kind of breaking change that affected the top level TreeViews ("Locations" and "Related Locations"). This change moves the TreeViewItem container templates out to standalone styles so they can be shared.


yongyan-gh (approved):
:shipit:

PR #542, opened 2022-09-12
Dismiss SARIF Explorer window when close/open solution.

yongyan-gh:
# Description

After navigating to an SARIF result item in the Error List window, the SARIF Explorer windows shows detailed information of current selected item.

If closing the solution or re-opening another solution when the SARIF Explorer window is opened showing information of selected SARIF result, the Error List is cleaned up, but SARIF Explorer window remains the same content.

The fix is to dimiss SARIF Explorer window whenever user closes current solution or re-opens another solution.


EasyRhinoMSFT (approved):
:shipit:

Issue #453, opened 2022-01-24
VS2019 fresh download of VSIX did not work for me


michaelcfanning:
@EasyRhinoMSFT, how about for you?

yongyan-gh:
@michaelcfanning what error/issue did you see? 
I refresh installed the marketplace SARIF viewer on VS 2019, it works for me.

yongyan-gh:
Also clean installed the VS2019 VSIX built from latest main branch. It also works. Pls let me know what issue did you see thanks!

EasyRhinoMSFT:
I don't see the tool window, and if I go to View > Other windows, it comes up empty. This is the live package.

yongyan-gh:
@EasyRhinoMSFT I cannot repo your issue neither...  Ran the APEX tests all passed too. Wondering if its environment related. Feel free to setup a meeting/live session we can investigate more.

![image](https://user-images.githubusercontent.com/76094515/151058349-307e3a42-79d0-47de-81de-37f7041ec86a.png)

yongyan-gh:
Worked with @EasyRhinoMSFT showed me the issue:
1. Open VS 2019 installed the SAIRF viewer extension.
2. Open a valid SARIF log file with results.
3. SARIF explorer does not popup automatically
4. Open SARIF explorer from menu 'View' -> 'Other Windows' -> 'SARIF Explorer', the explorer window is blank.

As I understand this is current viewer's behavior. Viewer does not open SARIF explorer after opening a SARIF log file. 
Only when   you navigate to the result in Error List pane (double click on the error list item), it will open SARIF explorer and populate information. 

I found an old issue discussing about this behavior at #159. Please take a look.
We could show a message in the blank SARIF explorer "Select an error to see details" to the user.

EasyRhinoMSFT:
I find this very weird. The original purpose of the viewer was to provide an enhanced interaction with SARIF logs with a UI that presents their details. The current behavior makes it almost completely invisible and difficult to discover.

tgriff:
I too am confused by the behavior here. This morning I downloaded and installed it on VS2019. I opened a valid SARIF file (works fine with the VSCode extension) but I don't see anything in the Error List or in the SARIF Explorer window. I even opened the Solution to which the SARIF file applies but I still don't see anything. My expectation is that I should see results for any SARIF file I open _somewhere_. Any logs or anything I can collect here?

michaelcfanning:
[...]

PR #552, opened 2022-10-05
Fix an issue cannot navigate to the source file

yongyan-gh:
# Description

When the user launches Visual Studio using command line at the repo's root directory, to open the folder view in VS. E.g.
`C:\repos\devcanvas\devenv.exe .`

The current working directory is set to the repo's root directory. The File.Exists() check for a relative path under this directory now returns true. E.g. File.Exists("client/Console/InsightConsole/FileFetcher.cs") is false if current working directory is not the repo root, but its true if current working directory is set to repo root.

Fix the issue in the code assumes File.Exists() return true only for the full file path.

Fix for #453

### DIFF
--- a/src/Sarif.Viewer.VisualStudio.Core/ResultTextMarker.cs
+++ b/src/Sarif.Viewer.VisualStudio.Core/ResultTextMarker.cs
@@ -402,19 +402,19 @@ namespace Microsoft.Sarif.Viewer
                 return false;
             }
 
-            if (this.fileSystem.FileExists(this.resolvedFullFilePath) &&
-                Uri.TryCreate(this.resolvedFullFilePath, UriKind.Absolute, out Uri uri))
+            if (this.fileSystem.FileExists(this.resolvedFullFilePath))
             {
-                // Fill out the region's properties
-                this.fullyPopulatedRegion = FileRegionsCache.Instance.PopulateTextRegionProperties(this.region, uri, populateSnippet: true);
-            }
-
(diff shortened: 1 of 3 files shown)


src/Sarif.Viewer.VisualStudio.Core/ResultTextMarker.cs
```diff
@@ -389,13 +402,21 @@ private bool TryToFullyPopulateRegionAndFilePath()
...
 
-            if (File.Exists(this.resolvedFullFilePath) &&
+            if (this.fileSystem.FileExists(this.resolvedFullFilePath) &&
                 Uri.TryCreate(this.resolvedFullFilePath, UriKind.Absolute, out Uri uri))
             {
                 // Fill out the region's properties
                 this.fullyPopulatedRegion = FileRegionsCache.Instance.PopulateTextRegionProperties(this.region, uri, populateSnippet: true);
             }
 
+            if (this.fileSystem.FileExists(this.resolvedFullFilePath) &&
```
EasyRhinoMSFT, line 412: >if [](http://example.com/codeflow?start=12&length=2)

Should this be an "else if" to skip it if the earlier "if" has run?

yongyan-gh: yes combined into 1 "if" statement

EasyRhinoMSFT (approved):
:shipit:

Issue #288, opened 2020-11-26
No line numbers in error list if results have only offset/length


lgolding:
If you open a SARIF log file whose result location regions contain only offset/length properties (that is, they do _not_ contain line/column properties), then the Error List's Line and Col columns show 0.

This is because, for performance reasons, we only call `FileRegionsCache.PopulateTextRegionsProperties` when _navigating_ to an error from the Error List.

Another aspect of this bug is that, even after navigating, when we now have the line/column information in hand, we don't update the Error List item: it still shows Line, Col as 0.

@GabeDeBacker @michaelcfanning @eddynaka

lgolding:
Can we "fault in" the line numbers as Error List items become visible? Does the Error List display virtualization permit this?

yongyan-gh:
When we open a sarif log file in Sarif viewer, it creates error list items with line/column number based on results of Sarif log, without parsing actual file of the results. 
If the result location regions defined by offset/length, its start/end line/column properties should be 0. This is why line/column show 0 for offset/length results.

When we navigate to the error list item (e.g. double click it) with only offset/length region, viewer will try to resolve the file path and read file content, then calculate the start/end line/column using offset/length.

Question is should viewer try to resolve file path & read file content for each result when Sarif log is loaded? 
Or we can populate the line/column after navigate to that item.

EasyRhinoMSFT:
We definitely can't open files when the log is loaded. @yongyan-gh let's try your second suggestion, I think it could be a fairly simple fix.

yongyan-gh:
Tried the solution updating the Sarif error list item after resolving the actual line/column number.
But seems we didn't implement a mechanism to update a single ErrorList table entry. We can use `ITableDataSink` to remove old entry and add newly updated entry, or replace the entry. Looking into this solution.

PR #553, opened 2022-10-10
Refresh Error List item if property values change

yongyan-gh:
# Description

Some Sarif result defines region using `charOffset` and `charLength`, after Sarif loaded into Error List, the line/column number is 0. When the user navigates to the result, the viewer tries to resolve the source file path and populate the region with line/column numbers. But the Error List still shows 0 line/column number.

# Fix:
Update the `SarifErrorListItem`'s `LineNumber`/`ColumnNumber` after the viewer populated the region related to the result.
Implement a way to "refresh" an entry in the error list so if the entry's property changes, we can update the Error List.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio.Core/ErrorList/SarifTableDataSource.cs
+++ b/src/Sarif.Viewer.VisualStudio.Core/ErrorList/SarifTableDataSource.cs
@@ -117,7 +117,7 @@ namespace Microsoft.Sarif.Viewer.ErrorList
             {
                 if (this.logFileToTableEntries.TryGetValue(newItem.LogFilePath, out List<SarifResultTableEntry> entries))
                 {
-                    SarifResultTableEntry entryToRemove = entries.First(entry => (int)entry.Identity == oldItemIdentity);
+                    SarifResultTableEntry entryToRemove = entries.FirstOrDefault(entry => (int)entry.Identity == oldItemIdentity);
 
                     if (entryToRemove != null)
                     {
--- a/src/Sarif.Viewer.VisualStudio.Core/ErrorList/SarifTableDataSource.cs
+++ b/src/Sarif.Viewer.VisualStudio.Core/ErrorList/SarifTableDataSource.cs
@@ -107,6 +107,35 @@ namespace Microsoft.Sarif.Viewer.ErrorList
             }
         }
 
+        public void UpdateError(int oldItemIdentity, SarifErrorListItem newItem)
+        {
+            if (newItem == null)
+            {
+                return;
+            }
+
+            using (this.tableEntriesLock.EnterWriteLock())
+            {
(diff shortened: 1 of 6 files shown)


yongyan-gh:
#288

src/Sarif.Viewer.VisualStudio.Core/Models/SarifErrorListItem.cs
```diff
@@ -884,5 +894,33 @@ internal void MessageInlineLink_Click(object sender, RoutedEventArgs e)
                 SdkUIUtilities.OpenExternalUrl(uriString);
             }
         }
+
+        // Generates an unique hash value using the properties affect content of error list item.
+        // If any of these properties changes, needs to refresh error list item.
+        internal int GetIdentity()
```
yongyan-gh, line 900: >GetIdentity [](http://example.com/codeflow?start=21&length=11)

was thinking about adding PropertyChanged event for each property matters.
But we usually update multiple properties at once and don't want to update the ErrorList multiple times.
Let me know if any thought

marmegh: Please! I had to uninstall sarifer because of all of the error list "updates" (no actually changes, just every time sarifer completed a scan...error list popped up).

yongyan-gh: The Sarifer scans the content of current editor window after it opens, and re-scan after any keyboard input. It removes all existing results and adds the "new" results back, even the results are same. It should be a good place to call "update" instead of removing all and adding back.

You can disable the scan in Sarifer option to avoid uninstallation for now. :grinning:

src/Sarif.Viewer.VisualStudio.Core/ErrorList/SarifTableDataSource.cs
```diff
@@ -107,6 +106,35 @@ public void AddErrors(IEnumerable<SarifErrorListItem> errors)
...
+            if (newItem == null)
+            {
+                return;
+            }
+
+            using (this.tableEntriesLock.EnterWriteLock())
+            {
+                if (this.logFileToTableEntries.TryGetValue(newItem.LogFilePath, out List<SarifResultTableEntry> entries))
+                {
+                    SarifResultTableEntry entryToRemove = entries.First(entry => (int)entry.Identity == oldItemIdentity);
```
EasyRhinoMSFT, line 120: >First [](http://example.com/codeflow?start=66&length=5)

I think you'll need FirstOrDefault here.

yongyan-gh: Good catch, i wanted to use FirstOrDefault here. Also added a test case for not existing identity.

EasyRhinoMSFT (approved):
:shipit:

PR #557, opened 2022-10-31
GHAS alert dismissal

EasyRhinoMSFT:
Dynamically creating flyout menus and commands in the Error List context menu (and potentially elsewhere) is a necessary capability for the result source services. Unfortunately the VS extensibility SDK doesn't provide a nice API for this purpose.
- For menus, you specify the DynamicVisibility, DefaultInvisible, and TextChanges command flags. The latter indicates that you want the BeforeQueryStatus event to fire. Your handler for this event decides whether the menu should be displayed and what the text string is.
- For commands (menu items, called "Buttons" in the .vsct table), it's even clunkier. You define one Button with the same command flags mentioned earlier, plus the DynamicItemStart flag. This indicates that the button can be followed by additional dynamically-created buttons, which are identified by sequential IdSymbol values.
- Importantly, Menus cannot be dynamically created in code. Buttons can only be created via the above process. This means that everything has to be defined ahead of time in the .vsct file (think of them as placeholders). I have waffled on this but ultimately decided not to predefine a huge number of placeholders. If a future result source needs to create a complex menu structure, we can add more placeholders at that time. The GHAS service only needs one flyout for a handful (three currently) of buttons so I've only included one flyout menu, one child container group, and one DynamicItemStart button.

Because the service projects can't reference the Viewer project, the only way form them to request changes to the Error List context menu is via events. This is the best we can do right now but will improve once we break the Viewer project apart (https://dev.azure.com/mseng/1ES/_workitems/edit/2003961).

Another concession that will improve later: because the services can't reference the Viewer project, they can't know what a SarifErrorListItem is. To mitigate this, the SarifViewerPackage extracts the SarifLog object(s) and provides them to the service's Invoke handler(s). This issue will be resolved when the model classes are moved into a leaf project that can be referenced by both the Viewer and result source service projects.

image


src/Sarif.Viewer.VisualStudio.Core/SarifCommandPackage.vsct
```diff
@@ -183,5 +212,17 @@
...
+    <GuidSymbol name="guidSarifErrorListResultSourceServiceCommandSet" value="{b04424d9-49bc-4e04-9ecc-ad5b68cce4bc}">
+      <!-- MENUS -->
+      <IDSymbol name="9655cb55-c0b8-4636-9d9a-6a8685f76ee0" value="0x5000"/>
+
+      <!-- GROUPS -->
+      <IDSymbol name="2ee78ada-af4b-4e2c-aaa8-0326ef12d611" value="0x6000"/>
+      <IDSymbol name="9abdbc25-a08d-49dd-9065-4eb669194441" value="0x6001"/>
+
+      <!-- BUTTONS -->
+      <IDSymbol name="c4883543-de09-4989-83c2-16e40d105831" value="0x8B67"/>
```
yongyan-gh, line 224: >c4883543 [](http://example.com/codeflow?start=22&length=8)

should we use the meaningful names as other IDSymbols?

EasyRhinoMSFT: I did it this way because there could potentially be a lot of these. I actually wrote a tool to generate all of the XML

EasyRhinoMSFT: And this is why I created a separate command set.

src/Sarif.Viewer.VisualStudio.Core/SarifViewerPackage.cs
```diff
@@ -265,10 +297,136 @@ private void SolutionEvents_OnAfterBackgroundSolutionLoadComplete(object sender,
...
+                                }
+                            }
+                        }
+                    }
+
+                    break;
+            }
+        }
+
+        private void InvokeResultSourceServiceMenuCommand(object sender)
```
yongyan-gh, line 373: >InvokeResultSourceServiceMenuCommand [](http://example.com/codeflow?start=21&length=36)

nit: menu command related code can be in a separate class e.g. ErrorListCommand.cs

src/Sarif.Viewer.VisualStudio.Core/ErrorList/ErrorListService.cs
```diff
@@ -62,7 +62,10 @@ static ErrorListService()
 
         public static void ProcessLogFile(string filePath, string toolFormat, bool promptOnLogConversions, bool cleanErrors, bool openInEditor)
         {
-            ThreadHelper.JoinableTaskFactory.Run(() => ProcessLogFileWrapperAsync(filePath, toolFormat, promptOnLogConversions, cleanErrors, openInEditor));
+            if (!IsSarifLogOpened(filePath))
```
yongyan-gh, line 65: >IsSarifLogOpened [](http://example.com/codeflow?start=17&length=16)

should be the caller's response to check if the SarifLog is already opened? what if the caller want to re-open the save Sarif Log file after its updated?

src/Sarif.Viewer.VisualStudio.ResultSources.Factory.Core/ResultSourceFactory.cs
```diff
@@ -29,9 +29,9 @@ public class ResultSourceFactory : IResultSourceFactory
     {
         private readonly StandardKernel standardKernel;
         private readonly string solutionRootPath;
-        private readonly List<Type> resultSourceTypes = new List<Type>
+        private readonly Dictionary<Type, (int, int)> resultSources = new Dictionary<Type, (int firstMenuId, int firstCommandId)>
```
yongyan-gh, line 32: >(int, int) [](http://example.com/codeflow?start=42&length=10)

maybe create a new class/struct for the combination of menu id + command id? its also used in menu event args, and will be more readable

src/Sarif.Viewer.VisualStudio.Core/ErrorList/ErrorListService.cs
EasyRhinoMSFT: This is a bug fix, but you may be right that we don't necessarily want to do this.
The bug happens when I write the log file to the .sarif folder. The file watcher's Created event fires four times which results in four copies of each result in the error list.
The refactor I did in this change exposed this problem by changing the event sequence.
I'll see about a less brutal solution.

src/Sarif.Viewer.VisualStudio.Core/SarifViewerPackage.cs
EasyRhinoMSFT: Good point.

src/Sarif.Viewer.VisualStudio.Core/ErrorList/ErrorListService.cs
EasyRhinoMSFT: Weird. Once I moved the two event handlers out of SarifViewerPackage, this problem stopped happening. I wonder if it has something to do with IVsTaskList2. Or maybe it was some kind of bad state. Blech.

src/Sarif.Viewer.VisualStudio.ResultSources.Factory.Core/ResultSourceFactory.cs
EasyRhinoMSFT: I'd prefer to keep it this way, even though I agree it's a bit more work to parse visually. On the other hand, you get to see the unpacked property names, which helps with comprehension.
This approach also avoids having to deal with IEqualityComparer.

src/Sarif.Viewer.VisualStudio.Core/ErrorList/ErrorListService.cs
yongyan-gh: good to see its resolved!

src/Sarif.Viewer.VisualStudio.ResultSources.Factory.Core/ResultSourceFactory.cs
yongyan-gh: make sense

yongyan-gh (approved):
:shipit:

PR #558, opened 2022-11-07
Add support for key event adornments to GHAS service

EasyRhinoMSFT:
This change enables result source extensions to turn on adornments when their log is loaded. A couple of tweaks we necessary to defend against conditions where location regions are null.


yongyan-gh (approved):
:shipit:

PR #559, opened 2022-11-07
Support empty line text region

yongyan-gh:
# Description

In certain scenarios the region starts with an empty line (start column is 0, line length is also 0). The viewer cannot create correct `TextSpan` in these cases, so it doesn't show correct squiggles or text markers as expected.

E.g. in below screenshot, the region starts at line 52, an empty line.
image

Also, in the suggested fixes scenario, if the `deletedRegion` length is 0, e.g. 
```
"deletedRegion": {
    "charOffset": 20777,
    "charLength": 0
},
```
It means an insertion point, also not supported now.

The fix is to add supports in these cases.

Also test cases added for the coverage.


src/Sarif.Viewer.VisualStudio.UnitTests/SpanHelperTests.cs
```diff
@@ -0,0 +1,222 @@
...
+using Microsoft.VisualStudio.Text;
+using Microsoft.VisualStudio.TextManager.Interop;
+
+using Moq;
+
+using Xunit;
+
+namespace Microsoft.Sarif.Viewer.VisualStudio.UnitTests
+{
+    public class SpanHelperTests : SarifViewerPackageUnitTests
```
EasyRhinoMSFT, line 18: >SpanHelperTests [](http://example.com/codeflow?start=17&length=15)

Nice thorough test coverage

EasyRhinoMSFT (approved):
:shipit:

PR #562, opened 2022-11-15
Refactor result source plumbing so services can check Options dialog state

EasyRhinoMSFT:
This is clunky but will be fixed once the project restructuring is complete. All this does is pass a delegate from the extension package class through to the result source services.
The entire point is to get rid of the awareness of the GHAS service's option. With the restructuring, the services will be able to check this directly.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio.ResultSources.GitHubAdvancedSecurity.Core/Services/GitHubSourceService.cs
+++ b/src/Sarif.Viewer.VisualStudio.ResultSources.GitHubAdvancedSecurity.Core/Services/GitHubSourceService.cs
@@ -58,6 +58,7 @@ namespace Microsoft.Sarif.Viewer.ResultSources.GitHubAdvancedSecurity.Services
         private const string GetAnalysesEndpoint = "analyses";
         private const string DismissAlertEndpointFormat = "alerts/{0}";
         private const string GitLocalRefFileBaseRelativePath = @".git\refs\remotes\origin";
+        private const string OptionStateKey = "GitHubAdvancedSecurity";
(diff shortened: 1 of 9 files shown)


src/Sarif.Viewer.VisualStudio.ResultSources.GitHubAdvancedSecurity.Core/Services/GitHubSourceService.cs
```diff
@@ -58,6 +58,7 @@ public partial class GitHubSourceService : IResultSourceService, IGitHubSourceSe
         private const string GetAnalysesEndpoint = "analyses";
         private const string DismissAlertEndpointFormat = "alerts/{0}";
         private const string GitLocalRefFileBaseRelativePath = @".git\refs\remotes\origin";
+        private const string OptionStateKey = "GitHubAdvancedSecurity";
```
yongyan-gh, line 61: >GitHubAdvancedSecurity [](http://example.com/codeflow?start=47&length=22)

nit: define the const in a common place, SarifViewerOptions and this class can access.

yongyan-gh (approved):
:shipit:

src/Sarif.Viewer.VisualStudio.ResultSources.GitHubAdvancedSecurity.Core/Services/GitHubSourceService.cs
EasyRhinoMSFT: Agreed, but I'm going to address this kind of thing in the restructuring. Too much of a pain currently.

PR #580, opened 2023-01-03
Fix GetSourceFileUri for not slash terminated repositories

JarLob:
When the repository is not slash terminated like `https://github.com/user/repo` the `Uri.TryCreate(this.details.RepositoryUri, sourceRelativePath, out Uri sourceUri)` drops the `repo` from resulting Uri: `Uri.TryCreate("https://github.com/user/repo", "abc/def", out Uri sourceUri)` makes `sourceUri == "https://github.com/user/abc/def"`.

I made the fix in two functions that operate with the `RepositoryUri`, but maybe it can be fixed somewhere even earlier, because it is such a gotcha.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio.2022/Sarif.Viewer.VisualStudio.2022.csproj
+++ b/src/Sarif.Viewer.VisualStudio.2022/Sarif.Viewer.VisualStudio.2022.csproj
@@ -7,6 +7,9 @@
   <PropertyGroup>
     <DocumentationFile>..\..\bld\bin\AnyCPU_Debug\Sarif.Viewer.VisualStudio.2022\Microsoft.Sarif.Viewer.2022.xml</DocumentationFile>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Debug|AnyCPU'">
+    <DeployExtension>True</DeployExtension>
+  </PropertyGroup>
   <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
   <!-- Hack: https://github.com/dotnet/Nerdbank.GitVersioning/issues/404, since its a vsix with pages -->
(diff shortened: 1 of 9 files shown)


EasyRhinoMSFT:
@JarLob the checks appear stuck, please push an innocuous change

EasyRhinoMSFT:
>     public class AdoVersionControlParser : IVersionControlParser

These classes are really similar. Maybe we could take the opportunity to DRY them our with a base class.

---
Refers to: src/Sarif.Viewer.VisualStudio.Core/VersionControlParser/AdoVersionControlParser.cs:16 in e77293a. [](commit_id = e77293a7ad7ea7601681dd605ff95e6982ee414e, deletion_comment = False)

src/Sarif.Viewer.VisualStudio.Core/Sarif.Viewer.VisualStudio.Core.projitems
```diff
@@ -180,6 +180,7 @@
     <Compile Include="$(MSBuildThisFileDirectory)VersionControlParser\AdoVersionControlParser.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)VersionControlParser\GithubVersionControlParser.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)VersionControlParser\IVersionControlParser.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)versioncontrolparser\VersionControlParser.cs" />
```
EasyRhinoMSFT, line 183: >versioncontrolparser [](http://example.com/codeflow?start=49&length=20)

Nit: fix casing

EasyRhinoMSFT (approved):
:shipit:

EasyRhinoMSFT:
Hey @JarLob I had to make a change to the build pipeline for forks. I don't know of a way to trigger the build other than with a commit -- could you please push a no-op change such as adding a period to the end of a comment?

EasyRhinoMSFT:
No worries, there is still something wrong with the build pipeline settings which is why the fifth check is just sitting there. I'm working on it, don't need you to do anything. 👍

EasyRhinoMSFT:
@JarLob it looks like what we're doing is adding people as contributors so they don't have to use a fork. Please accept the invite, move this change to a working branch, and create a new PR so we can get this merged.

PR #583, opened 2023-02-14
Fix issue the key event selected row jumps

yongyan-gh:
# Description

Fix for the key event viewer bug mentioned @ https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1631052

If there are multiple key event steps which have overlapped regions, when you navigate to one of these key event steps by double clicking it, the selected key event step will jump to the first of these key event steps.


EasyRhinoMSFT (approved):
:shipit:

PR #585, opened 2023-03-01
Fix relative path resolution when SLN is in repo root

EasyRhinoMSFT:
DevOps bug 2008452
The fix is to check that the FullFilePath is rooted (absolute) before using it as the resolved path. This check is necessary because File.Exists will return true for relative paths under some circumstances.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio.Core/ResultTextMarker.cs
+++ b/src/Sarif.Viewer.VisualStudio.Core/ResultTextMarker.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.IO;
 
 using Microsoft.CodeAnalysis.Sarif;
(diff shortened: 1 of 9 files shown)


src/Sarif.Viewer.VisualStudio.Core/ResultTextMarker.cs
```diff
@@ -392,7 +401,7 @@ internal bool TryToFullyPopulateRegionAndFilePath()
                 return false;
             }
 
-            if (this.fileSystem.FileExists(this.FullFilePath))
+            if (this.fileSystem2.IsPathRooted(this.FullFilePath) && this.fileSystem.FileExists(this.FullFilePath))
```
yongyan-gh, line 404: 👍

yongyan-gh (approved):

PR #592, opened 2023-03-31
Fixing the bug of error list not populating on re-open

edkazcarlson-ms:
There is a bug where opening a folder after the first open does not cause the error list to populate.
This will re-start the folder monitor when opening a new project is opening, letting the error list and other parts populate and be usable.


EasyRhinoMSFT (approved):
:shipit:

PR #595, opened 2023-04-12
Adding user-settable highlight colors

edkazcarlson-ms:
In this PR I add the option for the user to tune their highlight colors in order to allow for more customization as well as making it so that a user can make sure highlights from different extensions do not collide.
The highlight colors can be tuned by going to the settings through Tools > Options > Sarif Viewer
image
Here you can use the dropdowns in order to select the kind of highlight you want each result to have, letting auser customize their experience. 
The following is an example of an adjusted color scheme from the normal one.
withnewhighlights
This feature supports hot reloads (if you change the setting, it will appear different without needing to close and reopen VS)
It also persists through sessions of VS, meaning that it will remember the settings even if you set them last session


EasyRhinoMSFT:
This is a nice feature. I'd like to propose a couple of changes to the Options dialog:
1. I think a new sub-page called Colors would be a sensible way to keep the General page from getting to busy, and to follow the VS pattern.
2. Also following VS, can we do a color swatch dropdown like what they do in Options > Environment > Fonts and Colors

edkazcarlson-ms:
> 

Sounds good, does something like this look like what you would want for what you see under Tools > Options? 
image

EasyRhinoMSFT:
Oh doesn't need to be that fancy. Just a color swatch with the color name:

image

edkazcarlson-ms:
Sounds good, updated the code and updated the initial description with new image of color picker

src/Sarif.Viewer.VisualStudio.Core/Options/ColorOption.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// </summary>
+        public string ColorName { get; set; }
+
+        /// <summary>
+        /// Gets or sets the text that is to be used when created the color swatch to display to the user.
+        /// </summary>
+        public string Text { get; set; }
+
+        /// <summary>
+        /// Gets or sets the text that is used by the VS API to determine how to highlight a particular span. Must be from the <see cref="PredefinedErrorTypeName"/> class or will always display a red underline.
```
EasyRhinoMSFT, line 19: >will always display a red underline. [](http://example.com/codeflow?start=174&length=36)

AHA! Nice, I couldn't figure that out. :thumbsup:

EasyRhinoMSFT (approved):
:shipit:

PR #600, opened 2023-07-07
Renaming the "I fixed this!!!!" menu option.

edkazcarlson-ms:
Currently the "I fixed this!!!!" menu option is confusing to users as it is unclear about the purpose and seems to do the same thing as the suppression menu item and so this PR hides it in the vsct file.
The "I fixed this!!!!" menu item allows for users to remove an item in the error list temporarily until the end of the VS session, where it will show up again the next time the user opens VS. This user can get similar benefits from the suppression menu item and so we decided to remove for clarity.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio.Core/SarifCommandPackage.vsct
+++ b/src/Sarif.Viewer.VisualStudio.Core/SarifCommandPackage.vsct
@@ -159,14 +159,15 @@
             <ButtonText>Suppress result in SARIF file</ButtonText>
           </Strings>
       </Button>
-      <Button guid="guidSarifErrorListCmdSet" id="cmdidIFixedThisCommand" priority="0x0100" type="Button">
+      <!--Can be used to hide an error list item until the end of the VS session.--> 
+      <!--<Button guid="guidSarifErrorListCmdSet" id="cmdidIFixedThisCommand" priority="0x0100" type="Button">
         <Parent guid="guidSarifErrorListCmdSet" id="ErrorListContextMenuGroup" />
         <CommandFlag>DefaultInvisible</CommandFlag>
         <CommandFlag>DynamicVisibility</CommandFlag>
         <Strings>
           <ButtonText>I fixed this!!!!!</ButtonText>
         </Strings>
-      </Button>
+      </Button>-->
       
       <!-- Error List result source service #1 menu items -->
       <Button guid="guidSarifErrorListResultSourceServiceCommandSet" id="c4883543-de09-4989-83c2-16e40d105831" priority="0x1000" >
@@ -209,7 +210,7 @@
       <IDSymbol name="cmdidNonShippingCodeResultCommand" value="0x0306"/>
       <IDSymbol name="cmdidOtherResultCommand" value="0x0307"/>
       <IDSymbol name="cmdidSuppressResultCommand" value="0x0308"/>
-      <IDSymbol name="cmdidIFixedThisCommand" value="0x0309"/>
+      <!--<IDSymbol name="cmdidIFixedThisCommand" value="0x0309"/>-->
     </GuidSymbol>
 
     <GuidSymbol name="guidSarifErrorListResultSourceServiceCommandSet" value="{b04424d9-49bc-4e04-9ecc-ad5b68cce4bc}">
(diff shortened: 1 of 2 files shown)


src/Sarif.Viewer.VisualStudio.Core/SarifCommandPackage.vsct
```diff
@@ -159,14 +159,15 @@
             <ButtonText>Suppress result in SARIF file</ButtonText>
           </Strings>
       </Button>
-      <Button guid="guidSarifErrorListCmdSet" id="cmdidIFixedThisCommand" priority="0x0100" type="Button">
+      <!--Can be used to hide an error list item until the end of the VS session.-->
```
EasyRhinoMSFT, line 162: ><!-- [](http://example.com/codeflow?start=6&length=4)

On second thought, I think it might be better to just keep the item hidden. To do that, comment out line 238 in .\src\Sarif.Viewer.VisualStudio.Core\ErrorList\ErrorListCommand.cs

michaelcfanning, line 162: >hide [](http://example.com/codeflow?start=25&length=4)

Did you consider retaining the feature and simply providing a clearer designation. Isn't a feature to temporarily remove a finding from the list useful in some contexts? This is a much clearer description, there could be something better.

'Clear finding from error list as presumed fixed.'

This 'I fixed this!!!' rendering makes a good case for a code standard I have labeled 'don't get cute.'. :) Eventually, cute things catch up with you. A subtle danger is that because we object to the cuteness on its surface, we under-value what's actually behind it. Is that happening here?

Just some thoughts. Please proceed as you see fit. :)

edkazcarlson-ms: Chris and I had discussed the naming a bit, we had various ideas on what to replace it with that were concise and clearly explained the feature but it was hard to think of any. 
We had ideas for "Remove from Error List" which is similar to the "Clear finding from error list as presumed fixed." but this doesn't accurately convey to the user that this is a temporary change until the next time the user opens Visual Studio.

I'm wondering what the use case the temporary suppression of results fulfills that the other suppression mechanism does not do properly.

EasyRhinoMSFT: The key difference is that Suppress actually modifies the sarif file, and actually isn't the same thing (suppress == "this isn't an error")

edkazcarlson-ms: I did notice that the suppression got added to the individual results but I'm wondering how we can communicate that to the end user effectively, and how it's different from the current "I fixed this" button.
Also wondering what the use case is to the end user for suppressions vs temporary hiding? To me at least they seem to do the same thing from a UX perspective except that I would need to re-click the temporary hiding every time I opened VS. 

My team has mentioned that it would be nice if there was an option to permanently suppress a warning as a false positive in code. For example, if there was a sarif log for the "VSTHRD108  Assert thread affinity unconditionally" errors we get in this repo, a permanent suppression would act like a fix and add the 
#pragma warning disable VSTHRD108 // Assert thread affinity unconditionally
and 
#pragma warning restore VSTHRD108
around the offending line.
This way, the suppression would be permanent, and subsequent runs of the static analysis tool would not cause it to appear over and over in different sarif logs.

I think that is out of scope for this PR, but I'm still curious about the motivations behind the 2 buttons to better understand what to do for this PR

ShiningMassXAcc: I generally agree with all the comments here.

I'm with Michael on there being a workflow locally that helps someone sort of navigate a checklist to get through items.  They may not be re-running the analysis locally, but it helps them manage their individual work.  To me, some potential verbiage for this might be 'dismiss' - but there is definitely still room for confusion.  I don't know if we have great customer insight into this.

That said, it is most important item (to me) is the accountability for these items, with a clear workflow - and that is where the current text would add confusion.

In modern engineering systems, these defects will be tracked by result systems - and these result systems will have a method that is exposed to the customer. SARIF clearly exposes this contract with the existing field for WorkItemUri.  The menu item could say "Resolve this defect" -> and this would result in executing the provided uri for the tracking work item.  In our case, this would result in an ADO bug being opened in the browser / GitHub issue.  They can view the item there and resolve if they like - but they may also wait to resolve as part of the PR flow (and there isn't currently any good ES UIUX that lets you declare what you are planning to work on at topic branch checkout time).

edkazcarlson-ms:
image
Updated text, keeping the functionality there

EasyRhinoMSFT (approved):
:shipit:

PR #601, opened 2023-07-27
Add wrapping and margins to text blocks

edkazcarlson-ms:
Currently plaintext popups have 2 problems
1. They do not properly wrap text, sending it offscreen where it is unreadable
2. They do not have margins between elements, meaning that when scanning it can be difficult to determine where one result starts and another ends.

Some before and after comparisons:
Before:
image

After:
image

### DIFF
--- a/src/Sarif.Viewer.VisualStudio.Core/Models/ScrollViewerWrapper.cs
+++ b/src/Sarif.Viewer.VisualStudio.Core/Models/ScrollViewerWrapper.cs
@@ -87,7 +87,7 @@ namespace Sarif.Viewer.VisualStudio.Core.Models
                 var stackPanel = new StackPanel();
                 Dictionary<string, int> errorTypeToSeverity = GetRankDictionary();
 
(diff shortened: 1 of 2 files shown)


src/Sarif.Viewer.VisualStudio.Core/Models/ScrollViewerWrapper.cs
```diff
@@ -96,11 +96,20 @@ public object ToolTipContent
...
+                }).ToList();
 
-                foreach (IErrorTag objectToWrap in sortedObjects)
+                for (int i = 0; i < sortedObjects.Count; i++)
                 {
+                    IErrorTag objectToWrap = sortedObjects[i];
                     UIElement tooltip = (UIElement)objectToWrap.ToolTipContent;
+                    if (tooltip is TextBlock textBlock)
+                    {
+                        if (i != 0)
```
EasyRhinoMSFT, line 107: Combine?

EasyRhinoMSFT (approved):
:shipit:

PR #602, opened 2023-08-03
Adding menu for swapping devcanvas endpoint

edkazcarlson-ms:
In order to help debugging processes, we have added a menu to allow the user to hot swap between the different endpoints (prod, ppe, dev). 
It will not show for non-MS users in order to make sure it doesn't confuse them. 
Also made 2 other structural changes to OptionStates, first being that its now constructed when called, not at the start as dictionaries copy the underlying values instead of referencing, meaning that the settings won't take into effect until the user closes and reopens the client. 
Second change is making it string -> obj instead of string -> bool so that we can have more types of settings other than just binary settings.


src/Sarif.Viewer.VisualStudio.ResultSources.GitHubAdvancedSecurity.Core/Services/GitHubSourceService.cs
```diff
@@ -188,7 +188,7 @@ public async Task InitializeAsync()
         /// <inheritdoc cref="IGitHubSourceService.IsActiveAsync()"/>
         public async Task<Result> IsActiveAsync()
         {
-            if (this.GetOptionStateCallback != null && this.GetOptionStateCallback("GitHubAdvancedSecurity"))
+            if (this.GetOptionStateCallback != null && (bool)this.GetOptionStateCallback("GitHubAdvancedSecurity"))
```
EasyRhinoMSFT, line 191: >(bool) [](http://example.com/codeflow?start=55&length=6)

I'd prefer TryParse just in case something is wonky with the menu item

src/Sarif.Viewer.VisualStudio.ResultSources.DeveloperCanvas.UnitTests/DevCanvasAccessorTests.cs
```diff
@@ -46,10 +46,10 @@ public async Task AuthFailureScenarioAsync()
             await handler.AddSendAsyncQueryAsync(DevCanvasWebAPIAccessor.ppeServer, "GET", "", System.Net.HttpStatusCode.Unauthorized);
 
             Mock<IAuthManager> authMockWrongCredentials = new Mock<IAuthManager>();
-            authMockWrongCredentials.Setup(x => x.GetHttpClientAsync())
+            authMockWrongCredentials.Setup(x => x.GetHttpClientAsync(0))
                 .Returns(Task.FromResult<HttpClient>(handler.GetClient()));
 
-            accessor = new DevCanvasWebAPIAccessor(authMockWrongCredentials.Object);
+            accessor = new DevCanvasWebAPIAccessor(() => 0, authMockWrongCredentials.Object); ;
```
EasyRhinoMSFT, line 52: > ; [](http://example.com/codeflow?start=93&length=2)

Remove

src/Sarif.Viewer.VisualStudio.Core/Options/SarifViewerOptionsControlResources.xaml
```diff
@@ -2,7 +2,8 @@
                     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
                     xmlns:system="clr-namespace:System;assembly=mscorlib">
 
-    <system:String x:Key="SarifViewerOptionsControl_GroupBoxHeader">General</system:String>
+    <system:String x:Key="SarifViewerOptionsControl_GeneralGroupBoxHeader">General</system:String>
+    <system:String x:Key="SarifViewerOptionsControl_DevCanvasGroupBoxHeader">DevCanvas</system:String>
```
EasyRhinoMSFT, line 6: Why couldn't this be done in the extension's reserved space? I wanted to avoid this kind of leakage.

edkazcarlson-ms: What do you mean reserved space, should I setup a separate .xaml resources for the strings that we use for the devcanvas settings?

src/Sarif.Viewer.VisualStudio.ResultSources.GitHubAdvancedSecurity.Core/Services/GitHubSourceService.cs
EasyRhinoMSFT: You can call Object.ToString() which will be "True" or "False" if it's a boolean

src/Sarif.Viewer.VisualStudio.Core/Options/SarifViewerOptionsControlResources.xaml
EasyRhinoMSFT: The result source platform supports this, see GitHubSourceService.InitializeAsync as an example.

edkazcarlson-ms: I thought that GitHubSourceService's InitializeAsync method and the corresponding processing in ResultSourceHost_ServiceEvent creates new options for when you right click on error list items, not the options menu which this affects.
I also see that [SarifViewerOptionsControlResources.xaml](https://github.com/microsoft/sarif-visualstudio-extension/pull/602/files/9e1f67abf6a62bfc1c95d2113be5b8d4ab2a49d6#diff-d2cfa9f8d8b3ad366f111020ae0fc4098f35c7c97dd579c37f930d2b61934e6b) includes lines for the GhAS result source service on line 12 with "<system:String x:Key="SarifViewerOptionsControl_EnableGitHubAdvancedSecurity">Enable GitHub Advanced Security scan results</system:String>"

EasyRhinoMSFT: Ahh sorry, I misunderstood where the change was. This is fine; no reasonable way to do it cleaner. The only leakage I couldn't avoid!

src/Sarif.Viewer.VisualStudio.ResultSources.GitHubAdvancedSecurity.Core/Services/GitHubSourceService.cs
```diff
@@ -188,16 +188,20 @@ public async Task InitializeAsync()
         /// <inheritdoc cref="IGitHubSourceService.IsActiveAsync()"/>
         public async Task<Result> IsActiveAsync()
         {
-            if (this.GetOptionStateCallback != null && this.GetOptionStateCallback("GitHubAdvancedSecurity"))
+            if (this.GetOptionStateCallback != null)
             {
-                if (string.IsNullOrWhiteSpace(this.repoPath))
+                bool.TryParse(this.GetOptionStateCallback("GitHubAdvancedSecurity").ToString(), out bool ghAsEnabled);
+                if (ghAsEnabled)
```
EasyRhinoMSFT, line 194: You can collapse this into the if() above:

if (this.GetOptionStateCallback != null
&& bool.TryParse(this.GetOptionStateCallback("GitHubAdvancedSecurity").ToString(), out bool ghAsEnabled)
&& ghAsEnabled)
{
...

edkazcarlson-ms: Did not know I could use out variables in the same line, good to know :)

EasyRhinoMSFT (approved):
:shipit:

PR #603, opened 2023-08-15
Fixing highlights un necessarily stretching

edkazcarlson-ms:
Some users have experienced a bug where a highlighted span will keep stretching as the user keeps writing on the same line. 
For example if you have a file 
```
line 1;
line 2;
line 3;
line 4;
```
and have a highlight spanning from the start to the end of the first line, if you were to put your cursor on the end of line 1 and press enter to make a new line and starting writing, the new 2nd line would also be highlighted, even though it likely has no relevance to whatever error is on line 1.
before
Here is an example of what I mean, as the user puts text on the edges of the highlighted span, the highlight will grow past its initial intended range.
This change changes the SpanTrackingMode https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.text.spantrackingmode?view=visualstudiosdk-2022 to not track. 
As a result, the user is able to write before/after the block of text without flooding the code editor window.
post

### DIFF
--- a/src/Sarif.Viewer.VisualStudio.Core/Tags/SarifLocationErrorTagger.cs
+++ b/src/Sarif.Viewer.VisualStudio.Core/Tags/SarifLocationErrorTagger.cs
@@ -65,8 +65,6 @@ namespace Microsoft.Sarif.Viewer.Tags
         {
             if (this.tagsDirty)
             {
-                this.tagsDirty = false;
-
                 IEnumerable<SarifErrorListItem> errorsInCurrentFile = CodeAnalysisResultManager
                     .Instance
                     .RunIndexToRunDataCache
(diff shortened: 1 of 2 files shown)


yongyan-gh (approved):
:shipit:

src/Sarif.Viewer.VisualStudio.Core/Tags/SarifLocationErrorTagger.cs
```diff
@@ -65,8 +65,6 @@ public IEnumerable<ITagSpan<IErrorTag>> GetTags(NormalizedSnapshotSpanCollection
         {
             if (this.tagsDirty)
             {
-                this.tagsDirty = false;
-
```
yongyan-gh, line 69: wondering if any performance impact after deleting this line. `this.tagsDirty` will always be true and calculate `this.currentTags` every time.

edkazcarlson-ms: I just tested and I did not notice any performance degradation/change when comparing builds, however I also did not notice any difference in functionality so I have added the line back.

PR #605, opened 2023-09-06
Markdown rendering fix

edkazcarlson-ms:
Fixing some markdown bugs that were found as well as making it more appealing visually
- When switching between themes, the color of the font would not automatically update, making it so that the font color would blend into the background and be unreadable
- Fixing some margins
- Changing the color of the divider to be more visually appealing. 
- Preventing the markdown from fitting the width of the screen, taking up more space than usually needed.

Before and after example:
image

### DIFF
--- a/src/Sarif.Viewer.VisualStudio.Core/Tags/SarifLocationErrorTag.cs
+++ b/src/Sarif.Viewer.VisualStudio.Core/Tags/SarifLocationErrorTag.cs
@@ -33,11 +33,6 @@ namespace Microsoft.Sarif.Viewer.Tags
         private const int FontSize = 16;
 
         /// <summary>
-        /// The color for text. Changes based on user theme.
-        /// </summary>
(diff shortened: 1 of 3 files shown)


src/Sarif.Viewer.VisualStudio.Core/Tags/SarifLocationErrorTag.cs
```diff
@@ -129,27 +128,35 @@ public object ToolTipContent
...
+                        runBlock.Foreground = textBrush;
                     }
                     else if (blockChild is ListItem listBlock)
                     {
                         ParseBlocks(listBlock.Blocks);
+                        block.Margin = new Thickness(0, 0, 0, 1);
+                    }
+                    else if (blockChild is InlineUIContainer inlineUIContainer)
+                    {
+                        if (inlineUIContainer.Child is Line line)
```
EasyRhinoMSFT, line 154: >if (inlineUIContainer.Child is Line line) [](http://example.com/codeflow?start=24&length=41)

Merge with the else if

PR #606, opened 2023-10-05
File rooting bugfixes + VS 19 json fixes

edkazcarlson-ms:
This PR addresses 2 main bugfixes:
1. The extension does not play well in VS 19 after I tried to upgrade the ver of the json nuget. [Here](https://devblogs.microsoft.com/visualstudio/using-newtonsoft-json-in-a-visual-studio-extension/) is a blogpost about it but in short every ver of VS ships with its own ver of the newtonsoft json nuget, but when you use a ver that is outside the one that it is shipped with, it can cause issues. In this case, vs 19 comes w/ newtonsoft 12 and so it didn't like how I increased the ver to 13.
2. When opening files through the drag and drop or through the File > Open > File and you don't have the appropriate folder/solution open, it will fail to link the relative file path in the sarif object to the absolute file path on disk if they file seperators are different (/ vs \)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "devCanvas.insightsServerName": "PROD"
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "devCanvas.insightsServerName": "PROD"
+}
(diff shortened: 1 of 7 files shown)


EasyRhinoMSFT:
Shoot, I forgot to block on the Newtonsoft thing, sorry.

EasyRhinoMSFT (approved):
:shipit:

PR #608, opened 2023-10-20
Fix issue cannot open a SARIF log if there is any tool notification.

yongyan-gh:
If the SARIF log contains tool notifications (which is not a SARIF result), viewer got unhandled exception and not able to load the results.

Attached the repro SARIF log.
[repros.zip](https://github.com/microsoft/sarif-visualstudio-extension/files/13058566/repros.zip)

The fix is to skip the notifications in new added code finder process.


yongyan-gh:
[Bug 2116159](https://dev.azure.com/mseng/1ES/_workitems/edit/2116159): CRITICAL: vs viewer won't open SARIF

michaelcfanning (approved):
:shipit:

PR #611, opened 2023-10-26
Stopping un necessary file rooting requests

edkazcarlson-ms:
There is a bug where if you open just a .sarif file, even if the file paths given are absolute it will prompt the user to root the file path.


yongyan-gh (approved):
:shipit:

PR #612, opened 2023-11-14
Updating code finder to better work with multi line snippets

edkazcarlson-ms:
Previously our code finder had only been used with single line snippets. However due to new scenarios, we are encountering snippets that are multi line and so we have to update the code finder algorithm to be more resistant to the different line endings as well as modify the ending line number.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio.Core/CodeFinding/Internals/CodeFinderBase.cs
+++ b/src/Sarif.Viewer.VisualStudio.Core/CodeFinding/Internals/CodeFinderBase.cs
@@ -14,6 +14,12 @@ namespace Microsoft.Sarif.Viewer.CodeFinding.Internal
     internal abstract class CodeFinderBase
     {
         /// <summary>
+        /// The character or sequence of characters that designate a line end.
+        /// Either will be \r\n, \r, or \n.
+        /// </summary>
+        protected string lineEndings;
+
+        /// <summary>
         /// The complete contents of the file.
         /// </summary>
         protected string fileContents;
@@ -47,6 +53,19 @@ namespace Microsoft.Sarif.Viewer.CodeFinding.Internal
             lineCount = GetLineNumber(endOfFile);
 
             IgnoredSpans = GetIgnoredSpans();
+
+            if (fileContents.Contains("\r\n"))
+            {
+                lineEndings = "\r\n";
+            }
+            else if (fileContents.Contains("\r"))
+            {
+                lineEndings = "\r";
+            }
+            else
+            {
+                lineEndings = "\n";
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


edkazcarlson-ms:
@EasyRhinoMSFT / @yongyan-gh just want to make sure this PR doesn't get buried

src/Sarif.Viewer.VisualStudio.Core/CodeFinding/Internals/CodeFinderBase.cs
```diff
@@ -47,6 +53,19 @@ public CodeFinderBase(string fileContents)
...
+                lineEndings = "\r\n";
+            }
+            else if (fileContents.Contains("\r"))
+            {
+                lineEndings = "\r";
+            }
+            else
+            {
+                lineEndings = "\n";
+            }
```
EasyRhinoMSFT, line 68: Do files ever contain a mixture of line endings?

src/Sarif.Viewer.VisualStudio.UnitTests/CodeFinding/UnitTests/CsharpTestMultiLine.cs
```diff
@@ -0,0 +1,42 @@
...
+
+using FluentAssertions;
+
+using Microsoft.Sarif.Viewer.CodeFinding;
+
+using Xunit;
+
+namespace Microsoft.Sarif.Viewer.VisualStudio.UnitTests.CodeFinding.UnitTests
+{
+    public class CsharpTestMultiLine
```
EasyRhinoMSFT, line 18: >CsharpTestMultiLine [](http://example.com/codeflow?start=17&length=19)

Nit: generally cased as "CSharp" (capital S)

src/Sarif.Viewer.VisualStudio.Core/CodeFinding/Internals/CodeFinderBase.cs
edkazcarlson-ms: Not that I've seen, usually files just use 1 type of line ending so that they're easily read by text editors. I could imagine a user could manually add different types of line endings into the same file but I don't think I've seen it before.

https://en.wikipedia.org/wiki/Newline#Representation goes into it deeper but my understanding is that windows uses \r\n, modern *nix based os's like mac and linux use \n, and some tools like older mac os's and some codeql scripts give \r.

EasyRhinoMSFT: Makes sense. IMO, if producers are doing something weird like this they're on their own.

src/Sarif.Viewer.VisualStudio.UnitTests/CodeFinding/UnitTests/CsharpTestMultiLine.cs
EasyRhinoMSFT: Please change the file name as well.

EasyRhinoMSFT (approved):
:shipit:

PR #613, opened 2023-11-21
Using registry keys to limit devcanvas usage to just team members

edkazcarlson-ms:
This is a temporary change that will allow the DevCanvas team to make sure our changes do not affect the wider audience while also allowing us to merge back into main in the near future.

Our current plan for releasing this feature branch is:
1. Commit this PR to the feature branch
2. Get access to the app insights cluster from @michaelcfanning
3. Set up some telemetry points around our auth stack/plugin in order to validate that it is working
4. Merge this feature branch into main in order to guarantee nothing gets changed or refactored in main that would cause headaches in the feature branch
5. When we feel ready, remove the changes in this PR that check for the devcanvas reg key to release to the wider MSFT audience

While we are not going to commit it to the repo, the following is a simple .reg file that can be used to turn on the reg key flag that we are using in this PR.
```reg
Windows Registry Editor Version 5.00

[HKEY_CURRENT_USER\Software\Microsoft\VisualStudio\devcanvas]
"DevCanvasInsider"="True"

```
To use this .reg file, copy the content to a file and make sure the file extension is .reg. Once you have the file saved, double click it and it should automatically apply the registry key changes needed to active the plugin.

### DIFF
--- a/src/Sarif.Viewer.VisualStudio.ResultSources.DeveloperCanvas.Core/Services/DevCanvasResultSourceService.cs
+++ b/src/Sarif.Viewer.VisualStudio.ResultSources.DeveloperCanvas.Core/Services/DevCanvasResultSourceService.cs
@@ -144,13 +144,10 @@ namespace Sarif.Viewer.VisualStudio.ResultSources.DeveloperCanvas.Core.Services
         /// <inheritdoc/>
         public System.Threading.Tasks.Task InitializeAsync()
         {
-            try
+            string devcanvasKey = (string)Registry.GetValue("HKEY_CURRENT_USER\\Software\\Microsoft\\VisualStudio\\devcanvas", "sampleKey", null);
+            if (string.IsNullOrEmpty(devcanvasKey))
             {
-                RegistryKey devCanvasKey = Registry.LocalMachine.OpenSubKey(@"Software\Microsoft\VisualStudio\devcanvas", false);
-            }
-            catch (Exception e)
-            {
-                Console.WriteLine(e.ToString());
+                return System.Threading.Tasks.Task.FromResult(Result.Failure("Not a DevCanvas user."));
             }
 
             // TODO: Remove this when merging into main.
--- a/src/Sarif.Viewer.VisualStudio.ResultSources.DeveloperCanvas.Core/Services/DevCanvasResultSourceService.cs
+++ b/src/Sarif.Viewer.VisualStudio.ResultSources.DeveloperCanvas.Core/Services/DevCanvasResultSourceService.cs
@@ -144,13 +144,19 @@ namespace Sarif.Viewer.VisualStudio.ResultSources.DeveloperCanvas.Core.Services
         /// <inheritdoc/>
         public System.Threading.Tasks.Task InitializeAsync()
         {
+            return IsActiveAsync();
+        }
+
+        /// <inheritdoc/>
+        public Task<Result> IsActiveAsync()
+        {
+            // TODO: remove below to release to general users.
             string devcanvasKey = (string)Registry.GetValue("HKEY_CURRENT_USER\\Software\\Microsoft\\VisualStudio\\devcanvas", "sampleKey", null);
             if (string.IsNullOrEmpty(devcanvasKey))
             {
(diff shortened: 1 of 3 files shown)


src/Sarif.Viewer.VisualStudio.ResultSources.DeveloperCanvas.Core/Services/DevCanvasResultSourceService.cs
```diff
@@ -144,7 +144,19 @@ public class DevCanvasResultSourceService : IResultSourceService
...
+        }
+
+        /// <inheritdoc/>
+        public Task<Result> IsActiveAsync()
+        {
+            // TODO: remove below to release to general users.
+            string devcanvasKey = (string)Registry.GetValue("HKEY_CURRENT_USER\\Software\\Microsoft\\VisualStudio\\devcanvas", "sampleKey", null);
+            if (string.IsNullOrEmpty(devcanvasKey))
+            {
+                return System.Threading.Tasks.Task.FromResult(Result.Failure("Not a DevCanvas user."));
```
ShiningMassXAcc, line 157: `Result.Failure()` does not seem correct here - we have successfully not activated for the user.  This is by design - I would not categorize by-design behavior as a failure.

src/Sarif.Viewer.VisualStudio.ResultSources.DeveloperCanvas.Core/Services/DevCanvasResultSourceService.cs
```diff
@@ -144,7 +144,19 @@ public class DevCanvasResultSourceService : IResultSourceService
...
         {
-            // TODO: Remove this when merging into main.
+            return IsActiveAsync();
+        }
+
+        /// <inheritdoc/>
+        public Task<Result> IsActiveAsync()
+        {
+            // TODO: remove below to release to general users.
+            string devcanvasKey = (string)Registry.GetValue("HKEY_CURRENT_USER\\Software\\Microsoft\\VisualStudio\\devcanvas", "sampleKey", null);
```
ShiningMassXAcc, line 154: Do we want to formalize `sampleKey` a bit more in case we leverage this type of functionality further?  At least name this something more meaningful.

ShiningMassXAcc:
Tagging @michaelcfanning directly - we are still blocked here behind access to application insights.

We want to do the right engineering practice and use Observability to track successes/failures in production given that most interactions should not result in a crash (which we get automatic telemetry for).

src/Sarif.Viewer.VisualStudio.ResultSources.DeveloperCanvas.Core/Services/DevCanvasResultSourceService.cs
edkazcarlson-ms: In a way the other result services do the same pattern, if the service is not active it returns a `Result.Failure` [example](https://github.com/microsoft/sarif-visualstudio-extension/blob/4818b476364eececfaf4dbe0cd3bb2e7eede73f2/src/Sarif.Viewer.VisualStudio.ResultSources.GitHubAdvancedSecurity.Core/Services/GitHubSourceService.cs#L189)

Also when the result source services are being initialized, if the result is not a failure it is added to the list of services that then get hooked into a bunch of events. [link](https://github.com/microsoft/sarif-visualstudio-extension/blob/4818b476364eececfaf4dbe0cd3bb2e7eede73f2/src/Sarif.Viewer.VisualStudio.ResultSources.Factory.Core/ResultSourceFactory.cs#L98C13-L98C13). Thought that it might be weird to constantly have a result source service be hooked into events just to swallow everything.

src/Sarif.Viewer.VisualStudio.ResultSources.DeveloperCanvas.Core/Services/DevCanvasResultSourceService.cs
edkazcarlson-ms: Would `DevCanvasInsider` be better?

ShiningMassXAcc (approved):

EasyRhinoMSFT (approved):
:shipit:

